PR #30, opened 2017-03-15
The new action execution framework with implementing ActionExecutorClient from SDK

DongyiYang:


pkg/action/action_handler.go
```diff
@@ -1,153 +1,208 @@
...
+
+	actionItems := actionExecutionDTO.GetActionItem()
+	// TODO: only deal with one action item.
+	actionItemDTO := actionItems[0]
+	h.execute(actionItemDTO)
+
+	result := <-h.resultChan
+	glog.V(4).Infof("Action result is %++v", result)
+	// TODO: currently the code in SDK make it share the actionExecution client between different workers. Once it is changed, need to close the channel.
+	//close(h.config.StopEverything)
```
enlinxu: If we don't close here, will it cause leak?

pkg/action/action_handler.go
```diff
@@ -1,153 +1,208 @@
...
-		glog.Errorf("Error execute action: %s", err)
-		this.sendActionResult(proto.ActionResponseState_FAILED, int32(0), msgID, "Failed")
+		glog.Errorf("Failed to execute action: %s", err)
+		h.sendActionResult(proto.ActionResponseState_FAILED, int32(0), "Failed to execute action")
+		return
+	}
+	executor, exist := h.actionExecutors[actionType]
+	if !exist {
+		glog.Errorf("action type %s is not support", actionType)
+		h.sendActionResult(proto.ActionResponseState_FAILED, int32(0), "Failed to execute action")
```
enlinxu: Add "The action is currently not supported"

pkg/action/action_handler.go
```diff
@@ -1,153 +1,208 @@
...
+		return
 	}
+
+	action, err := executor.Execute(actionItem)
+	if err != nil {
+		glog.Errorf("Failed to execute action: %s", err)
+		h.sendActionResult(proto.ActionResponseState_FAILED, int32(0), "Failed to execute action")
```
enlinxu: Can we have more specific failure reason to send back to the server?

pkg/action/action_handler.go
```diff
@@ -1,153 +1,208 @@
...
+	h.executedActionChan <- action
 }
 
-func (handler *ActionHandler) ResultChan() *proto.ActionResult {
-	return <-handler.resultChan
+func getActionTypeFromActionItemDTO(actionItem *proto.ActionItemDTO) (turboaction.TurboActionType, error) {
+	var actionType turboaction.TurboActionType
+
+	if actionItem == nil {
+		return actionType, errors.New("ActionItem received in is null")
```
enlinxu: What happened in the caller side if it's nil?

pkg/action/executor/binder.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// Binder is used to schedule pod onto node.
+type binder struct {
+	*client.Client
+}
+
+// Bind just does a POST binding RPC.
+func (b *binder) Bind(binding *api.Binding) error {
+	glog.V(2).Infof("Attempting to bind %v to %v", binding.Name, binding.Target.Name)
```
enlinxu: What happened if Bind failed for some reasons? Do we want to retry here or k8s will retry internally?

pkg/action/executor/horizontal_scaler.go
```diff
@@ -0,0 +1,308 @@
...
+	parentRefObject, _ := probe.FindParentReferenceObject(providerPod)
+	if parentRefObject != nil {
+		parentObjRef = &turboaction.ParentObjectRef{
+			ParentObjectUID:       string(parentRefObject.UID),
+			ParentObjectNamespace: parentRefObject.Namespace,
+			ParentObjectName:      parentRefObject.Name,
+			ParentObjectType:      parentRefObject.Kind,
+		}
+	} else {
+		return nil, errors.New("Cannot automatically scale")
```
enlinxu: Change it to "Cannot perform auto-scale, please make sure the pod is connected to RC"

pkg/action/executor/horizontal_scaler.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	return h.horizontalScale(action)
+}
+
+func (h *HorizontalScaler) buildPendingScalingTurboAction(actionItem *proto.ActionItemDTO) (*turboaction.TurboAction,
+	error) {
+	targetSE := actionItem.GetTargetSE()
+	targetEntityType := targetSE.GetEntityType()
+	if targetEntityType != proto.EntityDTO_CONTAINER_POD || targetEntityType != proto.EntityDTO_APPLICATION {
```
enlinxu: We have similar actionItem validation process for every action type. And every time we check the targetEntityType and actionType. Instead of having this process repeatedly for every action, we should have a method validateAction, which takes two para, allowedEntity and allowedActionType, and in the method, check if this actionItem's targetEntityType and actionType against the two parameters.

pkg/action/executor/horizontal_scaler.go
```diff
@@ -0,0 +1,308 @@
...
+		break
+
+	default:
+		return nil, fmt.Errorf("Error Scale Pod for %s-%s: Not Supported.",
+			parentObjRef.ParentObjectType, parentObjRef.ParentObjectName)
+	}
+
+	// Invalid new replica.
+	if scaleSpec.NewReplicas < 0 {
+		return nil, fmt.Errorf("Cannot scale %s/%s from %d to %d", parentRefObject.Namespace,
```
enlinxu: Make the reason more specific to scaleSpec.NewReplicas < 0

pkg/action/executor/horizontal_scaler.go
```diff
@@ -0,0 +1,308 @@
...
+	scaleSpec, ok := actionContent.ActionSpec.(turboaction.ScaleSpec)
+	if !ok || scaleSpec.NewReplicas < 0 {
+		return nil, errors.New("Failed to setup horizontal scaler as the provided scale spec is invalid.")
+	}
+
+	var key string
+	if actionContent.ParentObjectRef.ParentObjectUID != "" {
+		key = actionContent.ParentObjectRef.ParentObjectUID
+	} else {
+		return nil, errors.New("Failed to setup horizontal scaler consumer: failed to retrieve the key.")
```
enlinxu: Change it to failed to retrieve the UUID of RC or Deployment, since key is the var name which is a bit confusing

pkg/action/executor/horizontal_scaler.go
```diff
@@ -0,0 +1,308 @@
...
+	// So after changing the replica it can return immediately.
+	if action.Content.ActionType == turboaction.ActionUnbind {
+		// Update turbo action.
+		action.Status = turboaction.Executed
+		return action, nil
+	}
+
+	// 4. Wait for desired pending pod
+	// TODO, should we always block here?
+	pod, ok := <-podConsumer.WaitPod()
```
enlinxu: We need to think of a timeout mechanism to make sure it's not waiting forever. For example, what if there's other scheduler picks up the pending pod?

pkg/action/executor/rescheduler.go
```diff
@@ -0,0 +1,261 @@
...
+		// targetNode entityDTO.
+		vmData := targetNode.GetVirtualMachineData()
+		if vmData == nil {
+			return nil, errors.New("Missing VirtualMachineData in ActionItemDTO from server")
+		}
+		machineIPs = vmData.GetIpAddress()
+		break
+	case proto.EntityDTO_PHYSICAL_MACHINE:
+		// TODO
+		// machineIPS = <valid physical machine IP>
```
enlinxu: What is the gap here to support PM? We need to have a list that lists the gap between VM support and bare metal support, so that when we have a customer using bare metal, we know what is supported and what is not.

pkg/action/executor/rescheduler.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, errors.New("Failed to setup re-scheduler as move destination is nil after pre-precess.")
+	}
+
+	var key string
+	if actionContent.ParentObjectRef.ParentObjectUID != "" {
+		key = actionContent.ParentObjectRef.ParentObjectUID
+	} else if actionContent.TargetObject.TargetObjectUID != "" {
+		key = actionContent.TargetObject.TargetObjectUID
+	} else {
+		return nil, errors.New("Failed to setup re-scheduler consumer: failed to retrieve the key.")
```
enlinxu: Again, key is the variable name, the error is about cannot find UID of the pod or provider

pkg/action/executor/rescheduler.go
```diff
@@ -0,0 +1,261 @@
...
+		podClone := &podClone{r.kubeClient}
+		err := podClone.clone(originalPod)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// 5. Wait for desired pending pod
+	// TODO, should we always block here?
+	p, ok := <-podConsumer.WaitPod()
```
enlinxu: Same as before, we need to think a timeout mechanism to prevent thread hanging here.

enlinxu:
Dongyi, we need to add one more check before executing any actions, esp after k8s 1.6 is out. The check is to see if there's any other scheduler (including default scheduler) running, and if yes, then check if the pod we want to execute action on, is belonging to our scheduler or other scheduler foo. If it belongs to foo, we should return the error message in actionItemDTO, something like "You have other scheduler running, the selected Pod is not scheduled by kubeturbo, please restart the pod with kubeturbo as scheduler"

enlinxu:
Please also include an action executor workflow here.

pkg/action/action_handler.go
DongyiYang: No, it will be handled by garbage collection.

pkg/action/executor/rescheduler.go
DongyiYang: 1. We don't have a supply chain that directly link container on PM;
2. The MOVE action cannot be generate at server side;
3. Don't know how does the IPs is passed for a PM.

pkg/action/executor/horizontal_scaler.go
DongyiYang: Then how long should it wait?

enlinxu: Is it possible to find a similar case in k8s, and check how k8s handles it?

pkg/action/action_handler.go
DongyiYang: Yes. But I think you are not talking about simply print out the error returned, right?

pkg/action/action_handler.go
DongyiYang: the actionItemDTO from actionExecutionDTO is nil..

pkg/action/executor/binder.go
DongyiYang: It just fails and wont be retried.

PR #42, opened 2017-04-21
Fixed Incorrect Resource Used Value

DongyiYang:


pkg/discovery/probe/pod_probe.go
```diff
@@ -208,30 +208,40 @@ func (podProbe *PodProbe) getPodResourceStat(pod *api.Pod, podContainers map[str
...
 				continue
-				// return nil, fmt.Errorf("Not enough data for %s", podNameWithNamespace)
 			}
+			// Get the average of all the available data.
+			dataSampleNumber := len(containerStats)
+			glog.V(4).Infof("data sample number is %d", dataSampleNumber)
+
 			currentStat := containerStats[len(containerStats)-1]
-			prevStat := containerStats[len(containerStats)-2]
+			prevStat := containerStats[len(containerStats)-dataSampleNumber]
```
songbinliu: "len(containerStats) - dataSampleNumber"  is not useful, because dataSampleNumber is always equal to len(containerStats).   Should define dataSampleNumber to be other number.

DongyiYang: here we want to get all the data points, from index 0 to len(containerStats) - 1. So I will just change it to 0.

PR #45, opened 2017-04-28
UUID stitching support

DongyiYang:
Created Stitching Manager for Supporting Both IP Based Stitching And UUID Based Stitching.

When there are more than 16 containers deployed in an VM that runs inside VC, vmware tools fail to return the correct IP address of the VM. This is a long existing bug in vmware tools.  As we reply on the correct IP address for stitching process, if VC probe fails to find the IP of VM, stitching fails too.

The solution is to use UUID for stitching VM runs in VC. The same UUID that discovered by VC can also be found as SystemUUID in each nodes information. So we create a special case for VMware VMs. For other hypervisor or cloud environment, we still rely on IP for stitching. 

Now all the stitching related work is done by the new created [stitching manager](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:UuidStitching?expand=1#diff-abad71e122309d9a6b3ec0c807645d7a). It is responsible for finding stitching value based on whether it is UUID-based stitching or IP-based stitching. It also create stitching prop for Pod and reconciling prop for node.


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -32,21 +32,26 @@ const (
...
 
 	TurboServerAddress string
 	TurboServerPort    string
 	OpsManagerUsername string
 	OpsManagerPassword string
 
+	// If the underlying infrastructure is VMWare, we cannot reply on IP address for stitching. Instead we use the
+	// systemUUID of each node, which is equal to UUID of corresponding VM discovered by VM probe.
+	// The default value is false.
+	UseVMWare bool
```
chlam4: So if this bool is true, then we use UUID, or the other way around?

pkg/discovery/probe/node_probe.go
```diff
@@ -18,32 +18,33 @@ import (
...
-const (
-	proxyVMIP string = "Proxy_VM_IP"
-)
-
+// key: node name; value: cAdvisor host.
 var hostSet map[string]*vmtAdvisor.Host = make(map[string]*vmtAdvisor.Host)
 
-var nodeUidTranslationMap map[string]string = make(map[string]string)
+// key: node name; value: node uid.
+var nodeUIDMap map[string]string = make(map[string]string)
```
chlam4: UID here means UUID?

pkg/discovery/probe/pod_probe.go
```diff
@@ -37,18 +37,20 @@ var podAppTypeMap map[string]string
 type PodsGetter func(namespace string, label labels.Selector, field fields.Selector) ([]*api.Pod, error)
 
 type PodProbe struct {
-	podGetter PodsGetter
+	podGetter        PodsGetter
+	stitchingManager *stitching.StitchingManager
 }
 
-func NewPodProbe(getter PodsGetter) *PodProbe {
+func NewPodProbe(getter PodsGetter, config *ProbeConfig, stitchingManager *stitching.StitchingManager) *PodProbe {
```
songbinliu: config is defined, but not used.

pkg/discovery/probe/pod_probe_test.go
```diff
@@ -47,7 +47,7 @@ func (fpg *FakePodGetter) GetPods(namespace string, label labels.Selector, field
 func TestGetPods(t *testing.T) {
 	fakeGetter := &FakePodGetter{}
 
-	podProbe := NewPodProbe(fakeGetter.GetPods)
+	podProbe := NewPodProbe(fakeGetter.GetPods, nil)
```
songbinliu: function NewPodProble requires 3 parameters, only 2 given.

cmd/kubeturbo/app/kubeturbo_builder.go
DongyiYang: yes, if it is true, then use UUID

pkg/discovery/probe/stitching/stitching_manager.go
```diff
@@ -0,0 +1,201 @@
...
+
+type StitchingManager struct {
+	// key: node name; value: node IP address for stitching.
+	nodeStitchingUIDMap map[string]string
+
+	// key: node name; value: node IP address for stitching.
+	nodeStitchingIPMap map[string]string
+
+	//
+	useVMWare bool
```
songbinliu: define some other type of data, stitchIDType for example, to indicate whether IP is used, or UUID is used. And set this stitchIDType according to command line input or configuration.

Because we expect that in other cases (not only vmware), we may also need to use UUID for stitching.

pkg/discovery/probe/node_probe.go
DongyiYang: yes. I will change to UUID for consistency.

pkg/discovery/probe/stitching/stitching_manager.go
DongyiYang: This is a good point. I've thought about this. The reason why I didn't use "stitchIDType" for flag name is stitching is a term only makes sense inside Turbonomic probe. So it is not a good flag name to expose for users.

songbinliu: I agree that "useVMWare" is a good flag name. My thought was that: in implementation translating the command line input( "useVMWare") to some other internal variable could provide more flexible in the future. But it is not necessary for now.

DongyiYang: this is good point. I will change the internal variable name.

PR #52, opened 2017-05-17
Pod probe util change

DongyiYang:
Change GetAppType and GetResourceLimits.
Added Unit Test.

### DIFF
--- a/pkg/discovery/probe/pod_probe.go
+++ b/pkg/discovery/probe/pod_probe.go
@@ -428,7 +428,7 @@ func (podProbe *PodProbe) buildPodEntityDTO(pod *api.Pod, commoditiesSold, commo
 	glog.V(4).Infof("Pod %s will be stitched with VM with %s: %s", podDisplayName, *property.Name, *property.Value)
 
 	// monitored or not
-	monitored := monitored(pod)
+	monitored := monitored(pod, notMonitoredNodes)
 	if !monitored {
 		inactivePods[podNameWithNamespace] = struct{}{}
 	}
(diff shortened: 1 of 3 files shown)


pkg/discovery/probe/pod_probe.go
```diff
@@ -428,7 +428,7 @@ func (podProbe *PodProbe) buildPodEntityDTO(pod *api.Pod, commoditiesSold, commo
 	glog.V(4).Infof("Pod %s will be stitched with VM with %s: %s", podDisplayName, *property.Name, *property.Value)
 
 	// monitored or not
-	monitored := monitored(pod)
+	monitored := monitored(pod, notMonitoredNodes)
```
chlam4: <notMonitoredNodes> seems a global variable.  Why do we choose to pass it in as a argument now?

pkg/discovery/probe/pod_probe_util_test.go
```diff
@@ -127,3 +130,39 @@ func TestBreakdownTurboPodUUID(t *testing.T) {
...
+	for _, item := range table {
+		pod := &api.Pod{}
+		pod.Annotations = item.annotations
+		res := isMirrorPod(pod)
+		if res != item.expects {
+			temp := ""
+			if !item.expects {
+				temp = "not"
+			}
+			t.Errorf("pod is %s a mirror pod, but isMirror returned %t", temp, res)
```
chlam4: It may be helpful to print out the pod, so when there's a failure we know right away what actually fails the test.

pkg/discovery/probe/pod_probe.go
DongyiYang: monitored is a util method and notMonitoredNodes is a global map. I think we don't want it to access a global variable from inside the method. 
I've discussed with Cheuk and we can improve the overall implementation in the future.

PR #54, opened 2017-05-19
Added Schedulable Access Commodity

DongyiYang:
If a node is unschedulable, then the action that move one pod from original node to the unschedulable node is incorrect.
Previous, we set the unschedulable node to be not-monitored. But after stitching, the hypervisor probe will still set the monitored to be true. So this node still participates market activities and is considered as move destinations for pods.
To prevent pods from being scheduled or rescheduled to these nodes, we create another accessCommodity to indicate if the node is schedulable. If the node is schedulable, it sells the commodity with key *`schedulable`*. Otherwise, the node wont sell anything. On the buyer side, if a pod can be scheduled by scheduler and can be rescheduled, we make it buy this access commodity. So that we constrain these pods only shop with the nodes who sells the access commodity. As a result, unschedulable nodes are excluded.

### DIFF
--- a/pkg/discovery/probe/node_probe.go
+++ b/pkg/discovery/probe/node_probe.go
@@ -20,6 +20,12 @@ import (
 	cadvisor "github.com/google/cadvisor/info/v1"
 )
 
+const (
+	accessCommodityDefaultCapacity float64 = 1E10
+
+	schedAccessCommodityKey string = "schedulable"
+)
+
 // key: node name; value: cAdvisor host.
 var hostSet map[string]*vmtAdvisor.Host
 
@@ -104,6 +110,12 @@ func (nodeProbe *NodeProbe) parseNodeFromK8s(nodes []*api.Node, pods []*api.Pod)
 		nodePodsMap[pod.Spec.NodeName] = podList
 	}
 	for _, node := range nodes {
+		// We do not monitor node that is not ready or unschedulable.
+		if !nodeIsReady(node) || !nodeIsSchedulable(node) {
+			glog.V(3).Infof("Node %s is either not ready or unschedulable.", node.Name)
+			//continue
+			notMonitoredNodes[node.Name] = struct{}{}
+		}
 
 		// use cAdvisor to get node info
 		nodeProbe.parseNodeIP(node)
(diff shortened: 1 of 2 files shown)


pkg/discovery/probe/node_probe.go
```diff
@@ -196,11 +208,26 @@ func (nodeProbe *NodeProbe) createCommoditySold(node *api.Node, nodePodsMap map[
 		}
 	}
 
-	// Use Kubernetes service UID as the key for cluster commodity
+	// Create a new VMPMAccessCommodity to indicate if the node is schedulable. If the Node is schedulable, then
+	// the node will sell it. Otherwise, it doesn't sell. On the consumer side, if a pod can be scheduled by
```
enlinxu: "if a pod can be scheduled by scheduler, then the pod should buy the same VMPMAccessCommodity". So if a pod cannot be scheduled by scheduler, it won't by this VMPMAccessCommodity? But not buying an accesscommodity will not provide any constraint, right? In that case, do we still want to decide which pods should buy this commodity, and which ones shouldn't?

chlam4: I had the same question when reviewing it yesterday.  But from my understanding of the code, it seems if a pod is NOT schedulable, it is inactive and it will be marked "not monitored" - meaning it is not participating in the market at all (according to what I learned from Dongyi).  So looks like it's ok, though I agree that we should probably make this implication more explicit.

DongyiYang: yes, as Cheuk explained, if a pod cannot be scheduled by scheduler, we set it to not monitored in pod probe. 

Here I think the main purpose of this accessCommodity is to prevent pods can be scheduled from being scheduled to unschedulable nodes. we can think this schedulable status of a node like a node label, and pods want to be scheduled should select nodes with this schedulable label. Then the way we handle it is just like any other label-selector. 

As for other pods, such as mirror pods or created by DaemonSets, I think how do we control their behavior is a different problem. As Enlin mentioned, they don't buy this schedulable accessCommodity, so from our market stand point, they can run on any node, either schedulable or unschedulable. This is what we expected. For these pods, we only want to monitor them, and prevent them from moving, scaling, etc, basically exclude them from shopping around. We set them to not monitored to make market unaware of these pods.

enlinxu: I understand the approach of NOT_MONITORED (which is good). My point is, I am not sure it's necessary to have logic to decide if the pod should buy the VMPMAccessCommodity (with key schedulable) or not, we can simply let every pod to buy this commodity.

DongyiYang: talked to Enlin. pods running on unschedulable nodes don't buy this VMPMAccessCommodity.

PR #55, opened 2017-05-22
Use UID of Pod as UUID

DongyiYang:
Previously we construct a long UUID by combining UID, namespace and name of a Pod. This guarantees the uniqueness of UUID while make it easy to retrieve pod in action executor. However, the constructed UUID can be very long, which exceeds the max length of the defined length in MySQL database. 
To solve this problem, we use only the UID of pod as the UUID (for example, in[ pod_probe.go](https://github.com/turbonomic/kubeturbo/pull/55/files#diff-dcfdfa7c7a3c8083f5d44b9f3697818bR434)). This still guarantees the uniqueness both in cluster and among clusters. While we can no longer breakdown UUID to get pod namespace and name, we wanted to store those information in entityProperties. However, since there is issue at the server side, this approach is not successful right now. I've implemented code (in [node_util](https://github.com/turbonomic/kubeturbo/pull/55/files#diff-12fc911298180ef752eaabb9424c280aR34), [pod_util](https://github.com/turbonomic/kubeturbo/pull/55/files#diff-a042b1467a7e7cc2dce052e7c821b7a6R203), [application_util](https://github.com/turbonomic/kubeturbo/pull/55/files#diff-62304032211e71a1d48f1024a2e55bbfR18)), but commented out the place (in [util](https://github.com/turbonomic/kubeturbo/pull/55/files#diff-1f59bdcd3b7d27d10eca4a5687ae3433R165)) it calls them. Currently the walk around is to find pod by listing all pods and compare UID of them against the Entity UUID received in ActionItemDTO (see [GetPodFromUUID](https://github.com/turbonomic/kubeturbo/pull/55/commits/07593660a23d703face82b9e657f7a47b7bae39e#diff-1f59bdcd3b7d27d10eca4a5687ae3433R184)). This approach works but will have a lot of overhead if there the number of pods is large or if there is too much API calls. So we want to come back to the approach of storing namespace and name in entity properties.

### DIFF
--- a/pkg/discovery/probe/application_probe_util.go
+++ b/pkg/discovery/probe/application_probe_util.go
@@ -1,0 +1,68 @@
+package probe
+
+import (
+	"github.com/golang/glog"
+	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+)
+
+const (
+	// TODO currently in the server side only properties in "DEFAULT" namespaces are respected, ideally, we should use "Kubernetes-Application".
+	appPropertyNamespace = "DEFAULT"
+
+	appPropertyNameHostingPodNamespace = "Kubernetes-App-Pod-Namespace"
+
+	appPropertyNameHostingPodName = "Kubernetes-App-Pod-Name"
+)
+
+// Build properties of an application. The namespace and name of the hosting pod is stored in the properties.
+func buildAppProperties(podNamespaceName string) []*proto.EntityDTO_EntityProperty {
+	podNamespace, podName, err := BreakdownPodClusterID(podNamespaceName)
+	if err != nil {
+		glog.Errorf("Failed to build App properties: %s", err)
+		return nil
+	}
+	var properties []*proto.EntityDTO_EntityProperty
+	propertyNamespace := appPropertyNamespace
+	hostingPodNamespacePropertyName := appPropertyNameHostingPodNamespace
+	hostingPodNamespacePropertyValue := podNamespace
+	namespaceProperty := &proto.EntityDTO_EntityProperty{
+		Namespace: &propertyNamespace,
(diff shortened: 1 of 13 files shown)


pkg/discovery/probe/node_probe_util.go
```diff
@@ -21,3 +22,38 @@ func nodeIsReady(node *api.Node) bool {
...
+func GetNodeNameFromProperty(properties []*proto.EntityDTO_EntityProperty) (nodeName string) {
+	if properties == nil {
+		return
+	}
+	for _, property := range properties {
+		if property.GetNamespace() != nodePropertyNamespace {
+			continue
+		}
+		if property.GetName() == nodePropertyNameNodeName {
+			nodeName = property.GetValue()
```
songbinliu: for better performance, should return immediately after getting the result.

DongyiYang: OK, make sense. I will change it.

pkg/discovery/probe/application_probe_util.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	for _, property := range properties {
+		if property.GetNamespace() != appPropertyNamespace {
+			continue
+		}
+		if property.GetName() == appPropertyNameHostingPodNamespace {
+			hostingPodNamespace = property.GetValue()
+		}
+		if property.GetName() == appPropertyNameHostingPodName {
+			hostingPodName = property.GetValue()
```
songbinliu: improve performance: if namespace or name has already found, then not necessary to do the string comparison for the found one.

DongyiYang: will change

PR #57, opened 2017-05-23
Moved and expanded the schedule replacement; made it its own section …

DigitalCommodore:
Moved and expanded the schedule replacement; made it its own section and made it more clear that it is an optional step.


DigitalCommodore:
Hey, @DongyiYang - I made some other small modifications... can a get a re-shipit? I didn't realize it would put these commits into the same pull request....

PR #70, opened 2017-06-22
Reschedule

songbinliu:
implement Move by binding-on-creation:
    assign nodeName when creating a new Pod.

### DIFF
--- a/docs/design/actions/move.md
+++ b/docs/design/actions/move.md
@@ -1,0 +1,38 @@
+# movePod #
+[This project](https://github.com/songbinliu/movePod) demonstrates a method that moves pods, which either are created by ReplicationController, or by ReplicaSet(which may be created by Deployment).
+
+# Method #
+**1.** set the schedulerName of the parent object (ReplicationController, or ReplicaSet) of the pod to a **invalid scheduler**; 
+
+**2.** move the pod by [**Copy-Kill-Create**](https://github.com/songbinliu/movePod/blob/master/util.go#L273) steps, and uses the **Binding-on-Creation** way by assigning [pod.Spec.NodeName](https://github.com/kubernetes/client-go/blob/master/pkg/api/v1/types.go#L2470)
+when to create the new the Pod. 
+
+**3.** restore the schedulerName of the parent object.
+
+It should be noted that, if the pod has no parent object, then only the second step is necessary.
+
+# How it works #
+
+It is difficult to move a Pod controlled by ReplicationController/ReplicaSet, because in the second step of 
+the [**Copy-Delete-Create**] move operation, the ReplicationController/ReplicaSet will create a new Pod immediately 
+to make sure there is enough number of Running replicas. However, ReplicationController/ReplicaSet also amkes sure that 
+there is no more than desired number of Running replicas. So the pod created by our move operation have to 
+compete with the pod created by ReplicationController/ReplicaSet: [the first to get to the **running** 
+state will survive (see experiment)](https://gist.github.com/songbinliu/7576bd84bab50f4e399d979d7998cdf6#an-experiment).
+
+If we can make sure that the pod created by ReplicationController/ReplicaSet is scheduled **later than** the pod 
+created by our move operation, then our pod will almost alway be quicker to get to **running** state. We achive this 
+by assigning an none-exist scheduler name to the ReplictionController/ReplicaSet before the **Delete** step: which makes sure 
+the pod created by ReplicationController/ReplicaSet won't be scheduled. And because our pod don't need to be scheduled, 
+and bind to the new node directly. So our pod will get to the **running** state first. 
+(But if it takes too much time to run the pod or failed to run the pod, then our pod will be deleted.)
+
+In the end of the move operation, we restore the scheduler name of the ReplicationController/ReplicaSet, 
+to clear our modifiaction to the parent Controller.
+
+
+# Running Example #
+Test this method [here](https://github.com/songbinliu/movePod).
+
+# Other Info #
+Here are some [experiments](https://gist.github.com/songbinliu/6b28a15ac718a070ab66cff44f0cc056) about Kubernetes 1.6 [advanced scheduling feature](http://blog.kubernetes.io/2017/03/advanced-scheduling-in-kubernetes.html).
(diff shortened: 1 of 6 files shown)


pkg/discovery/probe/pod_probe_util.go
```diff
@@ -149,28 +149,23 @@ func GetAppType(pod *api.Pod) string {
...
-				res = res + na[i]
-			}
-			return res
-		default:
+
+		if parent == nil {
 			return pod.Name
 		}
+
+		return parent.Name
```
DongyiYang: if a pod is created by deployment, then the name of the associated replica set will be in the following format:
`<the name of the Deployment>-<hash value of the pod template>`
for example:
`frontend-1216308390`

Do we want to keep the hash-value in `appType`?

songbinliu: I am not sure. In fact, Deployment will create a ReplicaSet with a name = "deploymentName-hash"; then the ReplicaSet will create Pods. 

The parent of ReplicaSet can not be found through _metadata.annotations_, but can be found through _ownerReferences_.

I will do some tests about deployment, and discuss with you about the **appType**.

songbinliu: Shall we do some change in FindParentObjectRef: for pod created by ReplicaSet, we will try to find the Deployment?

DongyiYang: yes, I think so. As if we want to update the replicas, we can only update it in deployment instead of replicaSet.

docs/design/actions/move.md
```diff
@@ -0,0 +1,29 @@
+# movePod #
+[This project](https://github.com/songbinliu/movePod) demonstrates a method that can move pods, which either are created by ReplicationController, or by ReplicaSet(which may be created by Deployment).
```
DongyiYang: change to

[This project](https://github.com/songbinliu/movePod) demonstrates a method that moves pods, which are either created by ReplicationController, or by ReplicaSet(which may be created by Deployment).

docs/design/actions/move.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Method #
+**1.** set the schedulerName of the parent object (ReplicationController, or ReplicaSet) of the pod to a **invalid scheduler**;
```
DongyiYang: why set it to a invalid scheduler?

docs/design/actions/move.md
```diff
@@ -0,0 +1,29 @@
+# movePod #
+[This project](https://github.com/songbinliu/movePod) demonstrates a method that can move pods, which either are created by ReplicationController, or by ReplicaSet(which may be created by Deployment).
+
+# Method #
+**1.** set the schedulerName of the parent object (ReplicationController, or ReplicaSet) of the pod to a **invalid scheduler**; 
+
+**2.** move the pod by [**Copy-Delete-Create**](https://github.com/songbinliu/movePod/blob/master/util.go#L273) steps, and uses the **Binding-on-Creation** way by assigning [pod.Spec.NodeName](https://github.com/kubernetes/client-go/blob/master/pkg/api/v1/types.go#L2470)
```
DongyiYang: change `Copy-Delete-Create` to `Copy-Kill-Restart`?

docs/design/actions/move.md
```diff
@@ -0,0 +1,29 @@
...
+when to create the new the Pod. 
+
+**3.** restore the schedulerName of the parent object.
+
+It should be noted that, if the pod has no parent object, then only the second step is necessary.
```
DongyiYang: we don't need to check scheduler name to invalid and then restore it in this case?

docs/design/actions/move.md
```diff
@@ -0,0 +1,29 @@
...
+
+If we can make sure that the pod created by ReplicationController/ReplicaSet is scheduled **later than** the pod 
+created by our move operation, then our pod will almost alway be quicker to get to **running** state. We achive this by assigning an none-exist scheduler name to the ReplictionController/ReplicaSet before the **Delete** step: which makes sure 
+the pod created by ReplicationController/ReplicaSet won't be scheduled. And because our pod don't need to be scheduled, and bind to the new node directly. So our pod will get to the **running** state first. (But if the new node is too slow to run the pod, or failed to run the pod, then our pod will be deleted.)
```
DongyiYang: change
`But if the new node is too slow to run the pod, or failed to run the pod, then our pod will be deleted.` to ` But if it takes too much time to run the pod or failed to run the pod, then our pod will be deleted`

docs/design/actions/move.md
songbinliu: Make sure the Pod created by RelicaitonController to be Pending until the pod created by the Move operation to be Running.

docs/design/actions/move.md
songbinliu: yea, it will be better to check the scheduler name before restore it.
I will change the code.

docs/design/actions/move.md
songbinliu: "Create" seems better than "Restart".   Because we indeed create a new Pod with the old PodSpec.

pkg/action/supervisor/action_supervisor.go
```diff
@@ -86,8 +87,15 @@ func (s *ActionSupervisor) checkMoveAction(action *turboaction.TurboAction) (boo
...
 	targetPod, err := s.config.kubeClient.CoreV1().Pods(podNamespace).Get(podName, metav1.GetOptions{})
 	if err != nil {
-		glog.Errorf("Cannot find pod %s: %s", podIdentifier, err)
-		return false, fmt.Errorf("Cannot find pod %s in cluster after a move action", podIdentifier)
+		err = fmt.Errorf("move-check failed: cannot find pod %v: %v", podIdentifier, err.Error())
+		glog.Error(err.Error())
+		return false, err
+	}
+
+	if targetPod.Status.Phase == api.PodFailed {
```
DongyiYang: I think we only move a `Running` pod, so why not
```go
if targetPod.Status.Phase != api.PodRunning {
    //...
}
```

pkg/discovery/probe/pod_probe_util.go
DongyiYang: After talking with Songbin, we decide to change this in a separate PR. Add a **TODO** for now.

PR #72, opened 2017-06-28
A New Monitoring Framework With Discovery Client and Monitoring Worker

DongyiYang:
**NOTE**: this is the first PR about the new working discovery framework. The old one is still there and used to compare result in the [discovery_result_validator](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-e60bca4916a6dc4d03a3d0009f690e43).

# New Discovery Framework
Discovery process

1. Discovery client receives discovery request from server
2. Distribute discovery tasks to discovery workers. 
     - if there is not enough discovery worker, create new ones on-demand. 
3. Discovery worker invokes different monitor workers to scraping data from monitoring sources.
4. In each monitor worker, data are retrieved, transformed and put into a data sink
5. Once a monitor worker finishes its job,  the data sink from monitor worker is merged with the sink defined in discovery worker
6. As all monitoring worker finish their work, entityDTOs will be built.
7. Discovery result is sent back to discovery worker, then send back to server.

discovery - page 1

[Discovery worker
](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-22b3d81625ca5c3773f4175e48590fa7)A discovery worker is consisted of a collections of monitoring workers, a data sink and entityDTO builders. Once it receives a discovery task, it first performs some preprocessing and then calls different monitoring workers for collecting data. For each monitor worker, once it finishes its job, the data is merges into the sink defined in discovery worker. After all the monitor workers finish, the sink is passed to different entityDTO builders to build entityDTOs. The discovery result is consisted of these entityDTOs.

[Service Discovery worker
](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-068b04b9e508c4d55d326bb71cce8a31)Service discovery worker is a special worker. Why is it specially? Currently, tasks are distributed as a subset of nodes. In this case, if we discover a node, we can also discover all the pods and applications running on it and get all the required metrics. However, service is not a real entity running on a specific node. A service have different endpoints, which are pods potentially running on different nodes. So to get the consumer-provider relationship of services-application and the resource consumption between them, we need to collection info from the view of the whole cluster. For now, as all the resources (transaction, latency, etc) consumed by service are exactly equal to the values provided/sold by application. So here we don’t need any extra monitor worker to scrape a data source. A post-processing on the existing entityDTOs that returned by other discovery workers can do the job. On the other hand, this also means service worker should start to do its task after all the other discovery workers finish.

[Monitoring worker
](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-b597e6526040daa9b1060caefb9573ec)A monitoring worker is responsible for scraping metrics from a data source. Each monitoring worker is responsible for only one source and is either a resource monitor or state monitor. For now, we have Kubelet, K8sConntrack and Kube-api-server as data sources. The first two are ResourceMonitors while the last one is a stateMonitor. In the near future, we will have Prometheus.
In general, inside each monitoring worker, it builds a client to the monitor source, making API request to get metrics back. Then it transforms metrics’ values to the standard unit, like core for Cpu, MHz for Cpu frequency and  KB for memory.
Within each monitor worker, there is a data sink to store the metrics discovered during current cycle. Once the worker finishes its task, all the metrics are dumped into this sink and another component can retrieve data from the sink.

## Files to review:

worker package:
  - [k8s_discovery_worker](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-22b3d81625ca5c3773f4175e48590fa7)
  - [service_discovery_worker](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-068b04b9e508c4d55d326bb71cce8a31)
  - [dispatcher](https://github.com/turbonomic/kubeturbo/pull/72/commits/f92a45fba3b318ccd96d64ef1b93313a8c58995d#diff-43081c170f97526a53381762d2cb0f34)
  - [result_collector](https://github.com/turbonomic/kubeturbo/pull/72/commits/f92a45fba3b318ccd96d64ef1b93313a8c58995d#diff-79aa242984e3a78eab294e75c18bac29)

dtofactory package:
  - [application_entity_dto_builder](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-7d160b7e5098c846ad3cac07dbf63c0e)
  - [node_entity_dto_builder](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-8af8843edde8365329c44ab0c3b9d2dc)
  - [pod_entity_dto_builder](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-d1d0eaf1269b5f1629c2c49221e6e7f7)
  - [service_entity_dto_builder](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-a58d06d445b0f1eb8e1f4c420d67790c)

metrics package:
  - [metric](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-8f39f854b9577e1f9584f770bdaa16ca)
  - [metric_sink](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-327a4acbf7833f1b6d07508fbd0a3a2a)

monitoring package:
  - [k8sconntrack_monitor](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-4a54071899765a4004bc9b95b46bd575)
  - [kubelet_monitor](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-c190760e13d3a1ad94722f0b9e3e3b76)
  - [cluster_monitor](https://github.com/turbonomic/kubeturbo/compare/master...DongyiYang:NewFrameworkCommit?expand=1#diff-3789a2806c744a35021246fe45a93b22)


pkg/discovery/monitoring/kubelet/kubelet_client.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return nil, err
+	}
+	summary := &stats.Summary{}
+	client := kc.client
+	err = httputil.PostRequestAndGetValue(client, req, summary)
+	return summary, err
+}
+
+func (kc *kubeletClient) GetMachineInfo(host Host) (*cadvisorapi.MachineInfo, error) {
```
songbinliu: It is not working for Openshift 3.4 cluster.

screen shot 2017-07-05 at 1 01 45 pm

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -0,0 +1,174 @@
...
+func (m *KubeletMonitor) RetrieveResourceStat() error {
+	if m.nodeList == nil || len(m.nodeList) == 0 {
+		return errors.New("Invalid nodeList or empty nodeList. Finish Immediately...")
+	}
+	m.wg.Add(len(m.nodeList))
+
+	for _, node := range m.nodeList {
+		// TODO, do we want to add timeout?
+		n := node
+		go m.scrapeKubelet(n)
```
songbinliu: Performance concern:
should have a limit of the maximum number of concurrent goroutines.

pkg/discovery/metrics/metric_sink.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func NewEntityMetricSink() *EntityMetricSink {
+	return &EntityMetricSink{
+		data: turbostore.NewCache(),
+	}
+}
+
+// Add one or more metric entries to sink.
+func (s *EntityMetricSink) AddNewMetricEntries(metric ...Metric) {
```
songbinliu: Performance concerns:
      each s.data.Add() requires a Lock. 

It would be better to implement a AddList() interface to add bulk of items with one lock request.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -0,0 +1,174 @@
...
+
+		// application cpu and mem used are the same as pod's.
+		applicationCpuUsageCoreMetrics := metrics.NewEntityResourceMetric(task.ApplicationType,
+			util.PodStatsKeyFunc(podStat), metrics.CPU, metrics.Used,
+			float64(cpuUsageNanoCoreSum)/util.NanoToUnit)
+		applicationMemoryUsageCoreMetrics := metrics.NewEntityResourceMetric(task.ApplicationType,
+			util.PodStatsKeyFunc(podStat), metrics.Memory, metrics.Used,
+			float64(memoryUsageBytesSum)/util.KilobytesToBytes)
+
+		m.metricSink.AddNewMetricEntries(podCpuUsageCoreMetrics,
```
songbinliu: Performance concerns:
    accumulate the items into bulk, and add them to cache in one lock request.

pkg/discovery/monitoring/kubelet/kubelet_client.go
DongyiYang: if use curl, certificate and public key should be provided.

pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+		for _, p := range nodeNonTerminatedPodsList.Items {
+			pod := p
+			podList = append(podList, &pod)
+		}
+	}
+	return podList
+}
+
+func (worker *k8sDiscoveryWorker) findRunningPods(nodeName string) (*api.PodList, error) {
```
songbinliu: running pods list API is called four times:
   (1) pkg/discovery/monitoring/k8sconntrack/k8sconntrack_monitor.go
   (2) pkg/discovery/monitoring/master/cluster_info_scraper.go           
   (3)  pkg/discovery/worker/k8s_discovery_worker.go
   (4) pkg/discovery/worker/service_discovery_worker.go  
            ---buildPodClusterIDToPod()

  1. Shall we cache the Pods?
     cache the Pods by Node, or use a cache with index by node Name.
Here is a cache with index:
https://github.com/kubernetes/client-go/blob/master/tools/cache/thread_safe_store.go

  2. Seems you consider a lot about flexibility that new types of metrics scrapers can be added easily?
      However, from my understanding, there will not be too many kind of scrapers, nor will then change much. And even you can add new types of scrapers easily, you have to parse the new metrics, and change the code about buildDTOs.
     So my suggestion is to store the different metrics separately.  Let another stage of the Pipeline use all these kinds of metrics to build DTOs.

pkg/discovery/worker/dispatcher.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (d *Dispatcher) WorkerPool() chan chan *task.Task {
+	return d.workerPool
+}
+
+func (d *Dispatcher) Dispatch(nodes []*api.Node, workerCount int) int {
+	// make sure when len(node) < workerCount, worker will receive at most 1 node to discover
+	perTaskNodeLength := int(math.Ceil(float64(len(nodes)) / float64(workerCount)))
```
songbinliu: This is one-producer-multi-consumers framework.
Why not put the tasks into a channel, and all the workers fetch tasks from this public channel?
  (https://stackoverflow.com/questions/11075876/what-is-the-neatest-idiom-for-producer-consumer-in-go)

(1) Each consumer do not have to register its task channel to the producer.
(2) The tasks can be split in Node level, so there is no need to divide the tasks into bigger chunks for each consumer.

pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func (worker *k8sDiscoveryWorker) registerToDispatcher(dispatcher *Dispatcher) {
+	glog.V(3).Infof("Register current worker %s to dispatcher", worker.id)
+	dispatcher.WorkerPool() <- worker.taskChan
+}
+
+// Worker start to working on task.
+func (worker *k8sDiscoveryWorker) executeTask(currTask *task.Task) *task.TaskResult {
```
songbinliu: The goal of each Task is to build NodeDTO, PodDTOs, and ApplicationDTOs.
I suggest to use a two-stage pipeline to implement this goal:
    (1) stage-1: for each Node, get all the metrics(k8sconntrack metrics, kubelet metrics, k8sAPI info);
    (2) stage-2: once the metrics are ready for **a node,** build these DTOs for this node.

songbinliu: In current implementation, the process is in two stage, but they are in the same goroutine.

pkg/discovery/worker/dispatcher.go
DongyiYang: (1) The design is to create a pool of workers instead of a simple task queue. Here dispatcher is aware of the number of available consumers. If we want, we can also change the code a bit to make dispatcher aware of which consumer is consumer which task. 
(2) you approach decrease the dispatcher complexity, but will increase the number of API calls, especially for those getting information shared by entities running on different nodes, like clusterID.

In general, my original thoughts are: the max number of workers is the same as num of CPU. So lets say the num of CPU is 4, and there are 16 nodes, then each worker will be responsible for 4 nodes. If we let each worker only discover 10 nodes at maximum (based on the max time of each discovery takes). Then to discover a cluster with 160 nodes, each worker will be used 4 times. Probably the discovery time is still acceptable. But, when there are 1600, then each discovery worker will be used 40 times!! Discover time can be very long. The solution is to create more temporary workers running in different pods in the cluster. We can assume at this point the overhead of creating these worker is much less than the overall discovery time.  Ideally, we should come up with an equilibrium where some tasks are finished by reused workers, some are finished by temporary workers. Of course, the number of temporary workers depends on the average time of creating a worker pod, the average time of a worker to finish a discovery and the size of the cluster.

pkg/discovery/worker/k8s_discovery_worker.go
DongyiYang: I think currently it is a two-stage operating, although the logic are written in the same function. I can break down the executeTask function into two sub-functions

pkg/discovery/worker/dispatcher.go
songbinliu: (1) A queue of tasks does not conflict with worker pool, and the architecture is simpler.
        https://play.golang.org/p/Y_sl7Ioeqs
   (dispatcher awareness of the consumer is not related to the architecture.)
  
(2)  Since each worker keeps live as long as the channel is not close, 
        the public info can be kept in the worker. 

(3) If multi-pods are needed for a cluster, then we should implement a distributed solution, and review it then.  And It will be similar amount of coding from current implementation to a distributed implementation, in either way.

pkg/discovery/worker/k8s_discovery_worker.go
songbinliu: yea, it is in two stages. To make the best use of a two-stage pipeline,  fine-grained task splitting is necessary.  The first stage is mainly about network-IO/wait, and the second stage is cpu intensive workload. So it is a good place to use pipeline.

pkg/discovery/worker/dispatcher.go
songbinliu: I am wondering that why don't we use heapster to get cpu/memory usage?  currently we use kubelet_monitor to get the same thing. 
If we can use hipster, then there will be much less network coummunications:
    (1) one call to Heapster to get all the cpu/memory usage;
    (2) one call to K8s API server to get all the Pods;
    (3)  one call K8s API server to get all the Services;
     (4)  ... endpoints;

All the workload left is to build the DTOs based on these data, which should be much faster.

DongyiYang: like promethues, if we use heapster, we will have to ask user to deploy heapster in the cluster

DongyiYang: Also, as all the monitor worker are pluggable, we can definitely support heapster if necessary. For now kubelet is the only option to get cpu/mem consumption. But later, the metrics can come from promethues or heapster or other monitoring sources

pkg/discovery/monitoring/master/cluster_monitor.go
```diff
@@ -0,0 +1,301 @@
...
+	} else {
+		return metrics.EntityStateMetric{}
+	}
+}
+
+// ----------------------------------------------- Pod State -------------------------------------------------
+func (m *ClusterMonitor) findPodStates() {
+	for _, podList := range m.nodePodList {
+		for _, p := range podList.Items {
+			pod := p
```
songbinliu: pod := p  is not necessary. can use &p directly.
Or more efficient, travel the podList.Items in a way:
```go
 for i := 0; i < len(podList.Items); i ++ {
     p = &(podList.Items[i])
    ...
}
```

pkg/discovery/k8s_discovery_client.go
```diff
@@ -124,9 +170,40 @@ func (dc *K8sDiscoveryClient) Discover(accountValues []*proto.AccountValue) (*pr
...
+		return nil, fmt.Errorf("failed to list all nodes in the cluster: %s", err)
+	}
+	var nodes []*api.Node
+	for _, n := range nodeList.Items {
+		node := n
+		nodes = append(nodes, &node)
 	}
 
-	return discoveryResponse, nil
+	workerCount := dc.dispatcher.Dispatch(nodes, workerCount)
```
songbinliu: not necessary to transfer the []api.Node to  []*apiNode.
Slices are reference of array, so you can pass the nodeList.Items into dc.dispatcher.Dispatch() directly without worrying about efficiency.
And in the function Dispatch(), you can use the slices, without the necessary to change the content to points.
 

I get the conclusion based on this:
http://openmymind.net/The-Minimum-You-Need-To-Know-About-Arrays-And-Slices-In-Go/

https://play.golang.org/p/dT24ThNddw

pkg/discovery/monitoring/master/cluster_monitor.go
DongyiYang: we cannot use &p directly, as p's content keeps changing while its address remains the same.

I think the code you provided is the same, why is it more efficient?

songbinliu: why  &p won't work?

```go
 for i := 0; i < len(podList.Items); i ++ {
     p = &(podList.Items[i])
    key := util.PodKeyFunc(p)
     ...
    podResourceMetrics, err  := m.getPodResourceMetric(p)
    ...
}
```

In the "for ... range ..." style, for each entry  it **assigns** iteration values to corresponding iteration variables.  I think the assignment of a type of "Pod" is more expensive.

DongyiYang: I was wrong. It works fine in this case. I was worried it would break at other place where it builds slice like `[]*api.Pod` and when it is used in sth like
```go
for _, p := range pods {
  go func(pod *api.Pod) {...} (&p)
}
```

I am not sure if this java style `for-loop` outperforms `for...range...`  in go

pkg/cluster/cluster_info_scraper.go
```diff
@@ -0,0 +1,169 @@
...
+	return s.GetPods(api.NamespaceAll, listOption)
+}
+
+func (s *ClusterScraper) GetPods(namespaces string, opts metav1.ListOptions) ([]*api.Pod, error) {
+	podList, err := s.CoreV1().Pods(namespaces).List(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	pods := []*api.Pod{}
```
songbinliu: append is kind of less efficient. 
Since the total length of the array is known here,  it is better call make([]*api.Pod, length).

PR #73, opened 2017-06-28
fix bugs in Move

songbinliu:


pkg/action/executor/rescheduler.go
```diff
@@ -214,12 +214,12 @@ func (r *ReScheduler) reSchedule(action *turboaction.TurboAction) (*turboaction.
 	}
 
 	preScheduler, err := f(r.kubeClient, namespace, parentName, "", DefaultNoneExistSchedulerName)
+	defer f(r.kubeClient, namespace, parentName, DefaultNoneExistSchedulerName, preScheduler)
```
DongyiYang: if there is err, do we still get correct value for preScheduler?

songbinliu: yea, 
(1) If there is an  error, and fails to update the schedulerName, then preScheduler will be set to empty(""); in this case, the following call of f() in line-217 will do nothing.

(2) If there is an error, and the schedulerName  **may** have been updated, then the following call of f() in line-217 will try to restore the schedulerName.

songbinliu: or this way may be better:

restore := func() {
		if preScheduler != "" {
			f(client, nameSpace, *noexistSchedulerName, preScheduler)
		}
	}
	defer restore()

DongyiYang: OK, got it. I think a better way to write this is 
```go
preScheduler := getScheduler()
defer resetSchedler(preScheduler)
setScheduler(DefaultNoneExistSchedulerName)
```

But this way also works and I am fine with it.

PR #75, opened 2017-07-12
Expose Kubelet Https and Port Option

DongyiYang:
Exposed Kubelet https and port option.
By default, https is **disabled** and the kubelet port is `10255`, which is a `http` port of kubelet.

So when running in openshift, the following parameters should be provided to kubeturbo:

```
	 --kubelet-https=true 
	 --kubelet-port=10250 
```
So the pod definition will change to 
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: kubeturbo
  labels:
    name: kubeturbo
spec:
  nodeSelector:
    role: master
  containers:
  - name: kubeturbo
    image: vmturbo/kubeturbo:<VERSION>
    command:
      - /bin/kubeturbo
    args:
      - --v=3
      - --kubeconfig=/etc/kubeturbo/admin.kubeconfig
      - --turboconfig=/etc/kubeturbo/config
      - --kubelet-https=true
      - --kubelet-port=10250
    volumeMounts:
    - name: turbo-config
      mountPath: /etc/kubeturbo
      readOnly: true
  volumes:
  - name: turbo-config
    hostPath:
      path: /etc/kubeturbo
  restartPolicy: Always```

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -59,6 +59,10 @@ type VMTServer struct {
 	// systemUUID of each node, which is equal to UUID of corresponding VM discovered by VM probe.
 	// The default value is false.
 	UseVMWare bool
+
+	// Kubelet related config
+	KubeletPort        uint
+	EnableKubeletHttps bool
 }
 
 // NewVMTServer creates a new VMTServer with default parameters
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/discovery/monitoring/types/const.go [hunk 1]
```diff
@@ -15,3 +15,7 @@ const (
 	ResourceMonitor MonitorType = "ResourceMonitor"
 	StateMonitor    MonitorType = "StateMonitor"
 )
+
+type Source struct {
```
enlinxu: what is this for?

PR #78, opened 2017-07-20
Move action

songbinliu:
1. enable move action for Openshift & K8s
2. add some retry;
3. add clean up after move;


pkg/kubeturbo_service.go
```diff
@@ -24,12 +24,12 @@ type KubeturboService struct {
 	k8sTAPService *K8sTAPService
 }
 
-func NewKubeturboService(c *Config) *KubeturboService {
+func NewKubeturboService(c *Config, k8sVersion, noneScheduler string) *KubeturboService {
```
DongyiYang: what is this noneScheduler? why not put it and k8sVersion into config?

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,550 @@
...
+	id := fmt.Sprintf("%v/%v", pod.Namespace, pod.Name)
+	glog.V(2).Infof("move-pod: begin to move %v from %v to %v",
+		id, pod.Spec.NodeName, nodeName)
+
+	npod := &api.Pod{}
+	copyPodInfo(pod, npod)
+	npod.Spec.NodeName = nodeName
+
+	//2. kill original pod
+	var grace int64 = podDeletionGracePeriod
```
DongyiYang: why not
```go
grace := podDeletionGracePeriod
```

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,550 @@
...
+	if err != nil {
+		err = fmt.Errorf("move-failed: failed to delete original pod-%v: %v",
+			id, err.Error())
+		glog.Error(err.Error())
+		return nil, err
+	}
+
+	//3. create (and bind) the new Pod
+	//glog.V(2).Infof("sleep 10 seconds to test the behaivor of quicker ReplicationController")
+	time.Sleep(time.Duration(grace) * time.Second) // this line is for experiments
```
DongyiYang: the grace period is always in second?

songbinliu: yea, it is in second.

pkg/action/executor/rescheduler.go
```diff
@@ -172,6 +184,73 @@ func (r *ReScheduler) preActionCheck(podName, namespace, nodeName string) (*api.
 	return pod, nil
 }
 
+// move the pods controlled by ReplicationController/ReplicaSet
+func (r *ReScheduler) moveControllerPod(pod *api.Pod, parentKind, parentName, nodeName string) (*api.Pod, error) {
+
+	id := fmt.Sprintf("%v/%v", pod.Namespace, pod.Name)
```
DongyiYang: to get the cluster ID, we can use
```go
import "github.com/turbonomic/pkg/discovery/util"

id := util.GetPodClusterID(pod)
```

pkg/action/executor/rescheduler.go
```diff
@@ -172,6 +184,73 @@ func (r *ReScheduler) preActionCheck(podName, namespace, nodeName string) (*api.
...
+	var update func(*kclient.Clientset, string, string, string) (string, error)
+	switch parentKind {
+	case kindReplicationController:
+		glog.V(3).Infof("pod-%v parent is a ReplicationController-%v", id, parentName)
+		update = updateRCscheduler
+		if highver < 0{
+			update = updateRCscheduler15
+		}
+	case kindReplicaSet:
+		glog.V(2).Infof("pod-%v parent is a ReplicaSet-%v", id, parentName)
```
DongyiYang: change the log level to 3 or 4.

pkg/action/executor/rescheduler.go
```diff
@@ -172,6 +184,73 @@ func (r *ReScheduler) preActionCheck(podName, namespace, nodeName string) (*api.
...
+				return nil
+			})
+		}
+
+		err := cleanPendingPod(client, nameSpace, noexist, parentKind, parentName, highver)
+		if err != nil {
+			glog.Error("failed to clean pending pod for MoveAction:%v", err.Error())
+		}
+	}
+	defer restore()
```
DongyiYang: add some comments about what does restore do?

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,558 @@
...
+	}
+
+	pod.Annotations[SchedulerAnnotationKey] = newName
+	return current
+}
+
+
+//update the schedulerName of a ReplicationController
+// return the previous schedulerName, or return "" if update is not necessary or updated failed.
+func updateRCscheduler15(client *kclient.Clientset, nameSpace, rcName, schedulerName string) (string, error) {
```
DongyiYang: add comments about why we need to have a different func to deal with 15

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,558 @@
...
+
+	if currentName == expectedScheduler {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// shedulerName is set in annotations for kubernetes version < 1.6
+func updateAnnotedScheduler(pod *api.PodTemplateSpec, newName string) string {
```
DongyiYang: function name should be `updateAnnotatedScheduler`?

DongyiYang: one minor suggestion, change the parameter name from `newName` to `newSchedulerName`.

DongyiYang: I think this updateAnnotatedScheduler can be breakdown into two functions:
```go
currSchedulerName := getSchedulerName()
if currSchedulerName != newSchedulerName {
  updateScheduler(newSchedulerName)
}

File: pkg/action/executor/move_util.go [hunk 9]
```diff
@@ -0,0 +1,551 @@
...
+		FieldSelector: "status.phase=" + string(api.PodPending),
+	}
+
+	pods, err := podClient.List(option)
+	if err != nil {
+		glog.Error("failed to cleanPendingPod: %s", err.Error())
+		return err
+	}
+
+	var grace int64 = 0
```
DongyiYang: I would write it as:
```go
grace := int64(0)
```

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,551 @@
...
+	if rs.Spec.Template.Spec.SchedulerName == schedulerName {
+		glog.V(3).Infof("no need to update schedulerName for RS-[%v]", rsName)
+		return "", nil
+	}
+
+	//2. update schedulerName
+	rs.Spec.Template.Spec.SchedulerName = schedulerName
+	_, err = rsClient.Update(rs)
+	if err != nil {
+		err = fmt.Errorf("failed to update RC-%v:%v\n", id, err.Error())
```
DongyiYang: minor suggestion:
```go
err = fmt.Errorf("failed to update RC-%s: %s.\n", id, err)
```

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+
+	return currentName, nil
+}
+
+//---------------- for Kubernetes server version <= 1.6 -------------------------------
+// for Kubernetes version < 1.6, the schedulerName is set in Pod annotations, not in schedulerName field.
+const (
+	SchedulerAnnotationKey = "scheduler.alpha.kubernetes.io/name"
+	EmptyScheduler         = "None"
```
DongyiYang: Do they need to be public?

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,551 @@
...
+	t0 := time.Now()
+
+	var err error
+	for i := 0; ; i++ {
+		if err = myfunc(); err == nil {
+			glog.V(3).Infof("[retry-%d/%d] success", i+1, attempts)
+			return nil
+		}
+
+		glog.V(2).Infof("[retry-%d/%d] Warning %s", i+1, attempts, err.Error())
```
songbinliu: this is for debug, I changed the V from 2 to 4.

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,551 @@
...
+	"time"
+
+	"github.com/golang/glog"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kclient "k8s.io/client-go/kubernetes"
+	api "k8s.io/client-go/pkg/api/v1"
+)
+
+//compare two version strings
```
songbinliu: some tests are added now.

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,551 @@
...
+		}
+
+		glog.V(2).Infof("[retry-%d/%d] Warning %s", i+1, attempts, err.Error())
+		if i >= (attempts - 1) {
+			break
+		}
+
+		if timeout > 0 {
+			if delta := time.Now().Sub(t0); delta > timeout {
+				err = fmt.Errorf("after %d attepmts (during %d) last error: %s", i, delta, err.Error())
```
songbinliu: %v for time.Duration will print the time unit.

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,551 @@
...
+			return ref.Reference.Kind, ref.Reference.Name, nil
+		}
+	}
+
+	glog.V(4).Infof("cannot find pod-%v/%v parent by Annotations.", pod.Namespace, pod.Name)
+
+	return "", "", nil
+}
+
+func copyPodInfo(oldPod, newPod *api.Pod) {
```
songbinliu: I cannot explain why some fields are copied, some are not.
I tested them. It works now. I will keep a mind about this if there is some misbehavior of copied pod.

pkg/action/executor/rescheduler.go
```diff
@@ -172,6 +187,74 @@ func (r *ReScheduler) preActionCheck(podName, namespace, nodeName string) (*api.
...
+		return nil, addErrors(prefix, err, err2)
+	}
+
+	// restore: change back the schedulerName for controllers, and delete Pending Pod of the controllers
+	restore := func() {
+		//check it again in case somebody has changed it back.
+		if flag, _ := check(client, parentKind, nameSpace, parentName, noexist); flag {
+			retryDuring(defaultRetryMore, defaultTimeOut, defaultSleep, func() error {
+				update(client, nameSpace, parentName, preScheduler)
+				return nil
```
songbinliu: you are right.

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,363 @@
...
+	spec := oldPod.Spec
+	spec.Hostname = ""
+	spec.Subdomain = ""
+	spec.NodeName = ""
+
+	newPod.Spec = spec
+	return
+}
+
+func calcGracePriod(pod *api.Pod) int64 {
```
enlinxu: typo, period

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,363 @@
...
+	err := podClient.Delete(pod.Name, delOption)
+	if err != nil {
+		err = fmt.Errorf("move-failed: failed to delete original pod-%v: %v",
+			id, err)
+		glog.Error(err)
+		return nil, err
+	}
+
+	//3. create (and bind) the new Pod
+	time.Sleep(time.Duration(grace+1) * time.Second) //wait for the previous pod to be cleaned up.
```
enlinxu: what is the purpose of grace period here?

pkg/action/action_handler.go
```diff
@@ -144,6 +168,32 @@ func (h *ActionHandler) ExecuteAction(actionExecutionDTO *proto.ActionExecutionD
 	return result, nil
 }
 
+func keepAlive(tracker sdkprobe.ActionProgressTracker, stop chan struct{}) {
```
enlinxu: This Method should be at SDK level.

pkg/action/executor/move_util.go
```diff
@@ -0,0 +1,363 @@
...
+		if b1 != b2 {
+			return b1 - b2
+		}
+	}
+
+	return 0
+}
+
+//retry to execute a function with a timeout
+func retryDuring(attempts int, timeout time.Duration, sleep time.Duration, myfunc func() error) error {
```
enlinxu: This method could also be a generic util retry at SDK level.

pkg/action/supervisor/action_supervisor.go
```diff
@@ -92,21 +97,22 @@ func (s *ActionSupervisor) checkMoveAction(action *turboaction.TurboAction) (boo
...
-		return false, err
+	phase := targetPod.Status.Phase
+	if phase == api.PodRunning || (phase == api.PodPending && targetPod.DeletionGracePeriodSeconds == nil) {
+		moveDestination := moveSpec.Destination
+		actualHostingNode := targetPod.Spec.NodeName
+		if actualHostingNode == moveDestination {
+			glog.V(3).Infof("Move pod [%s/%s] action succeeded.", podNamespace, podName)
+			return true, nil
+		} else {
+			glog.Errorf("Move action failed. Incorrect move destination %s", actualHostingNode)
```
enlinxu: If this happens, does that mean the default scheduler scheduled the pod? do we try to move it again to the right node?

pkg/action/executor/move_util.go
songbinliu: The original Pod is deleted, but it will only disappear after gracePeriod seconds.  We will create a new Pod with the same nameSpace/name as the original one. So if we don't wait for the original Pod to disappear, the creation for the new Pod will fail immediately.

pkg/action/supervisor/action_supervisor.go
songbinliu: 1. how this can happen:  one case is that the desired node is down, and default scheduler moves the pod to another node.

2.whether to try again:  the environment (workloads, pods assignment, and cluster) may have changed since last move action, so it may not be right to move the pod again.  Will OpsMgr generate new recommendation for this Pod again?

pkg/action/executor/move_util.go
songbinliu: moved them to project level for now.

pkg/action/action_handler.go
songbinliu: this function currently is only useful for Move Action, and I am not sure whether it is right for all the actions (other actions may use ActionProgressTracker to report real progress). So it is subject to change in the future.

 In SDK, the discoeryClient has implemented a similar function to keep alive while waiting for the probe result.  For the actionWorker, the SDK does not implement the KeepAlive function in purpose, but implement the UpdageProgress() instead, to let the action handler to report its progress.

enlinxu: Can't we use updateProgress()?

songbinliu: yes, we are using the UpdateProgress in this function keepAlive().  Currently, actionHandler does not pass the ActionProgressTracker down the ActionExecutor(rescheduler, Scaler). So the UpdateProgress() is called in this keepAlive() function.

pkg/action/util/expiration_map.go
```diff
@@ -0,0 +1,165 @@
...
+		if item.isExpire() {
+			keys = append(keys, key)
+		}
+	}
+
+	return keys
+}
+
+//expireItem: call the callBack function, and delete the item
+func (s *ExpirationMap) expireItem(key string) bool {
```
DongyiYang: what's the returned bool used for?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -87,6 +92,8 @@ func (s *VMTServer) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.UseVMWare, "usevmware", false, "If the underlying infrastructure is VMWare.")
 	fs.UintVar(&s.KubeletPort, "kubelet-port", kubelet.DefaultKubeletPort, "The port of the kubelet runs on")
 	fs.BoolVar(&s.EnableKubeletHttps, "kubelet-https", kubelet.DefaultKubeletHttps, "Indicate if Kubelet is running on https server")
+	fs.StringVar(&s.K8sVersion, "k8sVersion", executor.HigherK8sVersion, "the kubernetes server version; for openshift, it is the underlying Kubernetes' version.")
```
DongyiYang: does this mean, user must provide the k8s version when it starts? is it possible to get the version  programmatically just like`kubectl version`

pkg/action/action_handler.go
```diff
@@ -144,6 +168,32 @@ func (h *ActionHandler) ExecuteAction(actionExecutionDTO *proto.ActionExecutionD
...
+		var progress int32 = 0
+		state := proto.ActionResponseState_IN_PROGRESS
+
+		for {
+			progress = progress + 1
+			if progress > 99 {
+				progress = 99
+			}
+
+			tracker.UpdateProgress(state, "in progress", progress)
```
songbinliu: Yea. It is currently for the only purpose to keep the action live, not showing the actual progress.

pkg/action/util/expiration_map.go
songbinliu: Currently no use. I may use it to count how many items have been successfully expired.

cmd/kubeturbo/app/kubeturbo_builder.go
songbinliu: It is possible to get the k8s cluster version. But I have not find the right API to get this info.

PR #84, opened 2017-08-11
deploy on Openshift with special image, fewer input

songbinliu:
(1) A dedicated docker image is for openshift. 
    With this image, fewer input is needed when to deploy kubeturbo in Openshift cluster.

(2) Since the new discovery framework gets data from kubelet, cAdvisor is not needed anymore.
    The instructions about deploy cAdvisor are deleted in the document.


songbinliu:
this is PR version of deployment for OpenShift.
https://github.com/songbinliu/kubeturbo/tree/openshiftDeployment/deploy/kubeturbo-for-openshift

PR #85, opened 2017-08-22
implement container-based supply chain

songbinliu:
supply chain is changed:
    vm -> pod -> application -> virtualApplication
                   to
    vm -> pod -> container -> application -> virtualApplication.

### DIFF
--- a/pkg/discovery/dtofactory/application_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/application_entity_dto_builder.go
@@ -18,7 +18,7 @@ import (
 
 const (
 	AppPrefix                  string  = "App-"
-	defaultTransactionCapacity float64 = 50.0
+	defaultTransactionCapacity float64 = 500.0
 )
 
 var (
@@ -140,23 +140,20 @@ func (builder *applicationEntityDTOBuilder) getTransactionUsedValue(pod *api.Pod
 // equally distribute Pod.Transaction.used to the hosted containers.
 func (builder *applicationEntityDTOBuilder) getAppTransactionUsage(index int, pod *api.Pod) float64 {
 	podTransactionUsage := builder.getTransactionUsedValue(pod)
-	if podTransactionUsage < 2.0 || len(pod.Spec.Containers) < 2 {
+	containerNum := len(pod.Spec.Containers)
+
+	// case1: if there is only one container, then it has all the transactions.
+	if containerNum < 2 {
 		return podTransactionUsage
 	}
 
-	containerNum := len(pod.Spec.Containers)
+	// case2: equally distribute transactions, the first container may have a little more
 	if containerNum < index {
 		glog.Errorf("potential bug: pod[%s] containerNum mismatch %d Vs. %d.", util.PodKeyFunc(pod), containerNum, index)
 		return 0.0
 	}
 
-	// case1: if there is only one container, then it has all the transactions.
-	if containerNum == 1 {
-		return podTransactionUsage
-	}
-
-	// case2: equally distribute transactions, the first container may have a little more
-	share := float64(int64(podTransactionUsage)/int64(containerNum))
+	share := float64(int64(podTransactionUsage) / int64(containerNum))
 	if index == 0 {
 		residue := (podTransactionUsage - (share * float64(containerNum)))
 		share += residue
--- a/pkg/discovery/dtofactory/application_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/application_entity_dto_builder.go
@@ -184,13 +184,6 @@ func (builder *applicationEntityDTOBuilder) getCommoditiesSold(appId string, ind
 	return result, nil
 }
 
-/*
-func FindTransactionCommodityKey(pod *api.Pod, clusterID string) string {
-	appType := util.GetAppType(pod)
-	return appType + "-" + clusterID
-}
-*/
-
 // Build the bought commodities by each application.
 // An application buys vCPU, vMem and Application commodity from a container.
 func (builder *applicationEntityDTOBuilder) getApplicationCommoditiesBought(appId, containerId string, cpuFrequency float64) ([]*proto.CommodityDTO, error) {
--- a/pkg/discovery/dtofactory/application_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/application_entity_dto_builder.go
@@ -61,9 +61,10 @@ func (builder *applicationEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]
 	var result []*proto.EntityDTO
 
 	for _, pod := range pods {
+		podFullName := util.GetPodClusterID(pod)
 		nodeCPUFrequency, err := builder.getNodeCPUFrequency(pod)
 		if err != nil {
-			glog.Errorf("failed to build ContainerDTOs for pod[%s]: %v", pod.Name, err)
+			glog.Errorf("failed to build ContainerDTOs for pod[%s]: %v", podFullName, err)
 			continue
 		}
 		podId := string(pod.UID)
@@ -72,7 +73,7 @@ func (builder *applicationEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]
 			//container := &(pod.Spec.Containers[i])
 			containerId := util.ContainerIdFunc(podId, i)
 			appId := util.ApplicationIdFunc(containerId)
-			displayName := AppPrefix + util.GetPodClusterID(pod)
+			displayName := AppPrefix + podFullName
 
 			ebuilder := sdkbuilder.NewEntityDTOBuilder(proto.EntityDTO_APPLICATION, appId).
 				DisplayName(displayName)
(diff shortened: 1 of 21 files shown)


pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -43,130 +42,184 @@ func NewApplicationEntityDTOBuilder(sink *metrics.EntityMetricSink) *application
...
+		return podTransactionUsage
 	}
-	clusterCommodityKey, ok := clusterInfo.GetValue().(string)
-	if !ok {
-		return nil, errors.New("Failed to get cluster ID")
+
+	//1. get pod CPU usage
+	podkey := util.PodKeyFunc(pod)
+	podUsage := builder.getCPUUsedValue(task.PodType, podkey)
+	if podUsage < 0.001 {
```
enlinxu: Why we need this?

pkg/discovery/dtofactory/container_dto_builder.go
```diff
@@ -0,0 +1,162 @@
...
+			commoditiesBought, err := builder.getCommoditiesBought(podId, containerId, nodeCPUFrequency)
+			if err != nil {
+				glog.Errorf("failed to create commoditiesBought for container[%s]: %v", name, err)
+				continue
+			}
+			provider := sdkbuilder.CreateProvider(proto.EntityDTO_CONTAINER_POD, podId)
+			ebuilder.Provider(provider).BuysCommodities(commoditiesBought)
+
+			//3. set properties
+			if !util.Monitored(pod) {
```
enlinxu: We may want to review this with M2 team. The reason why we set monitored false for some pod is because we don't want our market to MOVE them. But we might still want to be able to resize the container inside the pod. Also now there're new flags (moveable, resizable, etc.), we might not need to use Monitored to disable move of pod. Please ask Wei to confirm.

pkg/discovery/dtofactory/container_dto_builder.go
```diff
@@ -0,0 +1,162 @@
...
+	attributeSetter.Add(func(commBuilder *sdkbuilder.CommodityDTOBuilder) { commBuilder.Resizable(true) }, metrics.CPU, metrics.Memory)
+
+	commodities, err := builder.getResourceCommoditiesBought(task.ContainerType, containerId, commodityBought, converter, attributeSetter)
+	if err != nil {
+		return nil, err
+	}
+	result = append(result, commodities...)
+
+	//2. VMPMAccess
+	podAccessComm, err := sdkbuilder.NewCommodityDTOBuilder(proto.CommodityDTO_VMPM_ACCESS).
```
enlinxu: We might be able to use Moveable flag to control this. Again, let's talk to Wei.

pkg/discovery/dtofactory/application_entity_dto_builder.go
songbinliu: To distribute transaction usage among the containers.
Currently, we can only get the transaction usage in Pod level.  If pod has more than 1 containers, then we should find a way to distribute the transaction among the containers. Here, I use container-cpu usage as a hint to do the distribution.  The more cpu is used by a container, the more share of transactions will be given to that container.

pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -18,7 +17,8 @@ import (
 )
 
 const (
-	AppPrefix string = "App-"
+	AppPrefix                 string = "App-"
+	defaultTransactionCapaciy        = float64(50.0)
```
DongyiYang: why not 
```go 
defaultTransactionCapacity float64=50.0
```
also, there is typo in the name.

pkg/discovery/monitoring/master/cluster_monitor.go
```diff
@@ -14,17 +14,19 @@ import (
 	"github.com/golang/glog"
 )
 
+// This monitor is based on Kubenetes's Node, Pod and Container settings;
+// it will mainly generate CPU/Memory commodity's Capacity and Provision for Node, Pod, and Container;
 type ClusterMonitor struct {
 	config *ClusterMonitorConfig
 
+	//TODO: since this sink is not accessed by multiple goroutines
```
songbinliu: help to confirm that this **sink** is not accessed concurrently, so the synchronized Add() is not necessary.

pkg/discovery/dtofactory/property/application_properties.go
```diff
@@ -36,6 +37,15 @@ func BuildAppProperties(podNamespace, podName string) []*proto.EntityDTO_EntityP
 	}
 	properties = append(properties, nameProperty)
 
+	containerIndexName := appPropertyNameHostingContainerIndex
+	containerIndexValue := string(index)
```
songbinliu: string(index) will return an empty string "".
should use fmt.Sprintf("%d", index)

pkg/discovery/dtofactory/application_entity_dto_builder.go
enlinxu: Thanks Dongyi. :)

pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -43,130 +42,184 @@ func NewApplicationEntityDTOBuilder(sink *metrics.EntityMetricSink) *application
...
-		displayName := AppPrefix + util.GetPodClusterID(pod)
-		entityDTOBuilder.DisplayName(displayName)
 
-		// commodity sold
-		commoditiesSold, err := builder.getApplicationCommoditiesSold(pod)
+	for _, pod := range pods {
+		nodeCPUFrequency, err := builder.getNodeCPUFrequency(pod)
 		if err != nil {
-			glog.Errorf("Failed to create commodities sold by application %s: %s", displayName, err)
+			glog.Errorf("failed to build ContainerDTOs for pod[%s]: %v", pod.Name, err)
```
DongyiYang: Here I suggest to print out not only `pod.Name`, but also `pod.Namespace`. So maybe use `util.GetPodClusterID(pod)`.

pkg/discovery/dtofactory/application_entity_dto_builder.go
DongyiYang: Maybe you can distinguish different container by different port number. But I don't know how much effort would it be.

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -62,17 +62,22 @@ func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]*proto.E
 		// display name.
 		displayName := util.GetPodClusterID(pod)
 		entityDTOBuilder.DisplayName(displayName)
+		cpuFrequency, err := builder.getNodeCPUFrequency(pod)
+		if err != nil {
+			glog.Errorf("failed to build pod[%] EntityDTO: %v", displayName, err)
```
DongyiYang: should be `%s`.

pkg/discovery/dtofactory/application_entity_dto_builder.go
songbinliu: I think distributing transactions based on CPU is better.

The reason is that:
    (1) Some sidecar containers may not expose Port. For example, a sidecar container only push logs onto a log server;
     (2) Some sidecar containers may be very important.  For example, there are two containers, one is for network interaction; the other is responsible for computation, and sends the results back to the first one. In this case, it seems no outbound network activity for the second container.  But we'd better to make sure it has enough resources.

In these assumptions, I seems that

pkg/discovery/dtofactory/container_dto_builder.go
songbinliu: Currently, we cannot resize a static pod, or a pod controlled by DaemonSets yet.  They are not scheduled as the other kinds of pods.

pkg/discovery/dtofactory/application_entity_dto_builder.go
songbinliu: After a discussion with Enlin, we can simply equally distributed the Pod transactions to the hosted containers.  It does not matter which container get more transactions, since transaction is used to drive Pod to be cloned, not the containers.

pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -18,7 +17,8 @@ import (
 )
 
 const (
-	AppPrefix string = "App-"
+	AppPrefix                  string  = "App-"
+	defaultTransactionCapacity float64 = 50.0
```
enlinxu: let's make the default Transaction Capacity 500. 50 is more like a number we used for testing (to get provision action easier). In the future it should be configurable from the policy tab by user. It's a app by app scenario.

pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -43,130 +42,165 @@ func NewApplicationEntityDTOBuilder(sink *metrics.EntityMetricSink) *application
...
+	if containerNum < index {
+		glog.Errorf("potential bug: pod[%s] containerNum mismatch %d Vs. %d.", util.PodKeyFunc(pod), containerNum, index)
+		return 0.0
 	}
-	clusterCommodityKey, ok := clusterInfo.GetValue().(string)
-	if !ok {
-		return nil, errors.New("Failed to get cluster ID")
+
+	// case1: if there is only one container, then it has all the transactions.
+	if containerNum == 1 {
```
enlinxu: Isn't this already checked in the beginning of the method?

PR #88, opened 2017-08-28
refactor kubeletClient for actionHandler.resize

songbinliu:
1. Promote kubeletClient to a higher place (VMT Config.KubeletClient):
    So that kubeletClient can be used in package Discovery as well as in package ActionHander.
    For resize(CPU capacity) action,  we have to convert CPU Capacity in frequency into CPU cores.
    **kubeletClient is needed to get node CPU frequency to make this conversion possible.**  This is the main reason for this refactor.

2. Refactor KubeletClient: 
     change dependency from "k8s.io/Kubernetes" to "k8s.io/client-go".  I think this will make it easier to maintain the dependency.

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -62,7 +62,7 @@ type VMTServer struct {
 	UseVMWare bool
 
 	// Kubelet related config
-	KubeletPort        uint
+	KubeletPort        int
 	EnableKubeletHttps bool
 
 	// for Move Action
@@ -90,7 +90,7 @@ func (s *VMTServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.KubeConfig, "kubeconfig", s.KubeConfig, "Path to kubeconfig file with authorization and master location information.")
 	fs.BoolVar(&s.EnableProfiling, "profiling", false, "Enable profiling via web interface host:port/debug/pprof/.")
 	fs.BoolVar(&s.UseVMWare, "usevmware", false, "If the underlying infrastructure is VMWare.")
-	fs.UintVar(&s.KubeletPort, "kubelet-port", kubelet.DefaultKubeletPort, "The port of the kubelet runs on")
+	fs.IntVar(&s.KubeletPort, "kubelet-port", kubelet.DefaultKubeletPort, "The port of the kubelet runs on")
 	fs.BoolVar(&s.EnableKubeletHttps, "kubelet-https", kubelet.DefaultKubeletHttps, "Indicate if Kubelet is running on https server")
 	fs.StringVar(&s.K8sVersion, "k8sVersion", executor.HigherK8sVersion, "the kubernetes server version; for openshift, it is the underlying Kubernetes' version.")
 	fs.StringVar(&s.NoneSchedulerName, "noneSchedulerName", executor.DefaultNoneExistSchedulerName, "a none-exist scheduler name, to prevent controller to create Running pods during move Action.")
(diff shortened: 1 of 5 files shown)


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -110,29 +110,42 @@ func createRecorder(kubecli *kubernetes.Clientset) record.EventRecorder {
 	return eventBroadcaster.NewRecorder(scheme.Scheme, apiv1.EventSource{Component: "kubeturbo"})
 }
 
-func (s *VMTServer) createKubeConfig() (*restclient.Config, error) {
+func (s *VMTServer) createKubeConfigOrDie() *restclient.Config {
 	kubeConfig, err := clientcmd.BuildConfigFromFlags(s.Master, s.KubeConfig)
 	if err != nil {
-		glog.Errorf("Error getting kubeconfig:  %s", err)
-		return nil, err
+		panic(fmt.Sprintf("Error getting kubeconfig:  %s", err))
```
DongyiYang: why not use 
```go
glog.Fatalf()
```
https://godoc.org/github.com/golang/glog#Fatalf

pkg/discovery/monitoring/kubelet/kubelet_client.go
```diff
@@ -2,46 +2,44 @@ package kubelet
...
+// Since http.Client is thread safe (https://golang.org/src/net/http/client.go)
+// KubeletClient is also thread-safe if concurrent goroutines won't change the fields.
+type KubeletClient struct {
+	client *http.Client
+	scheme string
+	port   int
 }
 
-func (kc *kubeletClient) GetSummary(host Host) (*stats.Summary, error) {
+func (kc *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
```
DongyiYang: Maybe change the parameter name from `host` to `ip`, because only IP address is passed in. Same for other methods.

cmd/kubeturbo/app/kubeturbo_builder.go
songbinliu: I was not sure which is better. But I saw k8s is using [panic](https://github.com/kubernetes/apimachinery/blob/dc1f89aff9a7509782bde3b68824c8043a3e58cc/pkg/api/resource/quantity.go#L136).  

I think it is better to call glog.Error() and exit. This is more grace.

pkg/discovery/monitoring/kubelet/kubelet_client.go
songbinliu: It is possible to get information via hostName. I think it is more natural to use hostName in kubernetes cluster, because when deploying kubernetes, we provides a hostName list.

enlinxu: Host name is provided by the cluster admin while provisioning the cluster, it might be more reliable than IP, which is derived from the host name.

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -62,7 +62,7 @@ type VMTServer struct {
 	UseVMWare bool
 
 	// Kubelet related config
-	KubeletPort        uint
+	KubeletPort        int
```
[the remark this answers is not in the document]
songbinliu: I looked the code of Heapster and [Kubernetes](https://github.com/songbinliu/kubernetes/blob/6f7eac63c2d170100ef95f65aa2d57f9812243f4/pkg/apis/componentconfig/types.go#L201). They are all using Int32 as port type.

I guess the reason is that uint is more problemsome. For example, if we use uint as port number, someone may want to convert a Integer value (-10) to uint (4294967286), and assign to the port. This will cause more undetectable problem.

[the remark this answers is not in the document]
DongyiYang: I think the port number cannot be larger than 65536. Don't know why `int32` is used there.

PR #89, opened 2017-08-28
Refine Readme for deploy steps

enlinxu:
Refine the readme file for openshift deployment.

### DIFF
--- a/deploy/kubeturbo-for-openshift/README.md
+++ b/deploy/kubeturbo-for-openshift/README.md
@@ -3,10 +3,10 @@
 This guide is about how to deploy **Kubeturbo** service in **OpenShift**.
 
 ### Step One: Label Master Node
-As Kubeturbo is suggested to run on the master node, we need to create label for the Master node. To label the master node, simply execute the following command
+As Kubeturbo is suggested to run on the master nodes, please create a label to represent the master nodes, if there's no existed one. For example:
 
 ```console
-$oc label nodes <MASTER_NODE_NAME> role=kubeturbo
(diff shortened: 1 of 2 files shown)


deploy/kubeturbo-for-openshift/README.md
```diff
@@ -33,7 +33,7 @@ Create a file called **"config"** and put it under */etc/kubeturbo/*.
 {
 	"communicationConfig": {
 		"serverMeta": {
-			"turboServer": "<SERVER_ADDRESS>"
+			"turboServer": "https://<SERVER_IP_ADDRESS>:443"
```
songbinliu: Should also change this note in the previous paragraph:
     **NOTE**: Turbonomic server address is "**IP address of your ops manager**".

It should be a HTTP Host, with host name and port.  Will it be true that OpsMgr will always use port 443 on HTTPS?   In my local deployment, the turboServer is **https://10.10.200.97:9400**.


 


For example

DongyiYang: Don't know if 9400 or 8400 are needed to be mentioned since people won't use them in a real turbo instance.

songbinliu: Previously, OpsMgr runs on HTTP, and the default port for OpsMgr is 80, which can be omitted. But in a recent build, Opsmgr runs on HTTPS, and the default port is 443.  In this case, port number cannot be omitted in the test.

songbinliu: I tested with https on port 443.  It is ok even if we do not specify the port number.
So we can commit the port number.

PR #93, opened 2017-09-06
resize container cpu/memory capacity

songbinliu:


pkg/action/action_handler.go
```diff
@@ -22,11 +22,12 @@ import (
 )
 
 const (
-	defaultActionCacheTTL = time.Second * 50
+	defaultActionCacheTTL = time.Second * 100
 )
 
 type ActionHandlerConfig struct {
 	kubeClient     *client.Clientset
+	kubeletClient  *kubelet.KubeletClient
```
DongyiYang: Since not every action handler needs `kubeletClient`, so maybe to make `ActionHandlerConfig` an interface is a better idea.

pkg/action/action_handler.go
```diff
@@ -112,6 +114,9 @@ func (h *ActionHandler) registerActionExecutors() {
 	horizontalScaler := executor.NewHorizontalScaler(h.config.kubeClient, h.config.broker, h.scheduler)
 	h.actionExecutors[turboaction.ActionProvision] = horizontalScaler
 	h.actionExecutors[turboaction.ActionUnbind] = horizontalScaler
+
+	containerResizer := executor.NewContainerResizer(h.config.kubeClient, h.config.kubeletClient, h.config.broker, h.config.k8sVersion, h.config.noneSchedulerName, h.lockMap)
```
DongyiYang: If you make `ActionHandlerConfig` an interface, you can do something like
```
containerResizer := executor.NewContainerResizer(resizeHandlerConfig, h.lockMap)
```
Just don't passed in so many parameters.

pkg/action/executor/move_util.go
```diff
@@ -79,9 +79,10 @@ func movePod(client *kclient.Clientset, pod *api.Pod, nodeName string, retryNum
 	}
 
 	//3. create (and bind) the new Pod
-	time.Sleep(time.Duration(grace+1) * time.Second) //wait for the previous pod to be cleaned up.
-	du := time.Duration(grace+3) * time.Second
-	err = util.RetryDuring(retryNum, du*time.Duration(retryNum), defaultSleep, func() error {
+	time.Sleep(time.Duration(grace)*time.Second + defaultMoreGrace) //wait for the previous pod to be cleaned up.
+	interval := defaultPodCreateSleep
+	timeout := interval*time.Duration(retryNum) + time.Second*10
```
DongyiYang: define `time.Second* 10` as a constant

pkg/action/executor/move_util.go
```diff
@@ -128,18 +129,23 @@ type moveHelper struct {
...
 	key     string
 	version int64
+
+	//stop Renewing
+	stop       chan struct{}
+	isRenewing bool
 }
 
-func NewMoveHelper(client *kclient.Clientset, nameSpace, name, kind, parentName, noneScheduler string, highver bool) (*moveHelper, error) {
+func NewMoveHelper(client *kclient.Clientset, nameSpace, name, kind, parentName, noneScheduler string, highver bool) (*schedulerHelper, error) {
```
DongyiYang: if the type has been changed from `moveHelper` to `schedulerHelper`, so please make the function name consistent, change `NewMoveHelper` to `NewScheudler` helper. BTW, I think `reschedulerHelper` makes more sense.

pkg/action/executor/move_util.go
```diff
@@ -295,10 +312,42 @@ func (h *moveHelper) lockCallBack() {
...
+	if interval < time.Second {
+		interval = time.Second
+	}
+	h.isRenewing = true
+
+	go func() {
+		for {
+			select {
+			case <-h.stop:
+				glog.V(2).Infof("moveHelper stop renewlock.")
```
DongyiYang: change log, `moveHelper` -> `schedulerHelper`.

pkg/action/executor/resize_container.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+// get node cpu frequency, in KHz;
+func (r *ContainerResizer) getNodeCPUFrequency(host string) (uint64, error) {
+	return r.kubeletClient.GetMachineCpuFrequency(host)
+}
+
+func (r *ContainerResizer) setCPUCapacity(cpuMhz float64, host string, rlist k8sapi.ResourceList) error {
+	cpuFrequency, err := r.getNodeCPUFrequency(host)
```
DongyiYang: Do we need to call API every time we execute `setCPUCapacity`? I think this meta data can be store in a map and frequency of CPU is not likely to change so it is more likely a one time task.

pkg/action/action_handler.go
songbinliu: I think interface will not improve code readability, nor runtime performance.  Let's keep it simple.

pkg/action/executor/move_util.go
songbinliu: It is a tool to manipulate the original scheduler of pod: getScheduler/updateScheduler, not really rescheduling a pod.

pkg/action/executor/resize_container.go
songbinliu: Yea,  right, node meta data should be cached.  I'd like to submit another PR to add cache for Node and pod.

Issue #104, opened 2017-09-12
[bug] resizing/moving actions may fail in concurrent situations 


songbinliu:
In Kubernetes, there may be some bare Pods, which are created directly by users, and are not controlled by Replication Controllers.  In current implementation of [Move](https://github.com/turbonomic/kubeturbo/blob/abf0bf397d0d970295c4f64081cf2e067dd7f7eb/pkg/action/executor/rescheduler.go#L274) and [Resize](https://github.com/turbonomic/kubeturbo/blob/abf0bf397d0d970295c4f64081cf2e067dd7f7eb/pkg/action/executor/resize_container.go#L215), there is no concurrent control of the actions on the same bare Pod. 

However, as each of the Move or Resizing action may take from serval seconds to minutes, different operations on the same Pod may overlap with each other with a very high chance. For example, one action is to **Move** a Pod, one action is to **Resize** the CPU capacity, and another action is to **Resize** the Memory capacity. These three actions may overlap, and result in actions failure.

songbinliu:
I plan to add a lock Item to the lockMap before moving/resizing the bare Pod.

For Pods controlled by ReplicationControllers, it seems no necessary to use lock; because they would require a lock of the parent Controller.

PR #106, opened 2017-09-13
add lock for actions on bare pods

songbinliu:
This PR is to address [issue8](https://github.com/turbonomic/kubeturbo/issues/104)

To avoid concurrent contention of actions on bare pod, a concurrent control lock is added for the 
bare pod.


pkg/action/executor/rescheduler.go
```diff
@@ -289,3 +289,29 @@ func (r *ReScheduler) reSchedule(action *turboaction.TurboAction) (*turboaction.
 
 	return action, nil
 }
+
+// as there may be concurrent actions on the same bare pod:
+//   one action is to move Pod, and the other is to Resize Pod.container;
+// thus, concurrent control should also be applied to bare pods.
+func (r *ReScheduler) moveBarePod(pod *api.Pod, nodeName string) (*api.Pod, error) {
+	podkey := util.BuildIdentifier(pod.Namespace, pod.Name)
```
enlinxu: What if user has added more than one kubernetes clusters in Turbo, and in these clusters, there's duplicate pod.namespace/pod.name?

songbinliu: yes, this can be a problem in multiple clusters scenarios.

But because of similar constrains,  Kubeturbo won't work for multiple clusters during the discovery process.  So let's refactor the code to support multiple clusters in future.

enlinxu: Agree. And this might be a high priority case. (A simple solution would be we add target identifier as a prefix also)

Issue #110, opened 2017-09-27
HorizontalScaler is using turbo.scheduler to schedule new Pod, which dose not work.


songbinliu:
For horizontal scaling, [HorizontalScaler](https://github.com/turbonomic/kubeturbo/blob/9855fa0945724a50088422ed383be90f5a825ada/pkg/action/executor/horizontal_scaler.go#L28) is using [TurboScheduler](https://github.com/turbonomic/kubeturbo/blob/9855fa0945724a50088422ed383be90f5a825ada/pkg/scheduler/turbo_scheduler.go#L31) to [schedule the new Pod](https://github.com/turbonomic/kubeturbo/blob/9855fa0945724a50088422ed383be90f5a825ada/pkg/action/executor/horizontal_scaler.go#L272). 

Since the new Pods (in case of scaling up) will be scheduled `default scheduler`,  the schedule action by `TurboScheduler` will always fail, and hence the Scaling action will fail.

songbinliu:
One solution is:  let `Kubeturbo` not try to schedule the new Pods, but let the `default scheduler` to schedule the new Pods.   `Kubeturbo` only do one thing: modify the number of replicas.

songbinliu:
The history of this issue is that previously, it was assumed that there is only one scheduler-- `TuborScheduler`.  However, we later changed this assumption, and let `TurboScheduler` coexist with `default scheduler`, but forget to update the implementation of `HorizontalScaler`.

songbinliu:
This is solved by [PR113](https://github.com/turbonomic/kubeturbo/pull/113)

Issue #96, opened 2017-09-07
[bug] when sending several action results. 


songbinliu:
**Problem**   
When more than one actions are executed concurrently,  _kubeturbo_ may report the action results unordered, which means that results may mismatch the requests.
  
**Example**
     There are two actions, one is to move a **MovePod**, the other is to **ResizePod**, and both actions are failed. But the reported reasons mismatch the requests.

![screen shot 2017-09-07 at 10 18 17 am](https://user-images.githubusercontent.com/27221807/30181957-0c8eb3e2-93e3-11e7-9be5-66501d317446.png)

songbinliu:
After reading the code of pkg/action/action_handler.go, I find the reason is that [_ActionHandler.ExecuteAction()_](https://github.com/turbonomic/kubeturbo/blob/575e6dd31d26d3443adb5478c1829e38095f9694/pkg/action/action_handler.go#L165)  can be called concurrently, and _ActionHandler.resultChan_ can be read/written concurrently. However, _ActionHandler.ExecuteAction()_ assumes that the content in  _ActionHandler.resultChan_ should be result of current action, which can not be guaranteed.

songbinliu:
This is solved by [PR113](https://github.com/turbonomic/kubeturbo/pull/113).

PR #111, opened 2017-09-29
fix the bug in horizontal scaler

songbinliu:
This PR is for [Issue110](https://github.com/turbonomic/kubeturbo/issues/110):

1. don't care the scheduling of new Pod.
    Only update the replica num, and let default scheduler to schedule the new Pods;

2. send actions result directly to ActionHandler.
     To avoid the [mismatch action results report](https://github.com/turbonomic/kubeturbo/issues/96)
    
Note: this PR won't be merged. Plan to fix the two issues [Issue110](https://github.com/turbonomic/kubeturbo/issues/110) and [Issue96](https://github.com/turbonomic/kubeturbo/issues/96) together.

### DIFF
--- a/pkg/action/action_handler.go
+++ b/pkg/action/action_handler.go
@@ -4,16 +4,12 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
 	client "k8s.io/client-go/kubernetes"
 
 	"github.com/turbonomic/kubeturbo/pkg/action/executor"
-	"github.com/turbonomic/kubeturbo/pkg/action/supervisor"
 	"github.com/turbonomic/kubeturbo/pkg/action/turboaction"
 	"github.com/turbonomic/kubeturbo/pkg/action/util"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/kubelet"
-	turboscheduler "github.com/turbonomic/kubeturbo/pkg/scheduler"
-	"github.com/turbonomic/kubeturbo/pkg/turbostore"
 
 	sdkprobe "github.com/turbonomic/turbo-go-sdk/pkg/probe"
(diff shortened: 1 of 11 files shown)


pkg/action/action_handler.go
```diff
@@ -265,17 +165,61 @@ func getActionTypeFromActionItemDTO(actionItem *proto.ActionItemDTO) (turboactio
...
+	return &proto.ActionResult{
+		Response: res,
 	}
-	handler.resultChan <- result
+}
+
+func (h *ActionHandler) failedResult(msg string) *proto.ActionResult {
+
+	state := proto.ActionResponseState_FAILED
+	progress := int32(0)
```
enlinxu: It doesn't make sense to explicitly set the progress to 0 if the action has reached to final state (regardless of succeed or fail)

pkg/action/action_handler.go
```diff
@@ -265,17 +165,61 @@ func getActionTypeFromActionItemDTO(actionItem *proto.ActionItemDTO) (turboactio
...
+		Progress:            &progress,
+		ResponseDescription: &msg,
+	}
+
+	return &proto.ActionResult{
+		Response: res,
+	}
+}
+
+func keepAlive(tracker sdkprobe.ActionProgressTracker, stop chan struct{}) {
```
enlinxu: I still think the keepAlive method should be at SDK level, and the implementation here is definitely not K8s specific.

pkg/action/action_handler.go
songbinliu: so, set it to 100, or -1?

enlinxu: We can just remove the line.

songbinliu:
This PR is replaced by [PR113](https://github.com/turbonomic/kubeturbo/pull/113).

Issue #95, opened 2017-09-07
The IP_address of Move action may be improper sometimes.


songbinliu:
**Problem**:
     when kubeturbo executing a move action for Container_POD, it fails with the error log: "Cannot find node with IPs 10.128.0.1".

**Env**:
     Openshift Cluster 3.4
     OpsMgr 6.0.0, with Market1.
     Kubeturbo [PR93](https://github.com/turbonomic/kubeturbo/pull/93)  

**Possible Reason**
    The _newSE_  field of the **ActionItem** is a virtual machine: enMaster, whose official IP address is: 10.10.173.119, and Kubernetes uses this IP address to find the host.  However, the **ActionItem** only has an IP of "10.128.0.1".  

 So this maybe because the vCenter probe fails to get the correct IP address of the virtual machine. (stitching type is "uuid" in kubeturbo, and the stitching has successed).

<img width="839" alt="screen shot 2017-09-07 at 1 33 27 pm" src="https://user-images.githubusercontent.com/27221807/30176773-9d7ca632-93d1-11e7-9d2a-206d5f4165f7.png">

songbinliu:
Here is the content of the move action:
```
[...]

PR #113, opened 2017-10-12
major update of action.executor

songbinliu:
This PR will fix following issues about **Action Execution**:

(10 [Issue96 multiple responses mismatch the requests](https://github.com/turbonomic/kubeturbo/issues/96) 

**Solution**: Each action checks its result, and send the response by itself. Removed the _global result checking goroutine_ and the _global sending result goroutine_ of [ActionSupervisor](https://github.com/turbonomic/kubeturbo/blob/93dc4e87fb9c7150ba18c4b06caff8c252d8a201/pkg/action/supervisor/action_supervisor.go#L26).

(2) [Issuse110 using turbo.scheduler to schedule pod](https://github.com/turbonomic/kubeturbo/issues/110)

**Solution**: only update the replicaNum of ReplicaController/ReplicaSet/Deployment; 
let these controllers to create the new pods, and the default scheduler will schedule the new Pods;

(3) [Issue95 IP_address of Move action may be improper sometimes](https://github.com/turbonomic/kubeturbo/issues/95) 

**Solution**: Find the new hosting node by node.SystemUID or IP according to the stitching type.


pkg/action/action_handler.go
```diff
@@ -54,186 +56,92 @@ func NewActionHandlerConfig(kubeClient *client.Clientset, kubeletClient *kubelet
...
 	if err != nil {
-		glog.Errorf("Failed to execute action: %s", err)
-		msg := fmt.Sprintf("Action %s on %s failed.", actionType, actionItem.GetTargetSE().GetEntityType())
-		h.sendActionResult(proto.ActionResponseState_FAILED, int32(0), msg)
-		return
-	}
-	if action.Status == turboaction.Success {
-		h.sendActionResult(proto.ActionResponseState_SUCCEEDED, int32(100), "Success")
-		return
+		msg := fmt.Errorf("Action %s on %s failed.", actionType, actionItem.GetTargetSE().GetEntityType())
```
enlinxu: Why are we sending failure to Turbo when it's a success status?

PR #117, opened 2017-10-26
Add features to support non-disruptive actions

maxwangvmt:
Add features to support non-disruptive actions for container resize and pod move for ReplicationController and ReplicaSet. If the controller replica size is one, then perform the operations to enable non-disruptive action execution. The process is as follows:
- Update the replica to 2 (new pod p1 created by controller)
- Perform the action (resize or move)
- Delete the pod p1 (new pod p2 created by controller)
- Update the replica to 1 (pod p2 deleted by controller)

### DIFF
--- a/pkg/action/executor/nondisruptive_helper.go
+++ b/pkg/action/executor/nondisruptive_helper.go
@@ -1,0 +1,187 @@
+package executor
+
+import (
+	"fmt"
+	"github.com/golang/glog"
+	autil "github.com/turbonomic/kubeturbo/pkg/action/util"
+	"github.com/turbonomic/kubeturbo/pkg/util"
+	kclient "k8s.io/client-go/kubernetes"
+)
+
+// Performs operations for non-disruptive action (container resize and pod move).
+// If there is only one child pod for the associated controller, the process is:
+//  - update the replicas to 2 (pod p1 created by controller)
+//  - perform the action on the target object
+//  - delete p1 (pod p2 created by controller)
+//  - update the replicas to 1 (p2 deleted by controller)
+type NonDisruptiveHelper struct {
+	namespace string
+	contKind  string
+	contName  string
+	targetPod string
+	client    *kclient.Clientset
+
+	podToClean string
+}
+
+func NewNonDisruptiveHelper(client *kclient.Clientset, namespace, contKind, contName, targetPod string) *NonDisruptiveHelper {
+	p := &NonDisruptiveHelper{
+		namespace: namespace,
+		contKind:  contKind,
+		contName:  contName,
+		targetPod: targetPod,
+		client:    client,
+	}
+
+	return p
+}
+
+// Performs operations before executing a disruptive action (container resize and pod move).
+// It will update the controller replicas to 2 and save the new pod created by controller internally
+// for cleanup after the action is performed.
+func (h *NonDisruptiveHelper) OperateForNonDisruption() error {
+	replicas, _, err := autil.GetReplicaStatus(h.client, h.namespace, h.contName, h.contKind)
+	if err != nil {
+		glog.Errorf("Error getting replica status for %s %s/%s: %s", h.contKind, h.namespace, h.contName, err)
+		return fmt.Errorf("failed operations for non-disruptive action")
+	}
+
+	if replicas != 1 {
+		return nil
+	}
+
+	glog.V(2).Infof("Started non-disruptive operations on %s %s/%s", h.contKind, h.namespace, h.contName)
+
+	// Update replicas to 2
+	newReplicas := replicas + 1
+	glog.V(4).Infof("Updating replicas to %d", newReplicas)
+	if err := h.updateReplicaWithRetry(newReplicas); err != nil {
+		glog.Errorf("Error updating replica to 2 for %s %s/%s: %s", h.contKind, h.namespace, h.contName, err)
+		return fmt.Errorf("failed operations for non-disruptive action")
+	}
+
+	// Get the new pod
+	glog.V(4).Infof("Getting the new pod")
+	pods, err := autil.FindPodsByController(h.client, h.namespace, h.contName, h.contKind)
+	if err != nil {
+		glog.Errorf("Error finding pods by %s %s/%s: %s", h.contKind, h.namespace, h.contName, err)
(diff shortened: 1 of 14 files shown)


pkg/action/util/util.go
```diff
@@ -318,3 +325,123 @@ func GetPodGrandInfo(kclient *client.Clientset, pod *api.Pod) (string, string, e
...
+
+	pods := []*api.Pod{}
+
+	for _, pod := range podList.Items {
+		_, name, err := GetPodParentInfo(&pod)
+		if err != nil {
+			continue
+		}
+		if name == contName {
+			pod := pod
```
songbinliu: this is confusing `pod := pod`. Alternatively, traverse the podList by index:
```go
 for i := range podList.Items {
    pod := &(podList.Items[i])

    ...
   pods = append(pods, pod)
}
```

pkg/action/util/util.go
```diff
@@ -318,3 +325,123 @@ func GetPodGrandInfo(kclient *client.Clientset, pod *api.Pod) (string, string, e
...
+
+// Updates replica of a controller.
+// Currently, it supports replication controllers, replica sets, and deployments
+func UpdateReplicas(kubeClient *client.Clientset, namespace, contName, contKind string, newReplicas int) error {
+	newValue := int32(newReplicas)
+	switch contKind {
+	case util.KindReplicationController:
+		cont, err := kubeClient.CoreV1().ReplicationControllers(namespace).Get(contName, metav1.GetOptions{})
+		if err != nil {
+			return err
```
songbinliu: log the error here, so it will be easier for debug

songbinliu: This function is similar to those in [action/executor/hscale_helper.go](https://github.com/turbonomic/kubeturbo/blob/a682e010ee5a4d9c5cf60b3c3e24f38b3a2bb173/pkg/action/executor/hscale_helper.go#L121)

pkg/action/executor/resize_container.go
```diff
@@ -248,6 +248,19 @@ func (r *ContainerResizer) resizeControllerContainer(pod *k8sapi.Pod, parentKind
 	glog.V(3).Infof("resizeContainer [%s]: got lock for parent[%s]", id, parentName)
 	helper.KeepRenewLock()
 
+	// Performs operations for actions to be non-disruptive (when the pod is the only one for the controller)
+	// NOTE: It doesn't support the case of the pod associated to Deployment in version lower than 1.6.0.
+	//       In such case, the action execution will fail.
+	contKind, contName, err := util.GetPodGrandInfo(r.kubeClient, pod)
```
songbinliu: `SchedulerHelper` has pod's parent name and kind, its grandParent info may be get through on its parent's info. (one less API call)

pkg/action/executor/nondisruptive_helper.go
```diff
@@ -0,0 +1,174 @@
...
+	err = autil.DeletePod(h.client, h.namespace, h.podToClean)
+
+	if err != nil {
+		glog.Errorf("Error while deleting pod %s: %s", h.podToClean, err)
+	}
+
+	// Set replica = 1
+	glog.V(4).Infof("Updating replicas to 1")
+	if err := h.updateReplicaWithRetry(1); err != nil {
+		glog.Errorf("Error updating replica to 1 for %s %s/%s: %s", h.contKind, h.namespace, h.contName, err)
```
songbinliu: add log indicating the "**clean up failed**"

pkg/action/executor/resize_container.go
```diff
@@ -248,6 +248,19 @@ func (r *ContainerResizer) resizeControllerContainer(pod *k8sapi.Pod, parentKind
...
+	nonDisruptiveHelper := NewNonDisruptiveHelper(r.kubeClient, pod.Namespace, contKind, contName, pod.Name)
+
+	// Performs operations for non-disruptive resize actions
+	if err := nonDisruptiveHelper.OperateForNonDisruption(); err != nil {
+		glog.V(3).Infof("resizeContainer failed[%s]: failed to perform non-disruptive operations", id)
+		return err
+	}
+	defer nonDisruptiveHelper.CleanUp()
```
songbinliu: It would be much nicer that we can merge the two CleanUPs (update replicaNum and update SchedulerName),  if they are operating on the same controller object.  
But for now let's keep it simple and obvious.

pkg/action/executor/nondisruptive_helper.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	glog.V(2).Infof("Finished clean up for non-disruptive operations on %s %s/%s", h.contKind, h.namespace, h.contName)
+}
+
+func (h *NonDisruptiveHelper) updateReplicaWithRetry(newReplicas int) error {
+	// With retrying several times
+	glog.V(2).Infof("updateReplicaWithRetry: newReplicas = %d", newReplicas)
+	err := util.RetryDuring(defaultUpdateReplicaRetry, defaultUpdateReplicaTimeout, defaultUpdateReplicaSleep, func() error {
+		return autil.UpdateReplicas(h.client, h.namespace, h.contName, h.contKind, newReplicas)
```
songbinliu: In the cleanUp, before actually updateReplicas, it will be kind of safer if we can check that the current replicaNum setting  is still 2.  
As our action may take several minutes, user may modify the replicaNum during this period. So it will be safer to do a check first.

pkg/action/util/util.go
maxwangvmt: fixed

pkg/action/util/util.go
maxwangvmt: fixed

pkg/action/executor/resize_container.go
maxwangvmt: Discussed with Songbin. I will keep it as getting the parent's info doesn't invoke additional API call.

pkg/action/executor/nondisruptive_helper.go
maxwangvmt: fixed

pkg/action/executor/nondisruptive_helper.go
maxwangvmt: fixed

PR #123, opened 2017-11-17
non-disruptive flag and deployment support check

maxwangvmt:
This PR includes two changes:

- Add a flag to kubeturbo config to allow users to disable non-disruptive support
- Add k8s version check for non-disruptive support for Deployment controllers

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -70,7 +70,7 @@ type VMTServer struct {
 	NoneSchedulerName string
 
 	// Flag for supporting non-disruptive action
-	disableNonDisruptiveSupport bool
+	enableNonDisruptiveSupport bool
 }
 
 // NewVMTServer creates a new VMTServer with default parameters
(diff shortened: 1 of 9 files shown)


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -94,6 +97,7 @@ func (s *VMTServer) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.EnableKubeletHttps, "kubelet-https", kubelet.DefaultKubeletHttps, "Indicate if Kubelet is running on https server")
 	fs.StringVar(&s.K8sVersion, "k8sVersion", executor.HigherK8sVersion, "the kubernetes server version; for openshift, it is the underlying Kubernetes' version.")
 	fs.StringVar(&s.NoneSchedulerName, "noneSchedulerName", executor.DefaultNoneExistSchedulerName, "a none-exist scheduler name, to prevent controller to create Running pods during move Action.")
+	fs.BoolVar(&s.disableNonDisruptiveSupport, "disable-non-disruptive-support", false, "Indicate if nondisruptive action support is disabled")
```
songbinliu: Shall we change the names of the variables, like `disable-non-disruptive-support` to `enable-non-disruptive-support` , so that `if !r.disableNonDisruptiveSupport` can be `if r.enableNonDisruptiveSupport`, which is a little bit easier to get the meaning.

maxwangvmt: Yes, I modified it based on your suggestion in the updated PR.

PR #126, opened 2017-11-27
Remove target config from turbo config file and generate it based on KubeConfig

### LINKED ISSUE #101: kubeturbo config has unnecessary fields that requires user's input
In the current Kubeturbo config, we asked too many fields from users. 
{
	"communicationConfig": {
		"serverMeta": {
                    "version": "<TURBONOMIC_SERVER_VERSION>",
		    "turboServer": "<TURBONOMIC_SERVER_URL>"
		},
		"restAPIConfig": {
			"opsManagerUserName": "<TURBONOMIC_USERNAME>",
			"opsManagerPassword": "<TURBONOMIC_PASSWORD>"
		}
	},
	"targetConfig": {
		"probeCategory":"CloudNative",
		"targetType":"Kubernetes",
		"address":"http://<K8S_MASTER_IP>:8080",
		"username":"user1",
		"password":"pass1"
	}
}

The whole targetConfig should be handled in kubeturbo without user's input. Address, username and password are not necessary anymore since we require kubeconfig.

maxwangvmt:
Fixes #101. With the change, users need not input the target config, which will be populated automatically based on the target KubeConfig. Currently, only target address will be obtained from KubeConfig and others (probeCat, targetType, username, and password) will be set to default values.


pkg/discovery/configs/target_config.go
```diff
@@ -4,8 +4,10 @@ import "errors"
 
 const (
 	// When user doesn't specify username and password, the default username and password will be used.
-	defaultUsername = "defaultUser"
-	defaultPassword = "defaultPassword"
+	defaultUsername      = "defaultUser"
+	defaultPassword      = "defaultPassword"
+	defaultProbeCategory = "CloudNative"
```
songbinliu: same category/type will cause problem when adding more than one targets.

pkg/k8s_tap_service_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestParseK8sTAPServiceSpec(t *testing.T) {
+	targetName := "target-foo"
+	configPath := "../test/config/turbo-config"
+
+	kubeConfig := &rest.Config{Host: targetName}
+
+	got, err := ParseK8sTAPServiceSpec(configPath, kubeConfig)
+
+	if err != nil {
+		t.Errorf("Error while parsing the spec fiel %s: %v", configPath, err)
```
DongyiYang: typo ```t.Errorf("Error while parsing the spec field %s: %v", configPath, err)```
also, you should return here, right?

deploy/config
```diff
@@ -7,12 +7,5 @@
 			"opsManagerUserName": "foo",
 			"opsManagerPassword": "bar"
 		}
-	},
-	"targetConfig": {
```
songbinliu: Even user can use the default API server address as identifier, should we also provide user a way to specify the target Name(or Identifier)?

pkg/k8s_tap_service_test.go
maxwangvmt: Fixed the typo. t.Errorf will fail the test so no return needed. Thanks, Dongyi!!

pkg/discovery/configs/target_config.go
maxwangvmt: Discussed with Songbin, for more than one kubernetes clusters, the communication to OpsMgr will fail in one of them. One workaround here is to make catergory/type different for each cluster by appending some random strings.

enlinxu: Yes, Songbin and I discussed about it before. We don't have much choice here. The only thing we can improve is to make it less "random". Let's check together

songbinliu:
Will it be compatible to the older configuration files?

PR #130, opened 2017-12-06
Remove target config from turbo config file and generate it based on …

maxwangvmt:
Merged from master #126 (f1667daefa9a6283809a93b6ee80105a678a7fd1)

Remove target configuration from turbo config file


songbinliu:
Since there is no `gofmt` and `go vet` check in the ci process,  make sure `gofmt` and `go vet` are checked for this PR.

PR #133, opened 2017-12-15
Fix OM-28151 move action failed (for master 6.1)

songbinliu:
For bug [OM-28151](https://vmturbo.atlassian.net/browse/OM-28151), on master branch.
**Bug: Move action fails for bare metal**
The reason is that, if there is no stitching (in bare metal case), then the IPs of the node will be empty.

**Solution**: add node's IPs when discovering.


songbinliu:
Changes:
1. add IPs during discovery
    pkg/discovery/dtofactory/node_entity_dto_builder.go
    pkg/discovery/dtofactory/property/node_properties.go 

2. optimize the way to find the node in k8s
       pkg/action/executor/rescheduler.go 
       pkg/action/util/util.go

    (1) Try to find the k8s node by Entity.Properties();
    (2) Try to find the k8s node by Entity.DisplayName();
    (3) Try to find the k8s node by Entity.UUID;
    (4) Try to find the k8s node by IPs

3. (side work) Optimize the way to find the pod in k8s.
     pkg/action/executor/rescheduler.go  getTargetPod()
     (1) first try to find the pod by entity.Properties()

pkg/action/executor/rescheduler.go
```diff
@@ -85,45 +85,79 @@ func (r *ReScheduler) checkActionItem(action *proto.ActionItemDTO) error {
...
-		if hostSE.GetEntityType() == proto.EntityDTO_VIRTUAL_MACHINE {
-			vmData := hostSE.GetVirtualMachineData()
-			if vmData == nil {
-				err := fmt.Errorf("Missing virtualMachineData[%v] in targetSE.", hostSE.GetDisplayName())
-				glog.Error(err.Error())
-				return nil, err
-			}
-			machineIPs = vmData.GetIpAddress()
+	//0. check entity type
+	etype := hostSE.GetEntityType()
```
maxwangvmt: In some case, hostSe can be nil. Can it happen here?

pkg/action/executor/rescheduler.go
```diff
@@ -85,45 +85,79 @@ func (r *ReScheduler) checkActionItem(action *proto.ActionItemDTO) error {
...
+	//3. get node by UUID
+	node, err = util.GetNodebyUUID(r.kubeClient, hostSE.GetId())
+	if err == nil {
+		glog.V(2).Infof("Get node(%v) by UUID(%v).", node.Name, hostSE.GetId())
+		return node, nil
+	}
+
+	//4. get node by IP
+	vmIPs := getVMIps(hostSE)
+	if len(vmIPs) > 1 {
```
maxwangvmt: We don't accept len(vmIPs) == 1?

pkg/action/executor/rescheduler.go
```diff
@@ -85,45 +85,79 @@ func (r *ReScheduler) checkActionItem(action *proto.ActionItemDTO) error {
...
-			machineIPs = vmData.GetIpAddress()
+	//0. check entity type
+	etype := hostSE.GetEntityType()
+	if etype != proto.EntityDTO_VIRTUAL_MACHINE && etype != proto.EntityDTO_PHYSICAL_MACHINE {
+		err = fmt.Errorf("The target entity(%v) for move destiantion is neither a VM nor a PM.", etype)
+		glog.Error(err.Error())
+		return nil, err
+	}
+
+	//1. get node from properties
```
maxwangvmt: Here, we try four ways to get node:
1. from entity properties
2. from entity displayName
3. from entity uuid
4. from vm ip list

From my understand, the entity display name is set as the k8s node name. So, we can simply use the display name to get the k8s node. Why do we need the four ways to get it? Is it because some name truncating situation?

pkg/discovery/dtofactory/property/property_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	ps := BuildPodProperties(pod)
+	ns, name, err := GetPodInfoFromProperty(ps)
+	if err != nil {
+		t.Errorf("pod property test failed: %v", err)
+		return
+	}
+
+	if ns != pod.Namespace {
+		t.Error("Pod property test failed: namespace is wrong")
```
maxwangvmt: Better to log what's the desired and got

pkg/discovery/dtofactory/property/property_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if name != pod.Name {
+		t.Error("Pod property test failed: pod name is wrong")
```
maxwangvmt: Better to log what's the desired and got

pkg/discovery/dtofactory/node_entity_dto_builder_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	nodeIPs := getNodeIPs(node)
+	filter := make(map[string]struct{})
+	for _, ip := range nodeIPs {
+		filter[ip] = struct{}{}
+	}
+
+	addresses := node.Status.Addresses
+	if len(nodeIPs) != len(addresses) {
+		t.Errorf("Not equal: %d Vs. %d", len(nodeIPs), len(addresses))
```
maxwangvmt: Better to log what is not equal

pkg/action/executor/rescheduler.go
songbinliu: yea, because of name truncating, displayname is not very reliable.

pkg/action/executor/rescheduler.go
songbinliu: Yea, possible, if mediation server has bug. I will check this.

PR #134, opened 2017-12-15
Fix OM-28151 move action failed (for branch 6.0.0)

songbinliu:
For bug [OM-28151](https://vmturbo.atlassian.net/browse/OM-28151), on master branch.
**Bug: Move action fails for bare metal**
The reason is that, if there is no stitching (in bare metal case), then the IPs of the node will be empty.

**Solution**: add node's IPs when discovering.


songbinliu:
This is fix is for branch 6.0.0, while [PR133](https://github.com/turbonomic/kubeturbo/pull/133) for master branch.

PR #138, opened 2017-12-27
Deploy kubeturbo with Deployment and ConfigMap for turbo config

maxwangvmt:
The change is to deploy kubeturbo using Deployment controller, instead of bare pod. And instead of using turbo-config file, which requires to specify particular nodes to deploy, use ConfigMap for the configuration.

Use single yaml file for easy deployment. README files will be updated accordingly after the change  is finalized.


deploy/general-deploy/kubeturbo.yaml
```diff
@@ -1,30 +1,53 @@
-apiVersion: v1
-kind: Pod
+apiVersion: extensions/v1beta1
```
songbinliu: should we put the configMap before the deployment? Otherwise, the pod may not find the configMap when it tries to start.

deploy/general-deploy/turbo-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+            "serverMeta": {
+                "version": "6.1.0-SNAPSHOT",
+                "turboServer": "https://127.0.0.1:9400"
+            },
+            "restAPIConfig": {
+                "opsManagerUserName": "administrator",
+                "opsManagerPassword": "administrator"
+            }
+        }
+    }
```
DongyiYang: newline

deploy/general-deploy/kubeturbo.yaml
```diff
@@ -1,30 +1,33 @@
...
+            #- --kubelet-port=10250
+          volumeMounts:
+          - name: turbo-config
+            mountPath: /etc/kubeturbo
+            readOnly: true
+      volumes:
+      - name: turbo-config
+        configMap:
+          name: turbo-config
+      restartPolicy: Always
```
DongyiYang: newline

deploy/general-deploy/kubeturbo.yaml
```diff
@@ -1,30 +1,33 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: kubeturbo
+    spec:
+      containers:
+        - name: kubeturbo
+          # Replace the image with desired version
+          image: vmturbo/kubeturbo:6.1dev
```
DongyiYang: better put the tag as `latest`

deploy/general-deploy/kubeturbo.yaml
DongyiYang: This needs to be addressed in `README`.

deploy/general-deploy/kubeturbo.yaml
maxwangvmt: We don't have `latest` tag. Currently, 6.1dev is used and this example will be updated once we have new versions.

deploy/general-deploy/kubeturbo.yaml
maxwangvmt: Yes, `README` will be updated separately as it will also consider other changes to kubeturbo we recently made.

PR #141, opened 2018-01-11
Remove access commodities from supply chain 

pallavidn:
Edited the supply chain to remove access commodities from the node and pod templates.
This is to avoid supply chain errors  in the server for VMs that are validated using the template from kubeturbo instead of hypervisor. VMs without any pods running on them will not have not access commodities and will be removed from the market.
Also included an orthogonal change to the quota metric collection - to ensure that the capacity and used allocation metrics are collected for all the nodes in the cluster, iterate over the list of nodes in the cluster instead of the quota map keyset. Prior to this change, nodes without any running pods were not included for creating the allocation metrics. If the capacity/used values for a resource are not set DTO builder will not build  that commodity, so the allocation commodities for these nodes were not being created.

### DIFF
--- a/pkg/discovery/worker/metrics_collector.go
+++ b/pkg/discovery/worker/metrics_collector.go
@@ -203,10 +203,12 @@ func (collector *MetricsCollector) CollectNodeMetrics(podCollection PodMetricsBy
 	nodeMetricsCollection := make(map[string]*repository.NodeMetrics)
 
 	//Iterate over all the nodes in the collection
-	for nodeName, podsByQuotaMap := range podCollection {
-		node := collector.Cluster.NodeMap[nodeName]
-		if node == nil {
-			continue
+	for _, node := range collector.NodeList {
+		nodeName := node.Name
+		podsByQuotaMap, exists := podCollection[nodeName]
+		if !exists {
+			glog.V(4).Infof("cannot find pod metrics for node %s", nodeName)
+			podsByQuotaMap = make(map[string]PodMetricsList)
 		}
 
 		// collect the metrics for all the pods running on this node across all quotas
--- a/pkg/registration/supply_chain_factory.go
+++ b/pkg/registration/supply_chain_factory.go
@@ -5,22 +5,20 @@ import (
 
 	"github.com/turbonomic/kubeturbo/pkg/discovery/stitching"
 
+	"github.com/golang/glog"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
 	"github.com/turbonomic/turbo-go-sdk/pkg/supplychain"
 )
 
 var (
-	vCpuType           proto.CommodityDTO_CommodityType = proto.CommodityDTO_VCPU
-	vMemType           proto.CommodityDTO_CommodityType = proto.CommodityDTO_VMEM
-	cpuProvisionedType proto.CommodityDTO_CommodityType = proto.CommodityDTO_CPU_PROVISIONED
-	memProvisionedType proto.CommodityDTO_CommodityType = proto.CommodityDTO_MEM_PROVISIONED
-	cpuAllocationType  proto.CommodityDTO_CommodityType = proto.CommodityDTO_CPU_ALLOCATION
-	memAllocationType  proto.CommodityDTO_CommodityType = proto.CommodityDTO_MEM_ALLOCATION
-	transactionType    proto.CommodityDTO_CommodityType = proto.CommodityDTO_TRANSACTION
-
-	clusterType    proto.CommodityDTO_CommodityType = proto.CommodityDTO_CLUSTER
-	appCommType    proto.CommodityDTO_CommodityType = proto.CommodityDTO_APPLICATION
-	vmPMAccessType proto.CommodityDTO_CommodityType = proto.CommodityDTO_VMPM_ACCESS
+	vCpuType          proto.CommodityDTO_CommodityType = proto.CommodityDTO_VCPU
+	vMemType          proto.CommodityDTO_CommodityType = proto.CommodityDTO_VMEM
+	cpuAllocationType proto.CommodityDTO_CommodityType = proto.CommodityDTO_CPU_ALLOCATION
+	memAllocationType proto.CommodityDTO_CommodityType = proto.CommodityDTO_MEM_ALLOCATION
+	clusterType       proto.CommodityDTO_CommodityType = proto.CommodityDTO_CLUSTER
+	vmPMAccessType    proto.CommodityDTO_CommodityType = proto.CommodityDTO_VMPM_ACCESS
+	appCommType       proto.CommodityDTO_CommodityType = proto.CommodityDTO_APPLICATION
+	transactionType   proto.CommodityDTO_CommodityType = proto.CommodityDTO_TRANSACTION
 
 	fakeKey string = "fake"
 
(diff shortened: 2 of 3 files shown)


pkg/registration/supply_chain_factory.go
```diff
@@ -48,53 +46,72 @@ func NewSupplyChainFactory(pType stitching.StitchingPropertyType) *SupplyChainFa
...
 	return supplyChainBuilder.Create()
 }
 
+func (f *SupplyChainFactory) buildNodeSupplyBuilder() (*proto.TemplateDTO, error) {
+	nodeSupplyChainNodeBuilder := supplychain.NewSupplyChainNodeBuilder(proto.EntityDTO_VIRTUAL_MACHINE)
+
+	nodeSupplyChainNodeBuilder = nodeSupplyChainNodeBuilder.
+		Sells(vCpuTemplateComm). // sells to Pods
+		Sells(vMemTemplateComm). // sells to Pods
+		// also sells VMPMAccess, Cluster to Pods
```
songbinliu: the ClusterCommodity can be kept, it is always created during discovery.

pallavidn: The issue is that the VM without pods are not considered for stitching, so the access commodities  from kubeturbo are not transferred to these VMs, which will cause errors when the kubeturbo node template is used for validation check.

pkg/discovery/worker/metrics_collector.go
```diff
@@ -203,10 +203,12 @@ func (collector *MetricsCollector) CollectNodeMetrics(podCollection PodMetricsBy
 	nodeMetricsCollection := make(map[string]*repository.NodeMetrics)
 
 	//Iterate over all the nodes in the collection
```
songbinliu: There is an unnecessary comma in the argument list in several functions.

pkg/registration/supply_chain_factory.go
songbinliu: Nodes without Pods will not have VMPM_ACCESS commodity, but they still have CLUSTER commodity.  

Is CLUSTER commodity is also a kind of access commodity?

pallavidn: yes

songbinliu: Another question,  I thought the bug should be fixed from SDK or by changing the SE priority, but not by changing the supply chain definition in kubeturbo.

(1) Even node without pods are still can have VMPM_ACCESS commodities: 
[which are built from node labels](https://github.com/pallavidn/kubeturbo/blob/56c99d3204406143768ad41c5f155411c6f6e9be/pkg/discovery/dtofactory/node_entity_dto_builder.go#L147)).  The three nodes are all have labels. 

(2) The previous supplyChain definition for node does not have PMVM_ACCESS commodity.

pkg/registration/supply_chain_factory.go
```diff
@@ -110,17 +127,9 @@ func (f *SupplyChainFactory) buildQuotaSupplyBuilder() (*proto.TemplateDTO, erro
...
-		vmQuotaExtLinkBuilder.
-			ProbeEntityPropertyDef(supplychain.SUPPLY_CHAIN_CONSTANT_UUID, "UUID of the Node").
-			ExternalEntityPropertyDef(supplychain.VM_UUID)
-	case stitching.IP:
-		vmQuotaExtLinkBuilder.
-			ProbeEntityPropertyDef(supplychain.SUPPLY_CHAIN_CONSTANT_IP_ADDRESS, "IP of the Node").
-			ExternalEntityPropertyDef(supplychain.VM_IP)
-	default:
-		return nil, fmt.Errorf("Stitching property type %s is not supported.", f.stitchingPropertyType)
+	err := f.addVMStitchingProperty(vmQuotaExtLinkBuilder)
```
songbinliu: The Quota SE should also buy a `Cluster` commodity, which can confine it in the current kubernetes cluster.

pkg/registration/supply_chain_factory.go
pallavidn: - The access commodities should not really cause supply chain validation errors.
- This fix is here is to match the kubeturbo VM node template  as close as to the VM template from VC, to ensure that any missing kubeturbo commodities will not cause validation errors.
- The root cause is that the priority of the VM templates for both the probes is default to 0 and  one of the template is randomly picked.
This is a workaround, we will need to discuss with the VC team to increase the priority of the VC probe VM template  so it is always picked and kubeturbo commodities are not used for validation checks.

pkg/registration/supply_chain_factory.go
```diff
@@ -131,34 +140,15 @@ func (f *SupplyChainFactory) buildQuotaSupplyBuilder() (*proto.TemplateDTO, erro
...
+		Sells(vCpuTemplateComm). //sells to containers
 		Sells(vMemTemplateComm).
 		Sells(vmpmAccessTemplateComm).
 		Provider(proto.EntityDTO_VIRTUAL_MACHINE, proto.Provider_HOSTING).
-		// TODO we will re-include provisioned commodities bought by pod later.
-		//Buys(cpuProvisionedTemplateComm).
-		//Buys(memProvisionedTemplateComm).
-		Buys(clusterTemplateComm).
+		Buys(vCpuTemplateComm).
+		Buys(vMemTemplateComm).
```
songbinliu: Pod also buys `Cluster` and `VMPM_ACCESS` commodities. Is it better to only define these two commodities in the `ExternalLink`, or also defined these two here?

pkg/registration/supply_chain_factory.go
```diff
@@ -168,30 +158,35 @@ func (f *SupplyChainFactory) buildPodSupplyBuilder() (*proto.TemplateDTO, error)
...
 			ProbeEntityPropertyDef(supplychain.SUPPLY_CHAIN_CONSTANT_UUID, "UUID of the Node").
 			ExternalEntityPropertyDef(supplychain.VM_UUID)
 	case stitching.IP:
-		vmPodExtLinkBuilder.
+		extLinkBuilder.
 			ProbeEntityPropertyDef(supplychain.SUPPLY_CHAIN_CONSTANT_IP_ADDRESS, "IP of the Node").
 			ExternalEntityPropertyDef(supplychain.VM_IP)
 	default:
-		return nil, fmt.Errorf("Stitching property type %s is not supported.", f.stitchingPropertyType)
+		return fmt.Errorf("Stitching property type %s is not supported.", f.stitchingPropertyType)
```
songbinliu: Shall we log this error here?  This error msg is not able to be printed so far.

pkg/registration/supply_chain_factory.go
songbinliu: If the root cause is the priority or template type(Base Vs. Extension), then we can change the definition for the VirtualMachine node.
```go
/
type TemplateDTO struct {
	// The type of entity that the template represents. See EntityType
	// for the available types.
	TemplateClass *EntityDTO_EntityType `protobuf:"varint,1,req,name=templateClass,enum=common_dto.EntityDTO_EntityType" json:"templateClass,omitempty"`
	// The template type (Base or Extension), used during the validation process.
	TemplateType *TemplateDTO_TemplateType `protobuf:"varint,2,req,name=templateType,enum=common_dto.TemplateDTO_TemplateType" json:"templateType,omitempty"`
	// The priority of a Base template. For equivalent Base templates, Operations Manager uses the highest-priority
	// template, and discards discovered data from lower-priority Base templates.
	TemplatePriority *int32 `protobuf:"varint,3,req,name=templatePriority" json:"templatePriority,omitempty"`
	// This entity's list of {@link TemplateCommodity} items that it provides.
	CommoditySold []*TemplateCommodity `protobuf:"bytes,5,rep,name=commoditySold" json:"commoditySold,omitempty"`
	// The commodities bought from the different providers.
	// This Map contains the commodities bought where:
	CommodityBought []*TemplateDTO_CommBoughtProviderProp `protobuf:"bytes,6,rep,name=commodityBought" json:"commodityBought,omitempty"`
	// A map that defines the entity types that will be providers or consumers for this template entity.
	// The entry key is an entity type, from the EntityType enumeration. There can only be
	// one instance of each entity type in this map. The entry value is an instance of
	// ExternalEntityLink. Each entity link describes an entity type in the supply chain,
	// and the commodities it buys from or sells to the template entity.
	ExternalLink []*TemplateDTO_ExternalEntityLinkProp `protobuf:"bytes,7,rep,name=externalLink" json:"externalLink,omitempty"`
	// Each set represents a case where the entity must buy one commodity of the set ( a logical or of the set)
	// Note, the entity may buy more than one of the commodities in the set.
	CommBoughtOrSet  []*TemplateDTO_CommBoughtProviderOrSet `protobuf:"bytes,8,rep,name=commBoughtOrSet" json:"commBoughtOrSet,omitempty"`
	XXX_unrecognized []byte                                 `json:"-"`
}
```

songbinliu: If the node without any pod won't enter into the stitching loop, then the node won't get stitched. The problem is still not solved?

pallavidn: Will be opening a separate PR for this in the server.
We cannot use TEMPLATE.EXT if we want to run kubeturbo as standalone probe. Also, the default priority is 0, so it is the hypervisor probe that needs to increase the VM template priority to a much higher number. This also needs a PR and discussion on the server side.

songbinliu: Ok. Let's issue two bugs about the VC probe and the reconciling processing, so that the 6.0.1 image can work without this patch.

PR #142, opened 2018-01-16
OM-28809 Missing cluster summary argument to some tasks created by the dispatcher

pallavidn:
- Ensure that the ClusterSummary argument is set in all the tasks created by the dispatcher that creates the discovery workers
See pkg/discovery/worker/dispatcher.go for the primary change

- Added new unit tests in metrics collector

### DIFF
--- a/pkg/discovery/worker/k8s_discovery_worker.go
+++ b/pkg/discovery/worker/k8s_discovery_worker.go
@@ -134,6 +134,9 @@ func (worker *k8sDiscoveryWorker) executeTask(currTask *task.Task) *task.TaskRes
 		return task.NewTaskResult(worker.id, task.TaskFailed).WithErr(err)
 	}
 
+	for _, node := range currTask.NodeList() {
+		glog.V(4).Infof("%s : Node %s with %d pods\n", worker.id, node.Name, len(currTask.PodList()))
+	}
 	// wait group to make sure metrics scraping finishes.
 	var wg sync.WaitGroup
 	timeout := calcTimeOut(len(currTask.NodeList()))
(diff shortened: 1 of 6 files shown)


pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -134,6 +134,9 @@ func (worker *k8sDiscoveryWorker) executeTask(currTask *task.Task) *task.TaskRes
 		return task.NewTaskResult(worker.id, task.TaskFailed).WithErr(err)
 	}
 
+	for _, node := range currTask.NodeList() {
```
songbinliu: This 3 lines only for logging, can be changed to 
```
if glog.V(4) {
    for _, node := range currtass.NodeList() {
      glog.Infof(...)
   }
}

File: pkg/discovery/worker/metrics_collector.go [hunk 2]
```diff
@@ -96,7 +97,10 @@ func (podCollectionMap PodMetricsByNodeAndQuota) addPodMetric(podName, nodeName,
 
 // Create Pod metrics by selecting the pod compute resource usages from the metrics sink.
 // The PodMetrics are organized in a map by node and quota
-func (collector *MetricsCollector) CollectPodMetrics() PodMetricsByNodeAndQuota {
+func (collector *MetricsCollector) CollectPodMetrics() (PodMetricsByNodeAndQuota, error) {
+	if collector.Cluster == nil {
```
songbinliu: Print the log, as well as return the log to caller.

pkg/discovery/worker/metrics_collector.go
```diff
@@ -123,7 +129,7 @@ func (collector *MetricsCollector) CollectPodMetrics() PodMetricsByNodeAndQuota
 
 				var quotaComputeCap float64
 				quotaComputeCap, err := getQuotaComputeCapacity(quota, computeType)
-				if err == nil && quotaComputeCap < podCpuCap {
+				if err == nil && quotaComputeCap != repository.DEFAULT_METRIC_VALUE && quotaComputeCap < podCpuCap {
```
songbinliu: It will be easier for debugging if you can log the error, and the other conditions.

PR #145, opened 2018-01-22
Add dockerfile to redhat-img

enlinxu:
Using Rhel as base image to create kubeturbo docker image.

### DIFF
--- a/redhat-img/Dockerfile
+++ b/redhat-img/Dockerfile
@@ -1,0 +1,61 @@
+FROM registry.access.redhat.com/rhel7
+MAINTAINER Enlin Xu <enlin.xu@turbonomic.com>
+
+
+### Atomic/OpenShift Labels - https://github.com/projectatomic/ContainerApplicationGenericLabels
+LABEL name="Hybrid Cloud Container" \
+      vendor="Turbonomic" \
+      version="6.0" \
+      release="1" \
+      summary="Performance assurance for the applications in Openshift" \
+      description="Hybrid Cloud Container leverages Turbonomic control platform, to assure the performance of micro-services running in OpenShift, as well as the efficiency of underlying infrastructure." \
+### Required labels above - recommended below
+      url="http://www.turbonomic.com" \
+      run='docker run -tdi --name ${NAME} vmturbo/kubeturbo:latest' \
+      io.k8s.description="Hybrid Cloud Container will monitor and control the entire stack, from OpenShift down to your underlying infrastructure. " \
+      io.k8s.display-name="Hybrid Cloud Container" \
+      io.openshift.expose-services="" \
+      io.openshift.tags="turbonomic,Hybrid Cloud Container"
+
+### Atomic Help File - Write in Markdown, it will be converted to man format at build time.
+### https://github.com/projectatomic/container-best-practices/blob/master/creating/help.adoc
+COPY help.md /tmp/
+
+### add licenses to this directory
+COPY licenses /licenses
+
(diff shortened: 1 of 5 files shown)


redhat-img/Dockerfile
```diff
@@ -0,0 +1,61 @@
+FROM registry.access.redhat.com/rhel7
```
maxwangvmt: I think it's better to specify the version of the base OS image. Otherwise, we don't even know which version will be used when running docker build.

redhat-img/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+tag="red:v1"
```
maxwangvmt: This is not only the tag but also the image name. Should it be some name like kubeturbo:rhel7?

redhat-img/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+### add licenses to this directory
+COPY licenses /licenses
+
+COPY Dockerfile /Dockerfile
+
+#EXPOSE 10265
+
+### Add necessary Red Hat repos here
+RUN REPOLIST=rhel-7-server-rpms,rhel-7-server-optional-rpms \
```
maxwangvmt: I got error when building image here:

> Error getting repository data for rhel-7-server-rpms, repository not found

redhat-img/build.sh
songbinliu: yea, it should be `kubeturbo:<tag>`.

redhat-img/Dockerfile
songbinliu: you mean add a specific tag?  the latest version is `7.4-129`
https://access.redhat.com/containers/?tab=overview#/registry.access.redhat.com/rhel7


However, the [examples from redhat](https://github.com/RHsyseng/container-rhel-examples/blob/master/starter/Dockerfile) do not specify a tag.

songbinliu: So we change it to this, and update it regularly.
```yaml
FROM registry.access.redhat.com/rhel7:7.4-129
```

redhat-img/Dockerfile
songbinliu: change the block into:
```shell
### Add necessary Red Hat repos here
RUN REPOLIST=rhel-7-server-rpms,rhel-7-server-optional-rpms \
### Add your package needs here
    INSTALL_PKGS="golang-github-cpuguy83-go-md2man ca-certificates" && \
    yum -y update-minimal --disablerepo "*" --enablerepo rhel-7-server-rpms --setopt=tsflags=nodocs \
      --security --sec-severity=Important --sec-severity=Critical && \
    yum -y install --disablerepo "*" --enablerepo ${REPOLIST} --setopt=tsflags=nodocs ${INSTALL_PKGS} && \
### help file markdown to man conversion
    go-md2man -in /tmp/help.md -out /help.1 && \
    yum clean all
```

redhat-img/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+### Atomic Help File - Write in Markdown, it will be converted to man format at build time.
+### https://github.com/projectatomic/container-best-practices/blob/master/creating/help.adoc
+COPY help.md /tmp/
+
+### add licenses to this directory
+COPY licenses /licenses
+
+COPY Dockerfile /Dockerfile
+
+#EXPOSE 10265
```
songbinliu: remove this `EXPOSE` line

PR #147, opened 2018-01-22
Add logging-agent sidecar container

maxwangvmt:
Add logging-agent container. The logging agent is a sidecar container for kubeturbo to perform rotation and compression of log files.

The logging agent by default performs daily rotation and compression (to gz files) to the kubeturbo log files located in /var/log. The files will be kept for 30 days before cleaned up.

Users then can perform kubectl command to retrieve the log files: 

_kubectl cp <POD_NAME>:/var/log -c kubeturbo <DESTINATION_DIR>_

### DIFF
--- a/logging-agent/README.md
+++ b/logging-agent/README.md
@@ -5,7 +5,7 @@ The logging agent by default performs daily rotation and compression to the `kub
 
 ### Deploy `kubeturbo` with the sidecar container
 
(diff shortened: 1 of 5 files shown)


maxwangvmt:
In the updated commit, the base OS image is changed to the same one (rhel7) in kubeturbo to be consistent. To reduce the image size, both kubeturbo and logging agent can use the rhel-atomic base OS image in further improvement.

PR #149, opened 2018-01-24
Add prefix to kubernetes target id.

maxwangvmt:
Add "Kubernetes-" prefix to k8s target id to avoid having same id as other targets of different types (e.g., aws).


maxwangvmt:
The updated commit is to remove the random number of the probe category (the random number is not necessary and caused some problem in turbonomic UI side) and to add space between the words of cloud and native. As a result, the probe category becomes 'Cloud Native'.

PR #150, opened 2018-01-26
Connect to cluster and nodes during kubeturo target validation

pallavidn:
- Moved location of KubeletClient.
- Added methods to ClusterProcessor to access the Kuberentes API Server to get all nodes and then to access each node using kubeletClient. This is used to validate the kubeturbo target.
- Created interfaces for ClusterInfoScrapper and KubeletClient, to add unit tests for ClusterProcesor.
- Added unit tests for ClusterProcessor.


pkg/discovery/discovery_performance_tuning.go
```diff
@@ -83,7 +83,9 @@ func (dc *K8sDiscoveryClient) discoverWithNewFrameworkWithoutCompliance() ([]*pr
 	clusterProcessor := &processor.ClusterProcessor{
 		ClusterInfoScraper: dc.k8sClusterScraper,
 	}
-	kubeCluster, err := clusterProcessor.ProcessCluster()
+	// Check connection to the cluster
+	kubeCluster, err := dc.clusterProcessor.ConnectCluster(false)
```
maxwangvmt: What if err is not _nil_? We ignore it?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -6,36 +6,116 @@ import (
...
 )
 
 // Class to query the cluster data from the Kubernetes API server and create the KubeCluster entity
 // to represent the cluster, the nodes and the namespaces.
 type ClusterProcessor struct {
 	ClusterInfoScraper *cluster.ClusterScraper
+	NodeScrapper       *kubeclient.KubeletClient
 }
 
 // Query the Kubernetes API Server to get the cluster nodes and namespaces.
```
maxwangvmt: Please update the doc accordingly.

pkg/discovery/processor/cluster_processor.go
```diff
@@ -6,36 +6,116 @@ import (
...
+	err = processor.ConnectToNodes(checkAllNodes)
+
+	return kubeCluster, err
+}
 
-	kubeCluster := &repository.KubeCluster{
-		Name:       svcID,
-		Nodes:      make(map[string]*repository.KubeNode),
-		Namespaces: make(map[string]*repository.KubeNamespace),
+func (processor *ClusterProcessor) ConnectToNodes(checkAllNodes bool) error {
```
maxwangvmt: Adding documentation to describe this function will help to understand.

pkg/discovery/processor/cluster_processor.go
```diff
@@ -6,36 +6,116 @@ import (
...
+		return fmt.Errorf("Error getting nodes for cluster : %s\n", err)
 	}
+	glog.V(2).Infof("There are %d nodes\n", len(nodeList))
 
-	// Nodes
-	nodes, err := processor.processNodes(kubeCluster.Name)
+	var connectionErrors []error
+	var connectionMsgs []string
+
+	n := rand.Int() % len(nodeList)
```
maxwangvmt: Let's add some comments for the logic.

From my understanding, if _checkAllNodes_ is false, we only check one random node. If the random node is not responding, the validation will fail. Correct? If so, it may result in some inconsistent validation result. E.g., let's say there are two nodes with one node healthy and the other not responsive. If we click validation button 10 times, about 5 times will fail and others succeed. Is this behavior we want?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -8,111 +8,134 @@ import (
...
 		glog.V(2).Infof("Successfully connected to nodes : %s\n", strings.Join(connectionMsgs, ", "))
-		return nil
 	}
-	var errorStr []string
-	for i, err := range connectionErrors {
-		errorStr = append(errorStr, fmt.Sprintf("Error %d: %s", i, err.Error()))
+
+	errStr := strings.Join(connectionErrors, ", ")
+	if len(connectionErrors) > 0 {
+		glog.V(2).Infof("Errors connecting to nodes : %s\n", errStr)
```
maxwangvmt: Failure in connecting to nodes is an error. Should we use glog.Errorf here?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -8,111 +8,134 @@ import (
...
-		}
-		ip := repository.ParseNodeIP(node)
-		kc := processor.NodeScrapper
-		cpuFreq, err := kc.GetMachineCpuFrequency(ip)
+	for _, node := range nodeList {
+		nodeCpuFrequency, err := checkNode(node, processor.nodeScrapper)
 
 		if err != nil {
-			connectionErrors = append(connectionErrors, fmt.Errorf("%s:%s", node.Name, err))
+			connectionErrors = append(connectionErrors, fmt.Sprintf("%s:%s", node.Name, err))
```
maxwangvmt: If we fail connecting to the node, it will be helpful to know the node status. So, we can also add node.Status in the message.

PR #154, opened 2018-02-08
Not monitor Pods with special annotations

songbinliu:
FIX [OM-29864](https://vmturbo.atlassian.net/browse/OM-29864)

**Problem**:
1. Currently, actions on _Kubeturbo_ itself will failed;
2. There may be some other pods/services which may not want to be monitored by _Kubeturbo_;

**Solution**
_Kubeturbo_  will check the annotations of pods/services, to  decide whether the object should be monitored.  If these objects have the annotation indicating not to be monitored, then _Kubeturbo_ will not monitored these objects.


deploy/general-deploy/README.md
```diff
@@ -76,6 +76,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        kubeturbo.io/moitored: "false"
```
maxwangvmt: Typo: 'monitored'

deploy/general-deploy/kubeturbo.yaml
```diff
@@ -2,26 +2,32 @@ apiVersion: extensions/v1beta1
...
   name: kubeturbo
+  namespace: turbo
   labels:
     app: kubeturbo
 spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        kubeturbo.io/moitored-flag: "false"
```
maxwangvmt: Type: 'monitored' and it should be without -flag

pkg/discovery/util/pod_util.go
```diff
@@ -4,26 +4,57 @@ import (
...
+	TurboMonitorAnnotation string = "kubeturbo.io/monitored"
 )
 
+// check whether a Kubernetes object is monitored or not by its annotation.
+// the object can be: Pod, Service, Namespace, or others
+func IsMonitoredFromAnnotation(obj interface{}) bool {
+	acc, err := meta.Accessor(obj)
+	if err != nil {
+		glog.Errorf("Not a Kubernetes Object: %v", err)
+		return true
```
maxwangvmt: In such error case, shouldn't we throw error, instead of returning true?

PR #159, opened 2018-02-27
Fix OM-31513: new mechanism to move pod

songbinliu:
FIX [OM-31513](https://vmturbo.atlassian.net/browse/OM-31513)

**Problem**
With previous mechanism, kubeturbo can only move/resize pod controlled by ReplicaSet/ReplicationController, but cannot move/resize pod controlled by Deployment in newer version Kubernetes (1.7.8+). 

**Solution**
[This mechanism](https://github.com/songbinliu/podmove) works well for pod controlled by ReplicaSet/ReplicationController/Deployment.


pkg/action/executor/resize_container.go
```diff
@@ -256,151 +254,121 @@ func (r *ContainerResizer) Execute(actionItem *proto.ActionItemDTO) error {
...
 	fullName := util.BuildIdentifier(pod.Namespace, pod.Name)
 	parentKind, parentName, err := util.GetPodParentInfo(pod)
 	if err != nil {
 		glog.Errorf("failed to get pod[%s] parent info: %v", fullName, err)
-		return err
+		return nil, fmt.Errorf("Failed")
+	}
+
+	if !util.SupportedParent(parentKind) {
+		glog.Errorf("Move action aborted: parent kind(%v) is not supported.", parentKind)
```
maxwangvmt: Move -> Resize

pkg/action/executor/move_util.go
```diff
@@ -36,63 +40,141 @@ func copyPodInfo(oldPod, newPod *api.Pod) {
...
+	flag = true
+	return xpod, nil
+}
+
+func createClonePod(client *kclient.Clientset, pod *api.Pod, nodeName string) (*api.Pod, error) {
+	npod := &api.Pod{}
+	copyPodWithoutLabel(pod, npod)
+	npod.Spec.NodeName = nodeName
+	npod.Name = genNewPodName(pod.Name)
+	util.AddAnnotation(npod, TurboActionAnnotationKey, TurboMoveAnnotationValue)
```
maxwangvmt: Question: if we have a pod with both turbo move and resize actions applied, we will only see one annotation for the latest action, is it desired?

pkg/action/executor/constants.go
```diff
@@ -3,31 +3,17 @@ package executor
...
-	defaultCheckUpdateReplicaRetry   = 36
-	defaultUpdateReplicaSleep        = time.Second * 20
-	defaultUpdateReplicaTimeout      = time.Second * 180
-	defaultUpdateReplicaRetry        = 10
-	defaultMoreGrace                 = time.Second * 20
+	defaultPodCheckSleep      = time.Second * 10
+	defaultPodCreateSleep     = time.Second * 11
+	defaultUpdateReplicaSleep = time.Second * 20
+
+	TurboActionAnnotationKey   string = "kubeturbo.io/action"
```
maxwangvmt: Better to add some comments (in somewhere) to explain how these annotation get created and used.

pkg/action/executor/move_util.go
```diff
@@ -36,63 +40,141 @@ func copyPodInfo(oldPod, newPod *api.Pod) {
...
-	interval := defaultPodCreateSleep
-	timeout := time.Duration(retryNum+1) * interval
-	err = util.RetryDuring(retryNum, timeout, interval, func() error {
-		_, inerr := podClient.Create(npod)
-		return inerr
-	})
+	//2. delete the original pod--podA
+	delOpt := &metav1.DeleteOptions{}
+	if err := podClient.Delete(pod.Name, delOpt); err != nil {
+		glog.Warningf("Move pod warning: failed to delete original pod: %v", err)
```
maxwangvmt: Shouldn't it be an error message since failure deleting the pod causes some problem?

pkg/action/executor/resize_container_util.go
```diff
@@ -202,56 +175,116 @@ func genMemoryQuantity(newValue float64) (resource.Quantity, error) {
...
+	npod, changed, err := createResizePod(client, pod, spec)
+	if err != nil {
+		glog.Errorf("resizeContainer failed[%s]: failed to create a resized pod: %v", id, err)
+		return nil, err
 	}
 
-	// 1. copy the original pod
+	if !changed {
+		glog.Warningf("resizeContainer aborted[%s]: no need do resize container.", id)
+		return nil, nil
```
maxwangvmt: If it's not changed, do we want to return an error for the resize action?

pkg/action/util/util.go
```diff
@@ -545,3 +546,36 @@ func FindPodsByController(kubeClient *client.Clientset, namespace, contName, con
 	}
 	return pods, nil
 }
+
+// check whether parentKind is supported for MovePod/ResizeContainer actions
+// currently, these actions can only works on barePod, ReplicaSet, and ReplicationController
+//   Note: pod's parent cannot be Deployment.
```
maxwangvmt: Could you please add comments how it works if a pod is controlled by Deployment?

pkg/action/executor/move_util.go
```diff
@@ -36,63 +40,140 @@ func copyPodInfo(oldPod, newPod *api.Pod) {
...
+	if phase == api.PodRunning {
+		if len(nodeName) > 0 {
+			if !strings.EqualFold(pod.Spec.NodeName, nodeName) {
+				return false, fmt.Errorf("running on an unexpected node[%v].", pod.Spec.NodeName)
+			}
 		}
+		return false, nil
+	}
+
+	if pod.DeletionGracePeriodSeconds != nil {
```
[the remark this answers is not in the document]
songbinliu: My bug, should be **pod.DeletionTimestamp**. Will change it to the way [Kubernetes using](https://github.com/kubernetes/kubernetes/blob/0c7e7ae1d9cccd0cca7313ee5a8ae3c313b72139/pkg/controller/controller_utils.go#L830):
```go
func IsPodActive(p *v1.Pod) bool {
	return v1.PodSucceeded != p.Status.Phase &&
		v1.PodFailed != p.Status.Phase &&
		p.DeletionTimestamp == nil
}
```

pkg/action/executor/move_util.go
songbinliu: yes,  action annotation item is overwritten, so only the latest action info is saved.

pkg/action/executor/move_util.go
```diff
@@ -36,63 +40,140 @@ func copyPodInfo(oldPod, newPod *api.Pod) {
...
+func copyPodWithoutLabel(oldPod, newPod *api.Pod) {
+	copyPodInfo(oldPod, newPod)
+
+	// set Labels and OwnerReference to be empty
+	newPod.Labels = make(map[string]string)
+	newPod.OwnerReferences = []metav1.OwnerReference{}
+}
+
+// TODO: prevent podName to become too long
+func genNewPodName(name string) string {
```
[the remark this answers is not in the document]
songbinliu: Yes, the long name will be a issue. Let's fix this in another PR.

PR #160, opened 2018-02-27
Create KubeCluster object during each discovery with all the nodes

pallavidn:
We represent the node's schedulable status as an Access commodity in the server.
When a node's schedulable status is disabled, the VM continues to sell the commodity. 
- Changed the Cluster processor code to create the KubeCluster during each discovery. KubeCluster is an internal repository for all the entities in the cluster. This will update the KubeNodes with the api.Node entity (which contains the "schedulable" status) during each discovery.

### DIFF
--- a/pkg/discovery/k8s_discovery_client.go
+++ b/pkg/discovery/k8s_discovery_client.go
@@ -41,30 +41,30 @@ func NewDiscoveryConfig(probeConfig *configs.ProbeConfig,
 type K8sDiscoveryClient struct {
 	config            *DiscoveryClientConfig
 	k8sClusterScraper *cluster.ClusterScraper
-	clusterProcessor  *processor.ClusterProcessor
-	kubeCluster       *repository.KubeCluster
 
-	dispatcher      *worker.Dispatcher
-	resultCollector *worker.ResultCollector
+	clusterProcessor *processor.ClusterProcessor
+	dispatcher       *worker.Dispatcher
+	resultCollector  *worker.ResultCollector
 
 	wg sync.WaitGroup
 }
 
 func NewK8sDiscoveryClient(config *DiscoveryClientConfig) *K8sDiscoveryClient {
-	// make maxWorkerCount of result collector twice the worker count.
-	resultCollector := worker.NewResultCollector(workerCount * 2)
 	k8sClusterScraper := cluster.NewClusterScraper(config.probeConfig.ClusterClient)
 
+	// for discovery tasks
 	clusterProcessor := processor.NewClusterProcessor(k8sClusterScraper, config.probeConfig.NodeClient)
+	// make maxWorkerCount of result collector twice the worker count.
+	resultCollector := worker.NewResultCollector(workerCount * 2)
 
 	dispatcherConfig := worker.NewDispatcherConfig(k8sClusterScraper, config.probeConfig, workerCount)
 	dispatcher := worker.NewDispatcher(dispatcherConfig)
 	dispatcher.Init(resultCollector)
 
 	dc := &K8sDiscoveryClient{
-		clusterProcessor:  clusterProcessor,
-		k8sClusterScraper: k8sClusterScraper,
 		config:            config,
+		k8sClusterScraper: k8sClusterScraper,
+		clusterProcessor:  clusterProcessor,
 		dispatcher:        dispatcher,
 		resultCollector:   resultCollector,
 	}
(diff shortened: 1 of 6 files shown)


pkg/discovery/k8s_discovery_client.go
```diff
@@ -41,30 +41,30 @@ func NewDiscoveryConfig(probeConfig *configs.ProbeConfig,
 type K8sDiscoveryClient struct {
 	config            *DiscoveryClientConfig
 	k8sClusterScraper *cluster.ClusterScraper
-	clusterProcessor  *processor.ClusterProcessor
-	kubeCluster       *repository.KubeCluster
 
-	dispatcher      *worker.Dispatcher
-	resultCollector *worker.ResultCollector
+	clusterProcessor *processor.ClusterProcessor
```
songbinliu: Can be concurrency problem because the variables:`clusterProcessor, dispatcher` and `resultCollector` of  `K8sDiscoveryClient`.

The instance of `K8sDiscoveryClient` can be shared for multiple `Discovery` actions. These `Discovery` actions can overlap during their executions.  So these variables are shared, and modified in concurrently, which could cause data corruption.

So it is better to keep `K8sDiscoveryClient` readonly, and put these variables in another agent for each `Discovery` action.

pkg/discovery/k8s_discovery_client.go
```diff
@@ -41,30 +41,30 @@ func NewDiscoveryConfig(probeConfig *configs.ProbeConfig,
...
+	dispatcher       *worker.Dispatcher
+	resultCollector  *worker.ResultCollector
 
 	wg sync.WaitGroup
```
songbinliu: `wg` is defined, but never used.

pkg/discovery/k8s_discovery_client.go
songbinliu: We can solve this problem in another PR. The `resultCollector`  has been here since 6.0.

pkg/discovery/processor/cluster_processor.go
```diff
@@ -74,26 +86,25 @@ func (processor *ClusterProcessor) connectToNodes() error {
...
-
-	errStr := strings.Join(connectionErrors, ", ")
-	if len(connectionErrors) > 0 {
+	// All nodes are unreachable
+	if len(unreachable) == len(nodeList) {
+		errStr := strings.Join(connectionErrors, ", ")
 		glog.Errorf("Errors connecting to nodes : %s\n", errStr)
+		validationResult.IsValidated = false
+		return validationResult, fmt.Errorf(errStr)
 	}
```
songbinliu: should also log the _connectionErrors_ in cases of `len(unreachable) != len(nodeList)`

pkg/discovery/processor/cluster_processor.go
```diff
@@ -109,30 +120,32 @@ func checkNode(node *v1.Node, kc kubeclient.KubeHttpClientInterface) (float64, e
 }
 
 // Query the Kubernetes API Server to get the cluster nodes and namespaces and set in tge cluster object
-func (processor *ClusterProcessor) DiscoverCluster(kubeCluster *repository.KubeCluster) error {
+func (processor *ClusterProcessor) DiscoverCluster() (*repository.KubeCluster, error) {
 	// If connection to cluster is successful, then the cluster object is not null
-	if kubeCluster == nil {
-		return fmt.Errorf("Failed to connect to cluster")
+	if !processor.validationResult.IsValidated {
```
songbinliu: The logical here is: **_If the cluster is not get validated, then won't perform discovery?_**

I think this could be a problem: If the initial validation has failed, then before the next validation request, even if the cluster is reachable, discovery requests will still fail?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -36,35 +37,46 @@ func NewClusterProcessor(kubeClient *cluster.ClusterScraper, kubeletClient *kube
...
 
+type ClusterValidationResult struct {
+	IsValidated      bool
+	UnreachableNodes []string
+}
+
 // Connects to the Kubernetes API Server and the nodes in the cluster.
-// If connection is successful a KubeCluster entity to represent the cluster is created
-func (processor *ClusterProcessor) ConnectCluster() (*repository.KubeCluster, error) {
+// ClusterProcessor is updated with the validation result.
```
songbinliu: `ClusterProcessor` is also a global variable for `K8sDiscoveryClient`,  If it has state, make sure the update of the state is protected.

pkg/discovery/processor/cluster_processor.go
songbinliu: In addition,  If validation is failed, the `processor.validationResult` could be a `nil` pointer.

pkg/discovery/k8s_discovery_client.go
pallavidn: K8sDiscoveryClient will handle discovery requests for one instance of target. The worker pool and cluster processor (which holds the cluster API config/client) are persisted across requests. Multiple discovery requests for the same target are queued by the server.

PR #161, opened 2018-02-28
Fix-om-31364 set VirtualMachine priority and template_type

songbinliu:
[OM-31364](https://vmturbo.atlassian.net/browse/OM-31364)

Add two functions to set supplyChainNodeTemplate 's priority and template_type.

**Problem**
For probe kubeturbo, in supplyChain definition, virtualMahcine template's priority is set to 0, and template_type is set to template_base.
In this case, when stitching with VCProbe, which has the same priority and template_type setting for virtualMachine, DiscoveryContextUtil will choose randomly one of the two template, and report an error.

**Fix**
(1) In default settings, virtualMahcine tempalte _priority_ is **-1**,  and _template type_ is **tempate_base**;
(2) Expose two parameters to adjust these two values;

Together with [PR](https://github.com/turbonomic/turbo-go-sdk/pull/48)


songbinliu:
The other probes: [VMM](https://vmturbo.atlassian.net/browse/OM-31587),  [DB and AppD](https://vmturbo.atlassian.net/browse/OM-30208) have also set the proxy VM priority to -1, and keep it as template_base.

PR #162, opened 2018-03-01
Check if pod privilege is supported before executing resize/move actions.

maxwangvmt:
In Openshift, some pods may have higher privilege (e.g., anyuid) associated to security context constraints (scc) than the default one (restricted). For such pods, Kubeturbo, when executes resize/move actions, will fail cloning the pods with their required privilege since Kubeturbo doesn't have sufficient privilege. As a result, the cloned pods can get crashed due to some permission denied issues as observed in the case of Nginx container, which requires higher privilege than default.

Here, the change is to add the privilege check in Kubeturbo before executing actions. We currently allows the 'restricted' privilege for pods to be moved/resized. For pods with other privilege types, the action, when executed, will fail to avoid crashing the pod.

Assigning Kubeturbo with higher privilege is further direction to address this issue.

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -165,6 +165,11 @@ func movePod(client *kclient.Clientset, tpod *api.Pod, nodeName string, retryNum
 }
 
 func createClonePod(client *kclient.Clientset, pod *api.Pod, nodeName string) (*api.Pod, error) {
+	if !util.SupportPrivilegePod(pod) {
+		err := fmt.Errorf("The pod %s has privilege requirement unsupported", pod.Name)
+		return nil, err
+	}
+
 	npod := &api.Pod{}
 	copyPodWithoutLabel(pod, npod)
 	npod.Spec.NodeName = nodeName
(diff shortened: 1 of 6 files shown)


pkg/action/executor/move_util.go
```diff
@@ -165,6 +165,11 @@ func movePod(client *kclient.Clientset, tpod *api.Pod, nodeName string, retryNum
 }
 
 func createClonePod(client *kclient.Clientset, pod *api.Pod, nodeName string) (*api.Pod, error) {
+	if !util.SupportPrivilegePod(pod) {
```
songbinliu: why not put this check more forward before the action, for example the [ReScheduler.preActionCheck()](https://github.com/turbonomic/kubeturbo/blob/b312eb5787cc4edb599dbc42fb1ee778d29e7c66/pkg/action/executor/rescheduler.go#L232)

pkg/action/executor/resize_container_util.go
```diff
@@ -257,6 +257,11 @@ func resizeContainer(client *kclient.Clientset, tpod *k8sapi.Pod, spec *containe
 // create a pod with new resource limit/requests
 //    return false if there is no need to update resource amount
 func createResizePod(client *kclient.Clientset, pod *k8sapi.Pod, spec *containerResizeSpec) (*k8sapi.Pod, bool, error) {
+	if !util.SupportPrivilegePod(pod) {
```
songbinliu: can move this action check more forward

PR #165, opened 2018-03-09
OM-32235: discard dangling services

songbinliu:
[OM-32235](https://vmturbo.atlassian.net/browse/OM-32235) 

**Problem**: There are some dangling services, who will fail the supply chain check, and won't be able to be assigned to any hosting cluster (as shown in [OM-32235](https://vmturbo.atlassian.net/browse/OM-32235)).

**Dangling services**: services for whom kubeturbo fails to find the supporting pods.

**Solutions**: Discard these kinds of services during discovery.


songbinliu:
There are two kinds of Dangling services:

1. underlying pods are not ready: not created, pending, or error;
2. do not depend on Pods, but depend on other kinds of services, or external services;

For the first case, kubeturbo will pick up the service when the underlying pods are ready.

PR #166, opened 2018-03-12
Filter out pods that are not in ready status when executing discovery

maxwangvmt:
Currently, Kubeturbo sends to om server the discovered pods, including those pods not in ready status. It causes some supply-check-failed issues at the server side for those unready pods due to missing commodities (e.g., VMem).

The change here is to exclude those _unready_ pods from the topology sent to om server since currently we do (and should) not apply any actions to those pods with invalid statues (e.g., Error).

### DIFF
--- a/pkg/cluster/cluster_info_scraper.go
+++ b/pkg/cluster/cluster_info_scraper.go
@@ -10,6 +10,7 @@ import (
 	api "k8s.io/client-go/pkg/api/v1"
 
 	"github.com/golang/glog"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
 )
 
 const (
@@ -177,7 +178,7 @@ func (s *ClusterScraper) GetKubernetesServiceID() (svcID string, err error) {
 	return
 }
 
-func (s *ClusterScraper) GetRunningPodsOnNodes(nodeList []*api.Node) []*api.Pod {
+func (s *ClusterScraper) GetRunningAndReadyPodsOnNodes(nodeList []*api.Node) []*api.Pod {
 	pods := []*api.Pod{}
 	for _, node := range nodeList {
 		nodeRunningPodsList, err := s.findRunningPodsOnNode(node.Name)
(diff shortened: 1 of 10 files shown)


pkg/discovery/util/pod_util.go
```diff
@@ -137,3 +137,29 @@ func GroupPodsByNode(pods []*api.Pod) map[string][]*api.Pod {
...
+		} else {
+			glog.V(4).Infof("Pod %s is not ready", pod.Name)
+		}
+	}
+
+	return readyPods
+}
+
+// PodIsReady checks if a pod is in Ready status.
+func podIsReady(pod *api.Pod) bool {
```
songbinliu: Should make sure  `condition.Type == api.PodReady` for all the conditions.
For example, for pod: `default/router-1-25zn2`, which is  in `Unknown` state:
screen shot 2018-03-12 at 4 24 39 pm

I think we may check the `Pod.Status.ContainerStatuses.Ready` for all the containers in the Pod.

songbinliu: screen shot 2018-03-12 at 4 22 30 pm

another pod in `Error` state, and its container is not running.

songbinliu: Talked with @maxwangvmt . I misunderstood it. This check works 👍

pkg/cluster/cluster_info_scraper.go
```diff
@@ -177,7 +178,7 @@ func (s *ClusterScraper) GetKubernetesServiceID() (svcID string, err error) {
 	return
 }
 
-func (s *ClusterScraper) GetRunningPodsOnNodes(nodeList []*api.Node) []*api.Pod {
+func (s *ClusterScraper) GetRunningAndReadyPodsOnNodes(nodeList []*api.Node) []*api.Pod {
```
songbinliu: Another potential problem: we get the list of running pods first, then get the container metrics.
If during the interval, a pod goes wrong, then there may be no metrics for the containers.

So we may need to remove the container/applications during building the entityDTOs:
[container bought commodities](https://github.com/turbonomic/kubeturbo/blob/e4b9156c5838631308e96fc75073bc9d8ca8375a/pkg/discovery/dtofactory/container_dto_builder.go#L161)

[application bought commodities](https://github.com/turbonomic/kubeturbo/blob/e4b9156c5838631308e96fc75073bc9d8ca8375a/pkg/discovery/dtofactory/application_entity_dto_builder.go#L195)

PR #167, opened 2018-03-15
Add key to allocation commodities bought by pod from resource quotas

pallavidn:
- Add key to the allocation commodities sold by resource quotas.
- Add key to the allocation commodities bought by the pods from resource quotas.

### DIFF
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -187,7 +187,7 @@ func (builder *podEntityDTOBuilder) getPodCommoditiesSold(pod *api.Pod, cpuFrequ
 }
 
 // Build the CommodityDTOs bought by the pod from the node provider.
-// Commodities bought are vCPU, vMem, cpuProvisioned, memProvisioned, access, cluster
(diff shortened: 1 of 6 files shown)


pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -252,21 +252,27 @@ func (builder *podEntityDTOBuilder) getPodCommoditiesBought(pod *api.Pod, cpuFre
...
-	if err != nil {
-		return nil, err
+	for _, resourceType := range podResourceCommodityBoughtFromQuota {
+		commBought, err := builder.getResourceCommodityBoughtWithKey(metrics.PodType, key,
+			resourceType, quotaUID, converter, nil)
+		if err != nil {
+			// skip this commodity
+			glog.Errorf("%s::%s: cannot build sold commodity %s : %s",
+				metrics.PodType, key, resourceType, err)
+			continue
```
songbinliu: To prevent from unexpected happening, should abort PodEntity building on error, 
as the latest [PR](https://github.com/turbonomic/kubeturbo/pull/166) does.

Unexpected thing:
If neither VCPU nor VMEM commodity bought from VDC get built, then market will assume that this pod does not need to buy any commodity from VDC.

PR #169, opened 2018-03-22
Implement cache for pod change after actions applied

maxwangvmt:
Problem:
If there are more than one action that operates on the same pod sent to Kubeturbo, only the first one will succeed. It's because the pod will be changed (deleted and recreated with a different name) after the action (pod move/container resize) succeeds.

Fix:
Implement a cache to store the pod change info after actions applied. So, the following actions can get the new pod name from the cache. To do this, the actions with same pod/controller involved need to be blocked earlier. Currently, we block those actions inside each executor, which might be too late as some outdated pod-related data has been populated at that time.

So, this PR also includes the refactoring of action handler and executors to apply the lock earlier at action_handler to prevent getting wrong data when executing the action.

Two commits are included:
1. The code change at kubeturbo side
2. Dependency update using Godep tool to add the [go-cache](https://github.com/patrickmn/go-cache) library

Testing has been done, including concurrently sending two actions on the same pod (vcpu/vmem resize or pod move), two actions on the pods on same controller, and four actions on two pods with same controller.

More tests will be done and unit tests will be added.


pkg/action/util/pod_manager.go
```diff
@@ -0,0 +1,124 @@
...
+	return nil, fmt.Errorf("Failed to find Pod with name %s and uid %s", name, uid)
+}
+
+// Need to find the latest pod as there might be multiple action applied to the pod.
+// E.g., If there are three actions in a row for the same pod, the third one need to get the pod name after the change made by the second action.
+//       action1: pod-foo (old) => pod-foo-c (new)
+//       action2: pod-foo (old) => pod-foo-c (from cache) => pod-foo-c-c (new)
+//       action3: pod-foo (old) => pod-foo-c (from cache) => pod-foo-c-c (from cache again) => pod-foo-c-c-c (new)
+func (p *PodCachedManager) getLatestValue(key string) (interface{}, bool) {
+	val, ok := p.podCache.Get(key)
```
enlinxu: This segment can be merged into the loop. Plus, what is this loop for? What do we expect to change during this 20 times continuous retry, what use case does this loop cover?

pkg/action/util/pod_manager.go
```diff
@@ -0,0 +1,124 @@
...
+	// might have been applied to the pod.
+	if cachedName, ok := p.getLatestValue(name); ok {
+		glog.V(3).Infof("Using cached name %s for pod %s", cachedName, name)
+		pod, err := podutil.GetPodInPhase(podClient, cachedName.(string), api.PodRunning)
+		if err != nil || pod != nil {
+			return pod, err
+		}
+	}
+
+	pod, err := podutil.GetPodInPhase(podClient, name, api.PodRunning)
```
enlinxu: We should wrap GetPodInPhase and GetPodInPhaseByUid in one method, GetPod with both uid and name parsing in, and in GetPod we can try one then the other. This way we prevent the caller to write these procedures again and again.

pkg/action/action_handler.go
```diff
@@ -221,3 +275,30 @@ func keepAlive(tracker sdkprobe.ActionProgressTracker, stop chan struct{}) {
 		}
 	}()
 }
+
+func checkActionExecutionDTO(actionExecutionDTO *proto.ActionExecutionDTO) error {
```
enlinxu: Add method comments

pkg/action/action_handler.go
```diff
@@ -106,60 +136,84 @@ func (h *ActionHandler) ExecuteAction(actionExecutionDTO *proto.ActionExecutionD
...
 		glog.Errorf(msg.Error())
 		return err
 	}
 
+	// Process the action execution output, including caching the pod name change.
+	h.processOutput(output)
+
 	return nil
 }
+func (h *ActionHandler) getRelatedPod(actionItem *proto.ActionItemDTO) *api.Pod {
```
enlinxu: Add method comments

pkg/action/util/pod_manager.go
maxwangvmt: Songbin and I discussed this before. This is intended to try using the pod name for cached first before trying using the uid. The reason is that retrieving pod by uid is more expensive for api server as it caches pod names, instead of uid.

pkg/action/util/pod_manager.go
maxwangvmt: To clarify, the loop is not "retry". It is to find the latest pod after multiple actions being applied.
One example is the case that three actions sent to kubeturbo concurrently and all these actions are on the same pod (pod-foo), e.g., Container vcpu resize, vmem resize, and Pod move.

After vcpu resize finishes, the pod will have name "pod-foo-c". The second action, say vmem resize, will find the cached name "pod-foo-c" by key "pod-foo" as designed. After it's done, it will change the pod to have name "pod-foo-c-c".

The loop is for the third action. The move action will use the key "pod-foo" to get the cached name "pod-foo-c". Without the loop, the action will fail since "pod-foo-c" has been changed to "pod-foo-c-c". The loop is to get the latest pod, which is "pod-foo-c-c" in this case. And then, the move action can be applied to the right pod.

pkg/action/util/pod_manager.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Need to find the latest pod as there might be multiple action applied to the pod.
+// E.g., If there are three actions in a row for the same pod, the third one need to get the pod name after the change made by the second action.
+//       action1: pod-foo (old) => pod-foo-c (new)
+//       action2: pod-foo (old) => pod-foo-c (from cache) => pod-foo-c-c (new)
+//       action3: pod-foo (old) => pod-foo-c (from cache) => pod-foo-c-c (from cache again) => pod-foo-c-c-c (new)
+func (p *PodCachedManager) getLatestValue(key string) (interface{}, bool) {
+	val, ok := p.podCache.Get(key)
+	if !ok {
```
songbinliu: It could be better if add some log in case of `!ok`.

pkg/action/util/pod_manager.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return nil, fmt.Errorf("Failed to find Pod with name %s and uid %s", name, uid)
+}
+
+// Need to find the latest pod as there might be multiple action applied to the pod.
+// E.g., If there are three actions in a row for the same pod, the third one need to get the pod name after the change made by the second action.
+//       action1: pod-foo (old) => pod-foo-c (new)
+//       action2: pod-foo (old) => pod-foo-c (from cache) => pod-foo-c-c (new)
+//       action3: pod-foo (old) => pod-foo-c (from cache) => pod-foo-c-c (from cache again) => pod-foo-c-c-c (new)
+func (p *PodCachedManager) getLatestValue(key string) (interface{}, bool) {
```
songbinliu: The type of the return value can be `string`, instead of `interface{}`. Because we already know it is   string during the iteration.

pkg/action/util/pod_manager.go
```diff
@@ -0,0 +1,127 @@
...
+	val, ok := p.podCache.Get(key)
+	if !ok {
+		return val, ok
+	}
+
+	// Value found. Continue to getting further value using the value found as the new key
+	// Set the maximum for the iteration as in any case (e.g., with a bug), we don't want to run the loop infinitely.
+	maxIte := 20
```
songbinliu: (1) define a constant for the value of `maxIte`;
(2) log a warning or error in case of reaching to the `maxIte`;

pkg/action/util/pod_manager.go
```diff
@@ -0,0 +1,127 @@
...
+	// Try the cached name first. If it exists, it means some action was applied so it got renamed.
+	// If try the original name first, it may find the pod that is in terminating process from the previous action
+	// as the action doesn't check if the termination process finishes.
+
+	// Pod is not found with the name. Try using the cached name because actions
+	// might have been applied to the pod.
+	if cachedName, ok := p.getLatestValue(name); ok {
+		glog.V(3).Infof("Using cached name %s for pod %s", cachedName, name)
+		pod, err := podutil.GetPodInPhase(podClient, cachedName.(string), api.PodRunning)
+		if err != nil || pod != nil {
```
songbinliu: Is it more conventional to check `err != nil` only?

pkg/discovery/util/pod_util.go
```diff
@@ -163,3 +165,69 @@ func podIsReady(pod *api.Pod) bool {
...
+// If pod not found, nil pod pointer will be returned without error.
+func GetPodInPhase(podClient v1.PodInterface, name string, phase api.PodPhase) (*api.Pod, error) {
+	pod, err := podClient.Get(name, meta_v1.GetOptions{})
+	if err != nil {
+		return pod, err
+	}
+
+	if pod == nil {
+		glog.V(3).Infof("Pod %s is not found", name)
+		return nil, nil
```
songbinliu: `return nil, fmt.Errorf("Pod %s is not found", name)` will make it easier for invoker.

pkg/discovery/util/pod_util.go
```diff
@@ -163,3 +165,69 @@ func podIsReady(pod *api.Pod) bool {
 	glog.Errorf("Unable to get status for pod %s", pod.Name)
 	return false
 }
+
+// GetPodInPhase finds the pod with the specific name in the specific phase (e.g., Running).
+// If pod not found, nil pod pointer will be returned without error.
+func GetPodInPhase(podClient v1.PodInterface, name string, phase api.PodPhase) (*api.Pod, error) {
+	pod, err := podClient.Get(name, meta_v1.GetOptions{})
+	if err != nil {
+		return pod, err
```
songbinliu: why not return `nil, err`?

songbinliu: In addition, I think it will easier for debugging if we can log the API call error.

pkg/action/util/pod_manager.go
songbinliu: It is safer to check ` err == nil && pod != nil`

pkg/action/action_lock_store.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+	return helper, nil
+}
+
+// Gets the lock key for the action item.
+func (a *ActionLockStore) getLockKey(actionItem *proto.ActionItemDTO) (string, error) {
+	pod := a.podFunc(actionItem)
+
+	if pod == nil {
```
songbinliu: add some comments about when Pod can be nil

pkg/discovery/util/pod_util.go
maxwangvmt: Here, it is intended to return nil pod without error. In the caller, we want to distinguish if it's an error from api call or it's simply no such pod exist (so we can try cached name). If it is an error with api call, we don't need to try cached value while if it's pod not found, we should try other possible cases (e.g., cached name, uid).

pkg/action/action_lock_store.go
```diff
@@ -0,0 +1,101 @@
...
+// the lock is used by other action item.
+// The key used to acquire the lock is as follows:
+//
+// 1. If the action item is associated to a container pod (meaning, the function podFunc returns a pod),
+//    the key is the pod id for bare-pod cases and its parent controller id for non-bare-pod cases.
+//
+// 2. Otherwise, the key is the id of the target SE of the action item.
+func (a *ActionLockStore) getLock(actionItem *proto.ActionItemDTO) (*util.LockHelper, error) {
+	id := actionItem.GetUuid()
+	if key, err := a.getLockKey(actionItem); err != nil {
```
songbinliu: I think it can not handle bare-Pod case well:
    Two actions on the same bare-Pod,  then the key for them will be same. The first action will update the cache, while the second action will use the same key to lock the bare-Pod.

Problem is that:  During the execution of second action,  a third action comes on the same bare-Pod, then the third action will use another key to get a lock, and it will get the lock. Then the second and third actions will both working on the same Pod.

songbinliu: For bare-Pod case, we should try to get the lock with the latest pod name

pkg/action/action_handler.go
```diff
@@ -106,60 +130,87 @@ func (h *ActionHandler) ExecuteAction(actionExecutionDTO *proto.ActionExecutionD
...
 
-	worker, exist := h.actionExecutors[actionType]
-	if !exist {
-		msg := fmt.Errorf("Action %s on %s is not supported.", actionType, actionItem.GetTargetSE().GetEntityType())
-		glog.Errorf(msg.Error())
-		return msg
+	// After getting the lock, need to get the k8s pod again as the previous action could delete the pod and create a new one.
+	// In such case, the action should be applied on the new pod.
+	// Currently, all actions need to get its related pod. If not needed, the pod is nil.
+	pod := h.getRelatedPod(actionItem)
```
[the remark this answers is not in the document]
maxwangvmt: The cases of our current action support all return a non-nil pod. However, in general, the pod could be nil. In such case, the TurboActionExecutorInput contains a nil pod so the associated executor simply do not use it.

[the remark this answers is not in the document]
songbinliu: Then some check should be added in the individual executors.

pkg/discovery/util/pod_util.go
songbinliu: Should we try to find it in the cache in either case if the pod is not available via k8s API?

maxwangvmt: That's similar to what we do now (if I understand it correctly). The caller will find the pod with the cached name first and if not found, try the original pod name. So, we need return nil error when this method doesn't find any pod.

songbinliu: Anyway, I think it is not safe to use `pod != nil` as a flag. Because in case of API call fail, the result of `pod` is not defined.  Right now, it is not `nil`, but it may change.  Would rather add another flag to distinguish the different cases if necessary.

PR #170, opened 2018-03-26
append containerName to Application

songbinliu:
[OM-33277](https://vmturbo.atlassian.net/browse/OM-33277)

**Problem**:  Application displayName does not contain the container's name

**Solution**: Append container's name  to application displayName.


pkg/discovery/util/key_func.go
```diff
@@ -67,16 +68,24 @@ func ParseContainerId(containerId string) (string, int, error) {
 
 // Application's displayName = "App-namespace/podName"
```
maxwangvmt: Update the comment.

PR #171, opened 2018-03-28
stitching-Azure

songbinliu:
[OM-33266](https://vmturbo.atlassian.net/browse/OM-33266)

*Problem*
 Kubeturbo stitching may fail with multiple VMs with the same stitching IP.

*Solution*
Use VM's UUID to do stitching.


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -68,7 +68,7 @@ type VMTServer struct {
 	// Instead we use the systemUUID of each node, which is equal to UUID of corresponding
 	// VM discovered by VM probe.
 	// The default value is false.
```
maxwangvmt: Update the comment. The default now is true.

pkg/discovery/stitching/node_uuid.go
```diff
@@ -0,0 +1,169 @@
...
+	return "Default"
+}
+
+func (d *defaultNodeUUIDGetter) GetUUID(node *api.Node) (string, error) {
+	uuid := node.Status.NodeInfo.SystemUUID
+	if len(uuid) < 1 {
+		glog.Errorf("Node uuid is empty: %++v", node)
+		return "", fmt.Errorf("Empty uuuid")
+	}
+	uuid = strings.ToLower(uuid)
```
maxwangvmt: Could you please add a comment about why we need lower case here?

pkg/discovery/stitching/node_uuid.go
```diff
@@ -0,0 +1,169 @@
...
+		glog.Warningf("Not a Azure node: % ++v", node)
+	}
+
+	suuid := node.Status.NodeInfo.SystemUUID
+	if len(suuid) < 1 {
+		glog.Errorf("Node system uuid is empty: %++v", node)
+		return "", fmt.Errorf("Empty uuid")
+	}
+
+	//1. get partially reversed uuid
```
maxwangvmt: Could you please add comment about why the reverse is needed?

PR #172, opened 2018-04-02
Fix DaemonSet pods as monitored issue

maxwangvmt:
In some pods of DaemonSet, they are set to be monitored if they don't have the "createdBy" annotation. This is undesired since we don't want to monitor those pods.

We already have right logic to check the controller info using "OwnerReference" object associated to pods in addition to the annotation check. The fix here is to cleanup the duplicate logic and keep the desired one to check pods' controller info. Hence, it can correct detect DaemonSet pods and set them as not monitored.

Most code change here is to move around the related functions for getting parent controller info.


pkg/discovery/util/pod_util_test.go
```diff
@@ -187,6 +187,27 @@ func TestGetReadyPods(t *testing.T) {
...
+	isController := true
+	podWithOwnerRef.OwnerReferences = []metav1.OwnerReference{{Kind: "DaemonSet", Name: "daemon-set-foo", Controller: &isController}}
+
+	// Without Owner reference
+	if Monitored(pod) {
+		t.Errorf("The pod in DaemonSet shouldn't be monitored")
+	}
+
+	// With Owner reference
+	if Monitored(podWithOwnerRef) {
```
songbinliu: `podWithOwnerRef` has the same `Annotations` as `pod` does.  If the plan is to test the case that Pod with `OwnerReference`, then the annotations should be cleared first.

PR #173, opened 2018-04-04
OM-33736: Handling CPU allocation resource usage values by the Quota DTO Builder

pallavidn:
Quota DTO Builder handles the conversion of CPU capacity/usage from number of cores to MHZ.
When the quota is defined for the CPU resource, the usage value obtained from the quota object is in number of cores and should be converted to MHZ.
When the quota is not defined for the CPU resource, the usage value is computed using the cpu usage  of all the pods running in that namespace and is represented in MHz and the conversion is not required. The additional conversion would increase the usage values to a larger number resulting in invalid utilization values.

### DIFF
--- a/pkg/discovery/dtofactory/quota_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/quota_entity_dto_builder.go
@@ -137,20 +137,21 @@ func (builder *quotaEntityDTOBuilder) getQuotaCommoditiesSold(quota *repository.
 		// For CPU resources, convert the capacity and usage values expressed in
 		// number of cores to MHz
 		if metrics.IsCPUType(resourceType) && quota.AverageNodeCpuFrequency > 0.0 {
-			// modify the capacity value
+			// always modify the capacity value
 			newVal := capacityValue * quota.AverageNodeCpuFrequency
-			glog.V(4).Infof("%s: changed %s capacity from %f -> %f\n",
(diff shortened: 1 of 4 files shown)


pkg/discovery/dtofactory/quota_entity_dto_builder.go
```diff
@@ -137,20 +137,21 @@ func (builder *quotaEntityDTOBuilder) getQuotaCommoditiesSold(quota *repository.
 		// For CPU resources, convert the capacity and usage values expressed in
 		// number of cores to MHz
 		if metrics.IsCPUType(resourceType) && quota.AverageNodeCpuFrequency > 0.0 {
```
songbinliu: what will happen if `quota.AverageNodeCpuFrequency <= 0.0`?

pkg/discovery/repository/kube_cluster.go
```diff
@@ -14,9 +14,10 @@ import (
 // New discovery will bring in changes to the nodes and namespaces
 // Aggregate structure for nodes, namespaces and quotas
 type KubeCluster struct {
-	Name       string
-	Nodes      map[string]*KubeNode
-	Namespaces map[string]*KubeNamespace
+	Name             string
+	Nodes            map[string]*KubeNode
+	Namespaces       map[string]*KubeNamespace
+	ComputeResources map[metrics.ResourceType]float64
```
songbinliu: where is `ComputeResources` is used?

PR #184, opened 2018-06-22
OM-34325: Fix timeouts on validation/discovery

win32pro:
1. Add cache to be used to smooth over drops in communications between kubeturbo and individual nodes
2. Make validation multithreaded and succeed on the first successful node validation. Fail if validation takes too long or no nodes are accessible.


Makefile
```diff
@@ -5,6 +5,7 @@ PACKAGES := go list ./... | grep -v /vendor | grep -v /out
 
 
 product: clean
+	env GOOS=darwin GOARCH=amd64 go build -gcflags='-N -l' -o ${OUTPUT_DIR}/kubeturbo.osx ./cmd/kubeturbo
```
songbinliu: `darwin` is not needed for product.

pkg/discovery/k8s_discovery_client.go
```diff
@@ -122,13 +122,14 @@ func (dc *K8sDiscoveryClient) Validate(accountValues []*proto.AccountValue) (*pr
 		errorDtos = append(errorDtos, errorDto)
 		validationResponse.ErrorDTO = errorDtos
 	} else {
-		glog.V(2).Infof("Validation response - connected to cluster and nodes\n")
+		glog.V(2).Infof("Validation response - connected to cluster\n")
```
songbinliu: we can remove the trailing `\n`

pkg/discovery/processor/cluster_processor.go
```diff
@@ -65,68 +71,107 @@ func (processor *ClusterProcessor) ConnectCluster() error {
 	return nil
 }
 
+// Perform a single node validation
+func (processor *ClusterProcessor) checkNodesWorker(work chan *v1.Node, done chan bool, index int) {
+	glog.Infof("node verifier worker %d starting", index)
```
songbinliu: should add Info level: `glog.V(x).Infof`

pkg/discovery/processor/cluster_processor.go
```diff
@@ -65,68 +71,107 @@ func (processor *ClusterProcessor) ConnectCluster() error {
...
+			glog.Infof("node verifier worker %d finished. Successful verification", index)
+			return
+		}
+	}
+}
+
+// Wait for at least one of the workers to complete successfully
+// or timeout
+func waitForCompletion(done chan bool) bool {
+	cycles := int(totalWaitTime.Nanoseconds() / sleepInterval.Nanoseconds())
```
songbinliu: time.Timer can replace the for loop.

pkg/discovery/processor/cluster_processor.go
```diff
@@ -65,68 +71,107 @@ func (processor *ClusterProcessor) ConnectCluster() error {
 	return nil
 }
 
+// Perform a single node validation
+func (processor *ClusterProcessor) checkNodesWorker(work chan *v1.Node, done chan bool, index int) {
```
songbinliu: IMO, we can add another input argument to the `checkNodeWorker`:
```golang
checkNodesWorker(work chan *v1.Node, stop chan struct{}, done chan bool, index int)
```
The worker will listen on both the `work` and `stop` channel. If it receives a signal from `stop` channel, then the worker should stop.   
In this way, we don't have to drain the `work` channel.

pkg/kubeclient/kubelet_client.go
```diff
@@ -81,14 +92,62 @@ func (client *KubeletClient) postRequestAndGetValue(req *http.Request, value int
...
+	// Get the data
 	summary := &stats.Summary{}
 	err := client.ExecuteRequestAndGetValue(host, summaryPath, summary)
+	// Lock and check cache
+	client.cacheLock.Lock()
+	defer client.cacheLock.Unlock()
+	entry, entryPresent := client.cache[host]
+	if err != nil {
+		if entryPresent {
+			glog.Infof("unable to retrieve machine[%s] summary: %v. Using cached value", host, err)
```
songbinliu: `warnig` or `error`?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -9,7 +9,13 @@ import (
...
 	"github.com/turbonomic/kubeturbo/pkg/kubeclient"
 	"k8s.io/client-go/pkg/api/v1"
-	"strings"
+	"time"
+)
+
+const (
+	workers = 10
+	sleepInterval = 10 * time.Second
+	totalWaitTime = 60 * time.Second
```
songbinliu: `totalWaitTime` should be configurable?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -65,68 +71,107 @@ func (processor *ClusterProcessor) ConnectCluster() error {
...
-		if err != nil {
-			connectionErrors = append(connectionErrors, fmt.Sprintf("%s [err:%s]", node.Name, err))
-			unreachable = append(unreachable, node.Name)
-		} else {
-			connectionMsgs = append(connectionMsgs,
-				fmt.Sprintf("%s [cpu:%v MHz]", node.Name, nodeCpuFrequency))
-		}
+	// The connection data
+	size := len(nodeList)
+	work := make(chan *v1.Node, size)
```
songbinliu: I think we can optimize this `single-producer, multi-consumers` pattern by better naming fo data input/output channels, control channels.

pkg/kubeclient/kubelet_client.go
```diff
@@ -81,14 +92,62 @@ func (client *KubeletClient) postRequestAndGetValue(req *http.Request, value int
...
+	// Get the data
 	var minfo cadvisorapi.MachineInfo
 	err := client.ExecuteRequestAndGetValue(host, specPath, &minfo)
+	// Lock and check cache
+	client.cacheLock.Lock()
+	defer client.cacheLock.Unlock()
+	entry, entryPresent := client.cache[host]
+	if err != nil {
+		if entryPresent {
+			glog.Infof("unable to retrieve machine[%s] machine info: %v. Using cached value", host, err)
```
songbinliu: warning, or error, or leveled Infof?

pkg/discovery/processor/cluster_processor.go
meirlaker: While not required, it might pay to collect and log a few simple summary stats, e.g., #nodes in cluster, #nodes failed while waiting for one to succeed.

Makefile
win32pro: For product no, this is for debugging. The final docker image won't contain it, and there's little danger in generating both.

pkg/discovery/processor/cluster_processor.go
win32pro: That complicates the flow. You have several scenarios, and you have to create additional channels to handle them. Any common control channel will have to contain enough messages to shut down every worker. The flow gets interesting at that point. Not sure if the approach is easier.

pkg/discovery/processor/cluster_processor.go
win32pro: Good question. Not sure.

enlinxu: Let's make it configurable, we can parse it as an argument when running kubeturbo. 60 can be the default value.

pkg/discovery/processor/cluster_processor.go
enlinxu: I agree with Michael here that we don't complicate our flow here.

Makefile
maxwangvmt: I think this is used locally and prefer not to commit.

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -36,6 +36,8 @@ const (
 	defaultVMPriority           = -1
 	defaultVMIsBase             = true
 	defaultDiscoveryIntervalSec = 600
+	defaultValidationWorkers    = 10
+	defaultValidationTimeout    = 60
```
maxwangvmt: How is the timeout related to the timeout at server side? Seems server has a timeout 30 seconds.

pkg/kubeclient/kubelet_client.go
```diff
@@ -33,12 +34,22 @@ type KubeHttpClientInterface interface {
 	GetMachineCpuFrequency(host string) (uint64, error)
 }
 
+// Cache structure.
+// TODO(MB): Make sure the nodes that are no longer discovered are being removed from the cache!
```
maxwangvmt: I think this is important implementation of the entire validation flow. For now, the cache never expires. However, the node status changes are common in kubernetes so it needs to be taken care here. We can expire the cached data with the cache library already used in kubeturbo.

win32pro: We've agreed that this isn't needed for this release. Also, the cache invalidation should not be a timing based, but driven by what's being discovered.

cmd/kubeturbo/app/kubeturbo_builder.go
win32pro: The timeout can be configured. These are just defaults. The server will timeout if something is horribly wrong, and server timeouts may change in the future.

PR #185, opened 2018-06-24
Mark applications and containers with providerMustClone.  Mark all en…

bcervenak:
…tities as monitored.  Mark pods that are mirrors or part of a daemon set as not controllable


pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -107,6 +107,11 @@ func (builder *applicationEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]
 				ebuilder.Monitored(false)
 			}
 
+			truep := true
+			ebuilder.ConsumerPolicy(&proto.EntityDTO_ConsumerPolicy{
```
enlinxu: Why are we making app mandatory provider true?

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -125,6 +125,14 @@ func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]*proto.E
 			glog.V(3).Infof("Pod %v is not monitored.", displayName)
 		}
 
+		controllable := util.Controllable(pod)
+		if !controllable {
+			glog.V(3).Infof("Pod %v is not controllable.", displayName)
+		}
+		entityDTOBuilder.ConsumerPolicy(&proto.EntityDTO_ConsumerPolicy{
```
enlinxu: We should make containerpod mandatory provider true also here.

pkg/discovery/dtofactory/application_entity_dto_builder.go
bcervenak: There is no longer a mandatory provider.  If providerMustClone is true, then the underlying provider will clone if the current SE needs to clone.

enlinxu: Sorry for the confusion, plz ignore my comment here and the one below!

pkg/discovery/dtofactory/pod_entity_dto_builder.go
bcervenak: See above.

pkg/util/topology_dumper.go
```diff
@@ -0,0 +1,94 @@
+package util
```
maxwangvmt: This file is used only for testing and we don't want it to be included in the product build. We can have it in some test util files, e.g., utils_test.go.

PR #188, opened 2018-06-27
OM-36531: Mark inactive nodes as such.

win32pro:


pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -61,18 +61,19 @@ func (builder *nodeEntityDTOBuilder) BuildEntityDTOs(nodes []*api.Node) ([]*prot
 		// display name.
 		displayName := node.Name
 		entityDTOBuilder.DisplayName(displayName)
+		nodeActive := util.NodeIsReady(node) && util.NodeIsSchedulable(node)
```
enlinxu: I am not sure we also want to mark the not schedulable node unknown. We may want to understand a bit better about how m2 handle unknown state. For example, do we copy over the unknown vm into plan? I knew that for not_monitored entity, we don't copy them into plan, I suspect we might do the same for unknown entities. In the case that all masters are not schedulable, we may not want to see that.

Also, do we want to add this logic here or when we return the value from cache (i.e., in GetSummary and GetMachineInfo in PR#184). For example, the node may be in ready state, but some how the kubelet port is not accessible from Turbo all of a sudden, we should also mark that node unknown?

esara: I agree with @enlinxu , not schedulable is a different semantics, it is not unknown. Lets stick with notready for this

PR #191, opened 2018-06-28
Do not create kubenode for notready node and notschedulable node, to …

enlinxu:
This is to fix OM-29813 and OM-36523. We don't want to count the capacity for VDC when the node is either not ready or not schedulable.
Updated the unit test as well.

### DIFF
--- a/pkg/discovery/processor/cluster_processor.go
+++ b/pkg/discovery/processor/cluster_processor.go
@@ -2,6 +2,8 @@ package processor
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/golang/glog"
 	"github.com/turbonomic/kubeturbo/pkg/cluster"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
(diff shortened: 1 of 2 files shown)


pkg/discovery/processor/cluster_processor.go
```diff
@@ -187,10 +188,14 @@ func (processor *ClusterProcessor) DiscoverCluster() (*repository.KubeCluster, e
 		return nil, fmt.Errorf("Error getting nodes for cluster %s:%s\n", clusterName, err)
 	}
 	glog.V(2).Infof("There are %d nodes\n", len(nodeList))
-
 	for _, item := range nodeList {
-		nodeEntity := repository.NewKubeNode(item, clusterName)
-		kubeCluster.SetNodeEntity(nodeEntity)
+		nodeActive := util.NodeIsReady(item) && util.NodeIsSchedulable(item)
+		if !nodeActive {
+			glog.V(2).Info("Node status is NotReady or NotSchedulable, skip in Quota creation")
```
maxwangvmt: Let's add the node name here.

PR #200, opened 2018-08-20
Do not include the notReady node in the quota build and do not includ…

enlinxu:
…e scheduledisabled node into quota capacity calculation

### DIFF
--- a/pkg/discovery/worker/quotas_discovery_worker.go
+++ b/pkg/discovery/worker/quotas_discovery_worker.go
@@ -6,6 +6,7 @@ import (
 	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/stitching"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/discovery/worker/quotas_discovery_worker.go
```diff
@@ -56,12 +57,17 @@ func (worker *k8sResourceQuotasDiscoveryWorker) Do(quotaMetricsList []*repositor
 	kubeNodes := worker.Cluster.Nodes
 	var nodeUIDs []string
 	var totalNodeFrequency float64
+	var activeNodeCount float64 = 0
```
maxwangvmt: Should be integer

pkg/discovery/worker/quotas_discovery_worker.go
```diff
@@ -77,8 +83,11 @@ func (worker *k8sResourceQuotasDiscoveryWorker) Do(quotaMetricsList []*repositor
 		// Bought resources from each node
 		// create provider entity for each node
 		for _, node := range kubeNodes {
-			nodeUID := node.UID
-			quotaEntity.AddNodeProvider(nodeUID, quotaMetrics.AllocationBoughtMap[nodeUID])
+			//Do not include the node that is not ready
+			if util.NodeIsReady(node.Node) {
```
maxwangvmt: Can you please explain why here we don't check if the node is schedulable (comparing to the checks above)?

enlinxu: Very good question Max! We still want the scheduledisabled nodes appearing in the relationship of vdc. My take is , scheduleDisabled is a normal status, it's by design. While the NotReady nodes are more like temporary down that we want to remove it from vdc, making it status unknown, etc.

PR #206, opened 2018-09-24
Fix NPE when GetMachineInfo is called before GetSummary.

win32pro:
Return err when using the cache in case machine info or summary hasn't been set yet.

### DIFF
--- a/pkg/kubeclient/kubelet_client.go
+++ b/pkg/kubeclient/kubelet_client.go
@@ -131,11 +131,12 @@ func (client *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
 	entry, entryPresent := client.cache[host]
 	if err != nil {
 		if entryPresent {
-			glog.V(2).Infof("unable to retrieve machine[%s] summary: %v. Using cached value", host, err)
 			entry.used = true
 			if entry.statsSummary == nil {
+				glog.V(2).Infof("unable to retrieve machine[%s] summary: %v. The cached value unavailable", host, err)
 				return nil, err
 			}
+			glog.V(2).Infof("unable to retrieve machine[%s] summary: %v. Using cached value", host, err)
 			return entry.statsSummary, nil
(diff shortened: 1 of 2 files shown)


pkg/kubeclient/kubelet_client.go
```diff
@@ -166,6 +169,9 @@ func (client *KubeletClient) GetMachineInfo(host string) (*cadvisorapi.MachineIn
 		if entryPresent {
```
enlinxu: Shall we add the extra condition check here? The thing is we are using one cache object for two different information, I am not sure if summary is nil, we should count as cache used.

pkg/kubeclient/kubelet_client.go
```diff
@@ -133,6 +133,9 @@ func (client *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
...
 			glog.V(2).Infof("unable to retrieve machine[%s] summary: %v. Using cached value", host, err)
 			entry.used = true
+			if entry.statsSummary == nil {
+				return nil, err
```
enlinxu: let's add another log message here to indicate that cache is used ,but summary is nil. Otherwise we can't distinguish the case whether we populate the summary from the cache or not.

pkg/kubeclient/kubelet_client.go
```diff
@@ -166,6 +169,9 @@ func (client *KubeletClient) GetMachineInfo(host string) (*cadvisorapi.MachineIn
 		if entryPresent {
 			glog.V(2).Infof("unable to retrieve machine[%s] machine info: %v. Using cached value", host, err)
 			entry.used = true
+			if entry.machineInfo == nil {
+				return nil, err
```
enlinxu: Same here. let's add another log message here to indicate that cache is used ,but machineinfo is nil. Otherwise we can't distinguish the case whether we populate the machineinfo from the cache or not.

PR #207, opened 2018-10-02
Updated with v6.2 turbo-go-sdk and turbo-api packages

pallavidn:
Using godep updated the turbo-go-sdk and turbo-api packages with the v6.2 tags.

### DIFF
--- a/vendor/github.com/turbonomic/turbo-api/pkg/client/request.go
+++ b/vendor/github.com/turbonomic/turbo-api/pkg/client/request.go
@@ -43,6 +43,7 @@ type Result struct {
 	status     string
 	body       string
 	err        error
+	cookies    map[string]*http.Cookie
 }
 
 func NewRequest(client HTTPClient, verb string, baseURL *url.URL, apiPath string) *Request {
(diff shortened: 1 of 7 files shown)


vendor/github.com/turbonomic/turbo-go-sdk/pkg/service/tap_service.go
```diff
@@ -70,6 +70,12 @@ func (tapService *TAPService) addTarget(isRegistered chan bool) {
 
 func (tapService *TAPService) ConnectToTurbo() {
 	glog.V(4).Infof("[ConnectToTurbo] Enter ******* ")
+
+	if tapService.RESTClient == nil {
+		glog.V(1).Infof("TAP service cannot be started - cannot create target")
```
maxwangvmt: It should be an error message, isn't it?

vendor/github.com/turbonomic/turbo-api/pkg/client/request.go
```diff
@@ -183,10 +184,10 @@ func (r *Request) request(fn func(*http.Response)) error {
 			req.Header.Set(key, value)
 		}
 	}
-	// Set basic authentication for the request if there is any.
-	if r.basicAuth != nil {
-		req.SetBasicAuth(r.basicAuth.username, r.basicAuth.password)
-	}
+	//// Set basic authentication for the request if there is any.
```
maxwangvmt: Remove the commented code if it's not kept on purpose.

pallavidn: This is a change in the turbo-api and should be handled there.

vendor/github.com/turbonomic/turbo-go-sdk/pkg/service/tap_service.go
pallavidn: This is a change in the turbo-go-sdk and should be handled there.

PR #208, opened 2018-10-03
Update master branch dependencies with latest versions for k8s.io

pallavidn:
Update master branch dependencies with latest (1.11) version for k8s.io using godep.
The versions for turbo-go-sdk, turbo-api were also updates in the godep.json file.


maxwangvmt:
There is some failing check from travis about: vendor/golang.org/x/oauth2/internal/token.go:21:2.

I am not sure the root cause but one question is why we have this dependency as we replaced it with gorilla for websocket client in turbo-go-sdk. Do we use golang.org/x/net anywhere?

PR #218, opened 2018-10-15
Add openshift scc flag to allow pod action execution

maxwangvmt:
Currently for Openshift, kubeturbo will only execute actions for pods with scc="restricted".

The change here is to add a scc flag (scc-support) to allow users to specify which scc's to allow action executions for Openshift clusters.

For example, if no scc flag included, the default is to only allow "restricted". To specify the flag, the allowed scc's should be listed: scc-support=scc1,scc2,scc3. The following are some examples to specify the scc flag.

--scc-support=restricted,anyuid,foo (allow action execution if pod has scc as one of the three)
--scc-support=.* (allow all scc's)

### DIFF
--- a/pkg/action/executor/resize_container.go
+++ b/pkg/action/executor/resize_container.go
@@ -375,7 +375,7 @@ func (r *ContainerResizer) preActionCheck(resizeSpec *containerResizeSpec, pod *
 
 	// Check if the pod privilege is supported
 	if !util.SupportPrivilegePod(pod, r.sccAllowedSet) {
-		err := fmt.Errorf("The pod %s has privilege requirement unsupported", fullName)
+		err := fmt.Errorf("Pod %s has unsupported SCC", fullName)
 		glog.Error(err)
 		return err
 	}
--- a/pkg/action/executor/resize_container.go
+++ b/pkg/action/executor/resize_container.go
@@ -38,6 +38,7 @@ type ContainerResizer struct {
 	k8sVersion                 string
 	noneSchedulerName          string
 	enableNonDisruptiveSupport bool
+	sccAllowedSet              map[string]struct{}
 
 	spec *containerResizeSpec
 }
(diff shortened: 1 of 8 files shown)


pkg/action/util/util.go
```diff
@@ -139,12 +139,18 @@ func AddAnnotation(pod *api.Pod, key, value string) {
...
+		return true
+	}
+
+	if _, ok := sccAllowedSet[osSccAllowAll]; ok {
+		return true
+	}
+
+	podScc := pod.Annotations[osSccAnnotation]
+	if _, ok := sccAllowedSet[podScc]; !ok {
+		glog.Warningf("The pod %s has unsupported privilege %s", pod.Name, podScc)
```
enlinxu: The log should be something like "Target pod has unsupported SCC, please add the SCC to --scc-support argument in kubeturbo yaml to allow execution."

pkg/action/executor/resize_container.go
```diff
@@ -372,7 +374,7 @@ func (r *ContainerResizer) preActionCheck(resizeSpec *containerResizeSpec, pod *
 	fullName := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 
 	// Check if the pod privilege is supported
-	if !util.SupportPrivilegePod(pod) {
+	if !util.SupportPrivilegePod(pod, r.sccAllowedSet) {
 		err := fmt.Errorf("The pod %s has privilege requirement unsupported", fullName)
```
enlinxu: The log should be something like "Target pod has unsupported SCC, please add the SCC to --scc-support argument in kubeturbo yaml to allow execution."

pkg/action/util/util.go
```diff
@@ -17,8 +17,8 @@ import (
 )
 
 const (
-	osSccAnnotation          = "openshift.io/scc"
-	supportedOsSccAnnotation = "restricted"
+	osSccAnnotation = "openshift.io/scc"
+	osSccAllowAll   = ".*"
```
enlinxu: Changed it to * based on our discussion yesterday.

PR #237, opened 2019-02-01
helm chart and readme updates

evat-pm:
Helm Chart updates:
1. Added targetName as an input parameter to allow user to differentiate multiple clusters
2. More environments will be OpenShift or k8s 1.11+, so have the default kubelet settings of https = true and port at 10250.  User will need to intentionally change values if running k8s 1.10 or older
3. turbo-volume does not need a unique suffix variable

deploy/readme updates:
1. Sample yamls are updated
2. Added wording to explain use of targetName configMap parameter (Note more main readme updates coming)


deploy/kubeturbo/templates/deployment.yaml
```diff
@@ -40,13 +40,13 @@ spec:
           - --k8sVersion=1.5
           {{- end }}
           volumeMounts:
-          - name: turbo-volume-{{ .Release.Name }}
```
esara: I think this would create a conflict where multiple kubeturbo instances are deployed in the same cluster

deploy/kubeturbo/templates/deployment.yaml
```diff
@@ -40,13 +40,13 @@ spec:
...
+          - name: turbo-volume
             mountPath: /etc/kubeturbo
             readOnly: true
           - name: varlog
             mountPath: /var/log
       volumes:
-      - name: turbo-volume-{{ .Release.Name }}
```
esara: same as above

deploy/kubeturbo/templates/configmap.yaml
```diff
@@ -15,6 +15,7 @@ data:
                 "opsManagerPassword": "{{ .Values.restAPIConfig.opsManagerPassword }}"
             }
         },
+        {{- if .Values.targetConfig {{-f if .Values.targetConfig.targetName }} }}
```
esara: did you include the next section as optional inside the if statement?

PR #239, opened 2019-02-25
Propagate action errors

ading1977:
### Problem
Kubeturbo action errors are not propagated to the UI. What's worse, the detailed error messages are not logged in the kubeturbo log either. Today when kubeturbo move/resize a pod, it clones a pod first on the desired node, or to the desired size on the same node. If the clone fails for whatever reasons, kubeturbo deletes the cloned pod, and all errors related to the deleted pod cannot be retrieved any more.

### Solution
Detailed error messages can be obtained from k8s events API, however, some of the error messages can be long, and not suitable to be displayed on the GUI. These messages should only be logged in kubeturbo log.
For example:
```
I0225 16:17:46.113839       1 pod_util.go:305] Pod task-pv-pod-1b1l7hjatjl31: pulling image "nginx"
I0225 16:17:46.113853       1 pod_util.go:305] Pod task-pv-pod-1b1l7hjatjl31: Successfully pulled image "nginx"
I0225 16:17:46.113857       1 pod_util.go:305] Pod task-pv-pod-1b1l7hjatjl31: Created container
E0225 16:17:46.113861       1 pod_util.go:303] Pod task-pv-pod-1b1l7hjatjl31: Error: failed to start container "task-pv-container": Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"rootfs_linux.go:58: mounting \\\"/mnt/data\\\" to rootfs \\\"/var/lib/docker/overlay2/3d3bb4345288eda3e55560ec47479df91cb85be9b17ed4b5176d005d4258c6fd/merged\\\" at \\\"/var/lib/docker/overlay2/3d3bb4345288eda3e55560ec47479df91cb85be9b17ed4b5176d005d4258c6fd/merged/usr/share/nginx/html\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
E0225 16:17:46.113868       1 pod_util.go:303] Pod task-pv-pod-1b1l7hjatjl31: Back-off restarting failed container

```

For error messages displayed on GUI, propose the following format:
```
// getPodWarnings gets a list of short warning messages that belong to the given pod
// These messages will be concatenated and sent to the UI
// The warning messages are logged in the following order:
//   - Pod states and reasons
//   - Pod conditions and reasons (false conditions only)
//   - Container states and reasons (abnormal states only)
// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-and-container-status
```
The **states** and **reasons** message from k8s API already have high level descriptions of the problems, and is short in general, so propose to use them for displaying. 

For example:
screen shot 2019-02-24 at 8 17 54 pm
screen shot 2019-02-25 at 11 17 08 am

### Changes
* Consolidate ``waitForReady`` and ``checkPod`` into a single function ``WaitForPodReady`` and move it into ``pod_util.go``
* Remove unnecessary pod status checking in move and resize (step 3 in ``Execute`` function)
* Remove ``moveControllerPod`` and ``moveBarePod``, just use ``movePod`` as the logic is the same
* Remove ``resizeControllerContainer`` and ``resizeBarePodContainer``, just use ``resizeContainer`` as the logic is the same
* Enhance error message format according to [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
* Remove duplicated log messages
* Add the logic to get pod warning messages (for UI display) and pod events (for logging)
* Modify the code to propagate the messages

### DIFF
--- a/pkg/action/executor/rescheduler.go
+++ b/pkg/action/executor/rescheduler.go
@@ -2,14 +2,13 @@ package executor
 
 import (
 	"fmt"
+
 	"github.com/golang/glog"
-	"time"
 
 	"github.com/turbonomic/kubeturbo/pkg/action/util"
 	api "k8s.io/api/core/v1"
 
 	podutil "github.com/turbonomic/kubeturbo/pkg/discovery/util"
-	goutil "github.com/turbonomic/kubeturbo/pkg/util"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
 )
 
@@ -25,7 +24,7 @@ func NewReScheduler(ae TurboK8sActionExecutor, sccAllowedSet map[string]struct{}
 	}
 }
 
-//Note: the error info will be shown in UI
+// Execute executes the move action. The error message will be shown in UI.
 func (r *ReScheduler) Execute(input *TurboActionExecutorInput) (*TurboActionExecutorOutput, error) {
 	actionItem := input.ActionItem
 	pod := input.Pod
@@ -33,26 +32,16 @@ func (r *ReScheduler) Execute(input *TurboActionExecutorInput) (*TurboActionExec
 	//1. get target Pod and new hosting Node
 	node, err := r.getPodNode(actionItem)
 	if err != nil {
-		glog.Errorf("Failed to execute pod move: failed to get target pod or new hosting node: %v", err)
-		return &TurboActionExecutorOutput{}, fmt.Errorf("Failed")
-	}
-
-	//2. move pod to the node
-	npod, err := r.reSchedule(pod, node)
-	if err != nil {
-		glog.Errorf("Failed to execute pod move: %v\n %++v", err, actionItem)
+		glog.Errorf("Failed to execute pod move: %v.", err)
 		return &TurboActionExecutorOutput{}, err
 	}
 
-	//3. check Pod
-	fullName := util.BuildIdentifier(npod.Namespace, npod.Name)
-	nodeName := npod.Spec.NodeName
-	glog.V(2).Infof("Begin to check pod move for pod[%v]", fullName)
-	if err = r.checkPod(npod, nodeName); err != nil {
-		glog.Errorf("Checking pod move failed: pod[%v] failed: %v", fullName, err)
-		return &TurboActionExecutorOutput{}, fmt.Errorf("Check Failed")
+	//2. move pod to the node and check move status
+	npod, err := r.reSchedule(pod, node)
+	if err != nil {
+		glog.Errorf("Failed to execute pod move: %v.", err)
+		return &TurboActionExecutorOutput{}, err
 	}
-	glog.V(2).Infof("Checking pod move succeeded: pod[%v] is on node[%v].", fullName, nodeName)
 
 	return &TurboActionExecutorOutput{
 		Succeeded: true,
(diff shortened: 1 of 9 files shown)


pkg/discovery/util/pod_util.go
```diff
@@ -276,3 +281,127 @@ func GetPodGrandInfo(kclient *client.Clientset, pod *api.Pod) (string, string, e
...
+// These events can be very long, and will only be written to the log file
+func getPodEvents(kubeClient *client.Clientset, namespace, name string) (podEvents []podEvent) {
+	podEvents = []podEvent{}
+	// Get the pod
+	pod, err := kubeClient.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		return
+	}
+	// Get all events
+	events, err := kubeClient.CoreV1().Events(namespace).List(metav1.ListOptions{})
```
maxwangvmt: Can we use the ListOptions to select only the events relevant? Some example using ListOptions is as below:
podList, err := podClient.List(meta_v1.ListOptions{
		FieldSelector: "status.phase=" + string(phase) + ",metadata.uid=" + uid,
	})

ading1977: Resolved

pkg/discovery/util/pod_util.go
```diff
@@ -276,3 +281,127 @@ func GetPodGrandInfo(kclient *client.Clientset, pod *api.Pod) (string, string, e
...
+func WaitForPodReady(client *client.Clientset, namespace, podName, nodeName string,
+	retry int, interval time.Duration) error {
+	// check pod readiness with retries
+	timeout := time.Duration(retry+1) * interval
+	err := goutil.RetrySimple(retry, timeout, interval, func() (bool, error) {
+		return checkPodNode(client, namespace, podName, nodeName)
+	})
+	// log a list of unique events that belong to the pod
+	// warning events are logged in Error level, other events are logged in Info level
+	podEvents := getPodEvents(client, namespace, podName)
```
pallavidn: Why is the *api.Pod not passed in with checkPodNode() and getPodEvents(), so the two calls to the api client to get the pod can be avoided?

ading1977: For ``checkPodNode()``, the api will need to be called each time to get the updated pod status.
For ``getPodEvents()``, the api is called once.

I think in the future, if we ever run into performance issue, we probably want to use k8s watch api to avoid polling.

pkg/action/executor/rescheduler.go
```diff
@@ -25,139 +24,97 @@ func NewReScheduler(ae TurboK8sActionExecutor, sccAllowedSet map[string]struct{}
...
+		glog.Errorf("Failed to execute pod move: %v.", err)
+		return &TurboActionExecutorOutput{}, err
 	}
 
-	//2. move pod to the node
+	//2. move pod to the node and check move status
 	npod, err := r.reSchedule(pod, node)
 	if err != nil {
-		glog.Errorf("Failed to execute pod move: %v\n %++v", err, actionItem)
+		glog.Errorf("Failed to execute pod move: %v.", err)
```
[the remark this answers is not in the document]
ading1977: Your point is valid that we should log the ``actionItem`` (in Error level) when action execution fails. I have updated the code to log the ``actionItem`` in ``ActionHander`` -> ``execute()``:
```
	output, err := worker.Execute(input)

	if err != nil {
		glog.Errorf("Failed to execute action %v on %s: %+v.",
			actionType, actionItem.GetTargetSE().GetEntityType(), actionItem)
		return err
	}
```

I removed the logging before because the ``actionItem`` has already been logged in the go sdk ``client_protobuf_endpoint.go`` in Info level when the action is received. The message is rather huge occupying two screens. It probably makes more sense to bump the logging to V(4) in go sdk. This can be addressed later if needed.

pkg/action/executor/rescheduler.go
```diff
@@ -25,139 +24,97 @@ func NewReScheduler(ae TurboK8sActionExecutor, sccAllowedSet map[string]struct{}
...
+		glog.Errorf("%v.", err)
 		return nil, err
 	}
 
-	//1. get node from properties
+	var err error
+	var node *api.Node
+
+	//3. get node from properties
 	node, err = util.GetNodeFromProperties(r.kubeClient, hostSE.GetEntityProperties())
```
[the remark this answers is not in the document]
ading1977: Updated:

```
	//3. get node from properties
	node, err := util.GetNodeFromProperties(r.kubeClient, hostSE.GetEntityProperties())
	if err == nil {
		glog.V(2).Infof("Get node(%v) from properties.", node.Name)
		return node, nil
	}
```

pkg/discovery/util/pod_util.go
```diff
@@ -276,3 +281,127 @@ func GetPodGrandInfo(kclient *client.Clientset, pod *api.Pod) (string, string, e
...
+		if pe.eType == api.EventTypeWarning {
+			glog.Errorf("Pod %s: %s", podName, pe.message)
+		} else {
+			glog.V(2).Infof("Pod %s: %s", podName, pe.message)
+		}
+	}
+	return err
+}
+
+func checkPodNode(kubeClient *client.Clientset, namespace, podName, nodeName string) (bool, error) {
```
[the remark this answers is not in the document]
ading1977: Updated:

```
// checkPodNode checks the readiness of a given pod
// The boolean return value indicates if this function needs to be retried
func checkPodNode(kubeClient *client.Clientset, namespace, podName, nodeName string) (bool, error) {
```

PR #240, opened 2019-03-01
Add support for detecting and sending daemon pod and master node status.

bcervenak:


pkg/discovery/util/node_util.go
```diff
@@ -59,3 +59,13 @@ func GetNodeIP(node *api.Node) (string, error) {
...
 }
+
+// Check whether the node is a master
+func NodeIsMaster(node *api.Node) bool {
+	return detectors.IsMasterDetected(node.Name, node.ObjectMeta.Labels)
+}
+
+// Returns whether the node is controllable
+func NodeIsControllable(node *api.Node) bool {
+	return !NodeIsMaster(node)
```
enlinxu: Add some logging here to help future troubleshooting when we detect the master.

pkg/discovery/util/pod_util.go
```diff
@@ -44,6 +45,14 @@ func IsControllableFromAnnotation(annotations map[string]string) bool {
 	return true
 }
 
+// Returns a boolean that indicates whether the given pod is a daemon pod.  A daemon pod
+// is a pod that resides in a system namespace or was created by a DaemonSet.  These pods
+// are not suspendable, clonable, or movable, and they are not considered when counting
+// customers of a supplier when checking whether the supplier can suspend.
+func Daemon(pod *api.Pod) bool {
+	return isPodCreatedBy(pod, Kind_DaemonSet) || detectors.IsDaemonDetected(pod.Name, pod.Namespace)
```
enlinxu: Add some logging here to help future troubleshooting when we detect the daemon.

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -124,8 +124,10 @@ func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]*proto.E
 		if !controllable {
 			glog.V(3).Infof("Pod %v is not controllable.", displayName)
 		}
+		daemon := util.Daemon(pod)
 		entityDTOBuilder.ConsumerPolicy(&proto.EntityDTO_ConsumerPolicy{
 			Controllable: &controllable,
+			Daemon:       &daemon,
```
enlinxu: I think one of the main motivation of not counting daemons as overhead in Turbo server is we still want to resize daemon (at least RECOMMEND), shall we mark them not controllable here?

bcervenak: As discussed, the controllable and daemon settings are independent of each other, which allows for daemons to be resized.

pkg/discovery/detectors/detector_config.go
```diff
@@ -0,0 +1,101 @@
...
+	// Pre-compile all regular expressions and ensure that they are valid
+
+	// Master node detection by node name
+	masterNodeNamePattern = buildRegexFromList(mconfig.NodeNamePatterns)
+
+	// Master node detection by label
+	masterLabelKeys = make([]*regexp.Regexp, len(mconfig.NodeLabels))
+	masterLabelValues = make([]*regexp.Regexp, len(mconfig.NodeLabels))
+	for i, entry := range mconfig.NodeLabels {
+		masterLabelKeys[i] = regexp.MustCompile(entry.Key)
```
ading1977: It seems that ``regex.MustCompile`` will panic if the expression cannot be parsed? Is it the intention to crash kubeturbo when there is invalid configuration? Or shall we use ``regex.Compile`` to catch and log the error and exit more elegantly?

bcervenak: Yes, it's my intention to crash.  I normally wouldn't do this, but I wanted a faulty configuration be be obvious.  If I simply generated an error in the log, nobody would ever notice the problem.  If the pod repeatedly crashes, it will cause the user to review the logs.

bcervenak: I just reread your comment, and yes, I agree that I should at least generate a log message before crashing/exiting.

enlinxu: In my opinion, if the regex is not parsed, we should use default value instead of crashing it. The fact that the regex doesn't work is not a catastrophe.

ading1977: If the regex doesn’t parse there could be undesirable consequences. User might have made a typo in the master node pattern, if kubeturbo does not exit, the master node will be controllable.

maxwangvmt: I would prefer to crash in this case. This is some error in the configMap. If kubeturbo crashes fast, we can notice it immediately and correct the error in the config.

pkg/discovery/util/node_util.go
bcervenak: Although NodeIsMaster already generated logging output, I changed that log message to V4 and added new log message here.

pkg/discovery/util/pod_util.go
bcervenak: Although the detector functions already generated logging output, I changed those log messages to V4 and added new log message here.

pkg/discovery/detectors/detector_config.go
bcervenak: As I stated earlier, we want the error to be obvious.  In addition, the various hosted Kubernetes clusters classify daemons and masters in different ways, so there are no reasonable defaults.

I now terminate kubeturbo by logging an error then using os.Exit(), so I am marking this resolved.

ading1977:
Loo

> LGTM. @maxwangvmt @ading1977 ?

Looks good to me.

PR #246, opened 2019-03-13
Suspend the correct target pod

ading1977:
Right now Turbo server makes a decision on the best container pod to suspend based on ROI. However, once the decision is made and sent to kubeturbo, kubeturbo implements the suspension by calling k8s API to decrease the number of replicas of the deployment. The actual pod that is suspended is determined by the Kubernetes deployment controller, which is very likely to be different than the one chosen by Turbo.

This pull request addresses this issue by terminating the designated container pod first before decreasing the number of replicas.

The following changes are made:
* Cleanup and reformat error messages
* Remove empty functions
* Small refactor of the code to make better use of the ``scaleHelper``
* Add the logic to suspend target pod first before scaling down the replica numbers
```
// scaleDeployment scales a deployment in the following steps:
// For scale in action:
// 1. Delete the target pod
// 2. Decrease the replicas of the deployment by one
// There can be a very short window of race condition between step 1 and step 2, at which time
// k8s deployment controller spins up a new pod. But as soon as the replica number is decreased,
// 1. Delete the target pod
// 2. Decrease the replicas of the deployment by one
// There can be a very short window of race condition between step 1 and step 2, at which time
// k8s deployment controller spins up a new pod. But as soon as the replica number is decreased,
// the newly started pod will be immediately terminated before any containers can have a chance
// to start.
//
// For scale out action:
// 1. Increase the replicas of the deployment by one
// TODO: In the future, if we know the target host of the pod to be provisioned, we can
// create a pod on the target host, and associate the pod with the deployment
func (helper *scaleHelper) scaleDeployment() error {
```
* Simplify logic to validate the resulting number of replicas after scale

### DIFF
--- a/pkg/action/executor/hscale_helper.go
+++ b/pkg/action/executor/hscale_helper.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"fmt"
+
 	"github.com/golang/glog"
 
 	"github.com/turbonomic/kubeturbo/pkg/util"
@@ -9,7 +10,7 @@ import (
 	kclient "k8s.io/client-go/kubernetes"
 )
 
-type updateReplicaNumFunc func(client *kclient.Clientset, nameSpace, name string, diff int32) error
+type scaleFunc func() error
 
 type scaleHelper struct {
 	client    *kclient.Clientset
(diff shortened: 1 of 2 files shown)


pkg/action/executor/hscale_helper.go
```diff
@@ -36,128 +37,156 @@ func NewScaleHelper(client *kclient.Clientset, nameSpace, podName string) (*scal
...
+			return 0, err
+		}
+	}
 	return result, nil
 }
 
-// update the number of pod replicas for ReplicationController
-func updateRCReplicaNum(client *kclient.Clientset, namespace, name string, diff int32) error {
-	rcClient := client.CoreV1().ReplicationControllers(namespace)
+// scaleRS scales a ReplicationController
```
maxwangvmt: Typo: scaleRS -> scaleRC

ading1977: Updated.

PR #258, opened 2019-04-18
Changing to use go mod to manage dependencies instead of godep, and adding the initial version of node scaler  code

chlam4:
This pull request changes our dependency management tool from godep to "go mod", which is a new official dependency management tool and fully integrated in GO.  Still claimed to be preliminary, but it helps us easily add the cluster api dependencies, which otherwise would require a lot of manual work with godep and we had struggled about for a long while.

To read more about Go Modules: https://blog.golang.org/using-go-modules


chlam4:
> LGTM. There's some format change required for Travis to pass.

Thanks for pointing this out.  I'll update with the format fix.

PR #262, opened 2019-04-26
Add initial version

lin007pro:


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -219,6 +220,11 @@ func (s *VMTServer) Run(_ []string) error {
 	// the certificate issue of 'doesn't contain any IP SANs'.
 	// See https://github.com/kubernetes/kubernetes/issues/59372
 	kubeletClient := s.createKubeletClientOrDie(kubeConfig, !isOpenshift)
+	caClient, err := clusterclient.NewForConfig(kubeConfig)
+	if err != nil {
+		glog.Errorf("Failed to generate correct TAP config: %v", err.Error())
+		os.Exit(1)
```
chlam4: Does any k8s cluster always have the cluster API deployed?  I assume it's optional, so perhaps not a good idea to exit Kubeturbo?

lin007pro: I've tested it without any cluster API present, and we didn't exit.
This is here to detect fatal errors creating the cluster API client.

pkg/action/action_handler.go
```diff
@@ -36,11 +37,13 @@ var (
 	turboActionContainerResize     turboActionType = turboActionType{proto.ActionItemDTO_RIGHT_SIZE, proto.EntityDTO_CONTAINER}
 	turboActionContainerPodSuspend turboActionType = turboActionType{proto.ActionItemDTO_SUSPEND, proto.EntityDTO_CONTAINER_POD}
 
-	//turboActionUnbind          turboActionType = "unbind"
+	turboActionMachineProvision turboActionType = turboActionType{proto.ActionItemDTO_PROVISION, proto.EntityDTO_VIRTUAL_MACHINE}
+	turboActionMachineSuspend   turboActionType = turboActionType{proto.ActionItemDTO_SUSPEND, proto.EntityDTO_VIRTUAL_MACHINE}
```
chlam4: if there isn't a valid caClient, then kubeturbo probably shouldn't declare support of the VM scaling actions?

lin007pro: We need to handle the situation where the cluster API is enabled after both kubernetes cluster and kubeturbo are deployed.
Which means that kubeturbo should be able to utilize the cluster API being deployed at any point of time without being restarted.

enlinxu: For now, I wouldn't suggest to spend too much effort to figure out how to avoid restarting, since restart Kubeturbo is cheap,  it's just a probe and it's stateless. We just need to make sure we don't declare VM scaling action like Cheuk said, if the Cluster API is not there when Kubeturbo starts.

pkg/kubeclient/kubelet_client.go
```diff
@@ -215,14 +215,14 @@ func (client *KubeletClient) HasCacheBeenUsed(host string) bool {
...
 type KubeletConfig struct {
-	kubeConfig       *rest.Config
-	enableHttps      bool
-	allowTLSInsecure bool
-	port             int
-	timeout          time.Duration // timeout when fetching information from kubelet;
-	tlsTimeOut       time.Duration
+	kubeConfig           *rest.Config
+	enableHttps          bool
+	forceSelfSignedCerts bool
```
chlam4: I agree the original name `allowTLSInsecure` isn't easy to understand; I think it came from the `Insecure` flag in `TLSConfig` in `k8s.io/client-go`.  This new name `forceSelfSignedCerts` is far better, but still seems a little odd with the word "force".  Would `allowSelfSignedCerts` better?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -112,6 +117,7 @@ func (s *VMTServer) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.UseUUID, "stitch-uuid", true, "Use VirtualMachine's UUID to do stitching, otherwise IP is used.")
 	fs.IntVar(&s.KubeletPort, "kubelet-port", DefaultKubeletPort, "The port of the kubelet runs on")
 	fs.BoolVar(&s.EnableKubeletHttps, "kubelet-https", DefaultKubeletHttps, "Indicate if Kubelet is running on https server")
+	fs.BoolVar(&s.ForceSelfSignedCerts, "kubelet-force-selfsigned-cert", true, "Indicate if we must use self-signed cert")
```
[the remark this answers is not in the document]
chlam4: I'd also suggest the default be false, meaning we shouldn't allow self signed certs by default.  The issue (https://github.com/kubernetes/kubernetes/issues/59372) has been resolved.

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -219,6 +229,11 @@ func (s *VMTServer) Run() {
 	// the certificate issue of 'doesn't contain any IP SANs'.
 	// See https://github.com/kubernetes/kubernetes/issues/59372
 	kubeletClient := s.createKubeletClientOrDie(kubeConfig, !isOpenshift)
```
chlam4: Based on the fact that you said it worked for you when you changed `!isOpenShift` to `isOpenShift`, it means your server has valid CA certs and there shouldn't be a need to ForceSelfSignedCerts?

cmd/kubeturbo/app/kubeturbo_builder.go
ading1977: We also need to update our deployment helm/charts [templates/values](https://github.com/turbonomic/kubeturbo/tree/master/deploy) to reflect the new flags.

pkg/action/action_handler.go
```diff
@@ -142,45 +153,51 @@ func (h *ActionHandler) ExecuteAction(actionExecutionDTO *proto.ActionExecutionD
...
-
-	// After getting the lock, need to get the k8s pod again as the previous action could delete the pod and create a new one.
-	// In such case, the action should be applied on the new pod.
-	// Currently, all actions need to get its related pod. If not needed, the pod is nil.
-	pod := h.getRelatedPod(actionItem)
-
-	if pod == nil {
-		err := fmt.Errorf("Cannot find the related pod for action item %s", actionItem.GetUuid())
-		return err
+	if !clusterAPIAction {
```
ading1977: Why do we ignore locks for machine related actions? I think it will be desirable to sequentialize the actions such that no concurrent actions can be applied to the same entity (machine). This may require us to update the `getLock` function to get the lock key for machine.

pkg/action/executor/machine_scalers.go
```diff
@@ -0,0 +1,370 @@
...
+
+// waitForNodeDeprovisioning waits for the new node to be de-provisioned with timeout.
+func (controller *machineDeploymentController) waitForNodeDeprovisioning(machine *clusterv1.Machine) error {
+	deletedNName := machine.Name
+	descr := fmt.Sprintf("node %s deleted or exited Ready state", deletedNName)
+	return controller.waitForState(descr, controller.isNodeDeletedOrNotReady, deletedNName)
+}
+
+// waitForState Is the function that allows to wait for a specific state, or until it times out.
+func (controller *machineDeploymentController) waitForState(stateDesc string, f stateCheck, args ...interface{}) error {
```
ading1977: Is it possible to reuse the existing `RetrySimple` or `RetryDuring` function defined in `go_util.go` file?

cmd/kubeturbo/app/kubeturbo_builder.go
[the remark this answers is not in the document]
lin007pro: The current approach keeps the original semantics. So if the flag is omitted, the behavior will be unchanged from before.

pkg/kubeclient/kubelet_client.go
lin007pro: We don't "allow" anything. We force the self-signed certs usage, since we remove all the references to the certs we do have.

pkg/action/action_handler.go
lin007pro: We do that because we need to fail fast here, rather than wait. That was a requirement that came from Endre.

pkg/action/executor/machine_scalers.go
lin007pro: The semantics are subtly different. Here I'm waiting for the state to happen, rather than retry several times. I could shoehorn it in, but the result will likely be more confusing, sorry.

pkg/action/action_handler.go
```diff
@@ -142,45 +153,51 @@ func (h *ActionHandler) ExecuteAction(actionExecutionDTO *proto.ActionExecutionD
...
+			return err
+		} else {
+			// Unlock the entity after the action execution is finished
+			defer glog.V(4).Infof("Action %s: releasing lock", actionItem.GetUuid())
+			defer lock.ReleaseLock()
+			lock.KeepRenewLock()
+		}
 	}
 
 	input := &executor.TurboActionExecutorInput{
```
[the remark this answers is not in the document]
lin007pro: I only lock under certain circumstances, and I need to unlock under those circumstances as well. The unlock happens right after the lock here.

pkg/action/action_handler.go
```diff
@@ -108,6 +110,15 @@ func (h *ActionHandler) registerActionExecutors() {
 
 	containerResizer := executor.NewContainerResizer(ae, c.kubeletClient, c.sccAllowedSet)
 	h.actionExecutors[turboActionContainerResize] = containerResizer
+
+	// Only register the actions when API client is non-nil.
+	if ok, err := executor.IsClusterAPIEnabled(c.cApiClient, c.kubeClient); ok && err != nil {
```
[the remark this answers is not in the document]
lin007pro: Thank you.

ading1977:
Looks good to me with the latest change to lock the action and fail fast when the second action on the same entity is being executed.

Currently, the market will not resize VM that hosts containers, so it is impossible to have both resize and horizontal scale actions on the same entity at the same time. We should be ok to fail fast.

PR #263, opened 2019-05-03
OM-44460 Add CPURequestAllocation and MemoryRequestAllocation into VDC

ading1977:
The code change is big, so I have splitted them into multiple commits. Please review each individual commit.


pkg/discovery/repository/kube_cluster.go
```diff
@@ -176,21 +173,23 @@ func NewKubeNode(apiNode *v1.Node, clusterName string) *KubeNode {
...
 
 	// Node compute resources and properties
-	nodeEntity.UpdateResources(apiNode)
+	nodeEntity.updateResources(apiNode)
 	return nodeEntity
 }
 
 // Set the compute resources and IP property in the node entity
-func (nodeEntity *KubeNode) UpdateResources(apiNode *v1.Node) {
+func (nodeEntity *KubeNode) updateResources(apiNode *v1.Node) {
```
enlinxu: Why do we want to change here? I thought UpdateResources is the right convention?

ading1977: `updateResources` is only called within the package. We are not exporting it to be used by any other package, so we should use lower case according to [Go language specification](https://golang.org/ref/spec#Exported_identifiers)

ading1977:
Format changes are checked with gofmt.
We cannot use golint as there are too many style mistakes in our existing code base :-(, that will require a separate effort to correct them.

pkg/discovery/repository/kube_cluster.go
enlinxu: I see, I wasn't aware of it. Then I am good with the review. Ship it!

PR #275, opened 2019-06-17
Add stitching metadata for kubernetes VM for stitching with XL server

ading1977:
This pull request adds stitching metadata for discovered kubernetes nodes so that they can be stitched to VMs discovered by other targets:
* This pull request is dependent on the [turbo-go-sdk pull 69](https://github.com/turbonomic/turbo-go-sdk/pull/69)
* Stitching can be done via either UUID or IP address

Tests have been done by stitching k8s node with vCenter VMs and AWS VMs, and verifying that the correct commodities values are patched.

**vCenter stitched environment**
image

Will update the diff with changes to go.mod and vendor package once turbo-go-sdk change has been committed.


ading1977:
The build fails as it depends on the change in turbo-go-sdk to be committed, which is still in review stage.

pkg/registration/supply_chain_factory.go
```diff
@@ -110,6 +124,53 @@ func (f *SupplyChainFactory) createSupplyChain() ([]*proto.TemplateDTO, error) {
...
+			ExternalMatchingFieldWithDelimiter(VMIPFieldName, VMIPFieldPaths, ",")
+	default:
+		return nil, fmt.Errorf("stitching property type %s is not supported",
+			f.stitchingPropertyType)
+	}
+	return mergedEntityMetadataBuilder.
+		PatchSoldMetadata(proto.CommodityDTO_CLUSTER, fieldsCapactiy).
+		PatchSoldMetadata(proto.CommodityDTO_VMPM_ACCESS, fieldsCapactiy).
+		PatchSoldMetadata(proto.CommodityDTO_VCPU, fieldsUsedCapacityPeak).
+		PatchSoldMetadata(proto.CommodityDTO_VMEM, fieldsUsedCapacity).
```
chlam4: Now that we patch vcpu peak in https://github.com/turbonomic/kubeturbo/pull/255, shall we also patch vmem peak?

ading1977: Thanks @chlam4 , yes we should patch VMEM peak as well. We set peak as the used value for sold commodities [here](https://github.com/turbonomic/kubeturbo/blob/6e9a5a9108e569d046e7e8cb85b526ca92703701/pkg/discovery/dtofactory/general_builder.go#L156).

PR #278, opened 2019-06-20
deploy readme bad link fix

evat-pm:
fixed typo in link to kubeturbo-operator


evat-pm:
Updated with sample yaml links

PR #281, opened 2019-06-26
Use providerID as a link between Node and a Machine

lin007pro:


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -254,7 +258,8 @@ func (s *VMTServer) Run() {
 		WithDiscoveryInterval(s.DiscoveryIntervalSec).
 		WithValidationTimeout(s.ValidationTimeout).
 		WithValidationWorkers(s.ValidationWorkers).
-		WithSccSupport(s.sccSupport)
+		WithSccSupport(s.sccSupport).
+		WithCAPINapespace(s.ClusterAPINamespace)
```
ading1977: Please correct typo

pkg/kubeturbo_service_config.go
```diff
@@ -99,3 +100,8 @@ func (c *Config) WithSccSupport(sccSupport []string) *Config {
 	c.SccSupport = sccSupport
 	return c
 }
+
+func (c *Config) WithCAPINapespace(CAPINamespace string) *Config {
```
ading1977: Please correct typo

pkg/action/action_handler.go
```diff
@@ -46,9 +46,10 @@ type ActionHandlerConfig struct {
 	kubeletClient  *kubeclient.KubeletClient
 	StopEverything chan struct{}
 	sccAllowedSet  map[string]struct{}
+	cAPINamespace  string
 }
 
-func NewActionHandlerConfig(kubeClient *client.Clientset, kubeletClient *kubeclient.KubeletClient, sccSupport []string) *ActionHandlerConfig {
+func NewActionHandlerConfig(cApiNamespace string, kubeClient *client.Clientset, kubeletClient *kubeclient.KubeletClient, sccSupport []string) *ActionHandlerConfig {
```
ading1977: Maybe move the **cApiNamespace** to the last of the function parameter?

cmd/kubeturbo/app/kubeturbo_builder.go
lin007pro: Hi,
I don't see a typo. I agree we need to allow for the logic to handle other providers, although the Cluster API is not provider specific from coding standpoint.
I propose to address this when we know more.

pkg/action/action_handler.go
lin007pro: I prefer to have it as a first one. The emphasis is important, IMHO.

PR #291, opened 2019-07-15
Fix Cluster API client passing

lin007pro:


pkg/registration/registration_client.go
```diff
@@ -112,6 +112,16 @@ func (rClient *K8sRegistrationClient) GetActionPolicy() []*proto.ActionPolicyDTO
 
 	rClient.addActionPolicy(ab, app, appPolicy)
 
+	// 4. node: move, provision; not resize;
+	node := proto.EntityDTO_VIRTUAL_MACHINE
+	nodePolicy := make(map[proto.ActionItemDTO_ActionType]proto.ActionPolicyDTO_ActionCapability)
+	nodePolicy[proto.ActionItemDTO_MOVE] = supported
```
enlinxu: kubeturbo is not handling node move, let's remove this line or change it to notsupported

pkg/registration/registration_client_test.go
```diff
@@ -65,6 +66,12 @@ func TestK8sRegistrationClient_GetActionPolicy(t *testing.T) {
 	expected_app[provision] = recommend
 	expected_app[suspend] = recommend
 
+	expected_node := make(map[proto.ActionItemDTO_ActionType]proto.ActionPolicyDTO_ActionCapability)
+	expected_node[move] = supported
```
enlinxu: Same here, remove this line or change it to notsupported

ading1977:
@lin007pro , I think what you are seeing is because you have a stitched environment, in which case, the actionDTO sent from the server contains the real VM name. However, if you only have a standalone kubeturbo target (i.e., remove the AWS target), you probably still need the original logic in order to find the machine from node. Can you please confirm?

Also, please add some description to this pull request, and update the title to correct the typo (passing -> parsing).

lin007pro:
The DTO request contains the machine name, which is expected by the Cluster API.
Fix the passing of the Cluster API client down the line properly.

PR #305, opened 2019-08-17
kubeturbo operator 6.4 update

esara:


esara:
> nit: Maybe we should have both kubeturbo and kubeturbo-operator have the same tagging convention?

I was considering the same first, but then I thought that the operator will not have the same lifecycle as kubeturbo, if we want to rev kubeturbo, we can just change the CR without changing the operator

enlinxu:
That makes sense. Feel free to merge!

PR #310, opened 2019-09-06
OM-50212: Create a static group for master nodes

pallavidn:
Create a static group of master nodes in a kubernetes cluster.
Master nodes will be detected using node name patterns or role descriptors.
Group is identified using the target Id, since Turbonomic server can manage multiple kubernetes clusters.

### DIFF
--- a/pkg/discovery/dtofactory/master_nodes_group_dto_builder.go
+++ b/pkg/discovery/dtofactory/master_nodes_group_dto_builder.go
@@ -1,0 +1,61 @@
+package dtofactory
+
+import (
+	"fmt"
+	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


pkg/discovery/dtofactory/master_nodes_group_dto_builder.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil
+	}
+
+	// static group
+	// group id created using the parent type, name and target identifier
+	groupId := "KubeMasterNodes"
+	id := fmt.Sprintf("%s-%s", groupId, builder.targetId)
+
+	displayName := fmt.Sprintf("%s[%s]",
+		groupId, builder.targetId)
```
chlam4: Nit: no need to split into 2 lines?

PR #320, opened 2019-09-14
fix docker build and move debug dockerfile under build

esara:


build/Dockerfile.debug
```diff
@@ -2,4 +2,4 @@ FROM turbonomic/kubeturbo:6.4.0
 COPY kubeturbo.debug /opt/turbonomic/bin/kubeturbo
```
ading1977: Shall we change `turbonomic/kubeturbo:6.4.0` to `turbonomic/kubeturbo` ?

build/Dockerfile.debug
```diff
@@ -2,4 +2,4 @@ FROM turbonomic/kubeturbo:6.4.0
...
 ADD dlv /usr/local/bin/dlv
 EXPOSE 40000
-ENTRYPOINT ["/usr/local/bin/dlv", "--listen=:40000", "--headless=true", "--api-version=2", "exec", "/opt/turbonomic/bin/kubeturbo", "--", "--turboconfig=/etc/kubeturbo/turbo.config", "--v=2", "--kubelet-https=true", "--kubelet-port=10250"]
+ENTRYPOINT ["/usr/local/bin/dlv", "--listen=:40000", "--headless=true", "--api-version=2", "exec", "/opt/turbonomic/bin/kubeturbo"]
```
ading1977: Why removing those options?

esara: because, the configmap feeds and and with this, I end up with duplicates
we should use the configmap, instead of hardcoding the entrypoint

ps: I was having problems on one of my clusters which was running the kubelet without https and I could not test it with the hardcoded options

build/Dockerfile.debug
esara: yes, that makes sense, I will include it in this PR

esara: actually, Michael made the debug tag configurable, so I think we are good

ading1977: The tag `${TURBO_REPO}/kubeturbo:${KUBE_VER}debug .` defines the output image which is used for pushing the debug image to personal registry. The base image is still hard coded as `turbonomic/kubeturbo:6.4.0` which seems odd. But it doesn't matter in the end, as the 6.4.0 binary is replaced by `kubeturbo.debug`.

esara: that makes sense, changed it

PR #324, opened 2019-09-18
Create HA node group and policy based on HANodeConfig policy in the configmap

ading1977:
This pull request creates HA node groups and anti-affinity policy for those groups based on the `HANodeConfig` policy specified in the `kubeturbo` configmap, for example:

```
apiVersion: v1
data:
  turbo.config: |-
    {
...
        "HANodeConfig": {
           "nodeRoles": [ "master", "worker" ]
        },
...
    }
kind: ConfigMap

```

### DIFF
--- a/pkg/discovery/detectors/detector_config_test.go
+++ b/pkg/discovery/detectors/detector_config_test.go
@@ -158,29 +158,31 @@ type TestConfig struct {
 	X2                   // Stubbed configs.K8sTargetConfig
 	*MasterNodeDetectors `json:"masterNodeDetectors,omitempty"`
 	*DaemonPodDetectors  `json:"daemonPodDetectors,omitempty"`
+	*HANodeConfig        `json:"HANodeConfig,omitempty"`
 }
 
 func loadConfig(config string) (*TestConfig, error) {
 	var spec TestConfig
 	err := json.Unmarshal([]byte(config), &spec)
 	if err != nil {
-		return nil, fmt.Errorf("Unmarshall error :%v", err.Error())
+		return nil, fmt.Errorf("unmarshall error :%v", err.Error())
 	}
 	//fmt.Println(spec)
 	return &spec, nil
 }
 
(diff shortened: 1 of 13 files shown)


pkg/discovery/detectors/detector_config_test.go
```diff
@@ -230,7 +232,7 @@ var masterNodeTests = []MasterNodeTestCase{
 func TestDetectorConfig_detectMasterNode(t *testing.T) {
 	for scenarioIndex, scenarioName := range []string{"validFullConfig", "missingMasterNodeDetectors",
 		"missingDaemonPodDetectors", "missingBoth", "missingList", "emptySection"} {
-		setupTest(t, scenarioName)
+		_, _ = setupTest(scenarioName)
```
chlam4: Is the `_, _` necessary here?

pkg/discovery/detectors/detector_config_test.go
```diff
@@ -262,7 +264,7 @@ var daemonPodTests = []DaemonPodTestCase{
 func TestDetectorConfig_detectDaemon(t *testing.T) {
 	for scenarioIndex, scenarioName := range []string{"validFullConfig", "missingMasterNodeDetectors",
 		"missingDaemonPodDetectors", "missingBoth", "missingList", "emptySection"} {
-		setupTest(t, scenarioName)
+		_, _ = setupTest(scenarioName)
```
chlam4: Same question as above.

pkg/discovery/detectors/detector_config_test.go
ading1977: This is not necessary but nice to have IMO for the following reasons:
* Get rid of the unhandled error code inspection warning in IntelliJ IDE
* Communicate to the code reader that we are deliberately ignoring a potential error

pkg/discovery/detectors/detector_config_test.go
ading1977: Same as above.

PR #326, opened 2019-09-23
OM-50651: Create groups for containers by name per parent group

pallavidn:
We want to set the resize policy independently for user app containers and the corresponding side car containers in the pod running the user app.
This PR handles the creation of additional container groups by container names as detected in a pod for different kinds of parent groups such as deployment, stateful sets etc.
For each parent group, there will one a container group for all the containers running in the pods belonging to this group. Additionally there will a container group for each container name found in the pod. The consistent resize is set to true for these container groups only.
Tested that the different container groups are created for deployment, stateful sets and other owner/parent types.

### DIFF
--- a/pkg/discovery/dtofactory/group_dto_builder.go
+++ b/pkg/discovery/dtofactory/group_dto_builder.go
@@ -24,6 +24,13 @@ var (
 	}
 )
 
+var (
+	ENTITY_TYPE_MAP = map[metrics.DiscoveredEntityType]proto.EntityDTO_EntityType{
+		metrics.PodType:       proto.EntityDTO_CONTAINER_POD,
+		metrics.ContainerType: proto.EntityDTO_CONTAINER,
+	}
+)
+
 // New instance of groupDTOBuilder.
 // Input parameters are map of discovered EntityGroup instances and
 // the target identifier of the kubeturbo probe.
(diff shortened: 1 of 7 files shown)


pkg/discovery/dtofactory/group_dto_builder.go
```diff
@@ -42,47 +49,76 @@ func NewGroupDTOBuilder(entityGroupMap map[string]*repository.EntityGroup,
...
+				WithEntities(containerList).
+				WithDisplayName(displayName)
 
-			// group resize policy - currently only for stateful sets
+			// if parent group resize policy is consistent resize, set it for the container sub-group
 			_, exists := CONSISTENT_RESIZE_GROUPS[entityGroup.ParentKind]
 			if exists && entityGroup.ParentName != "" {
-				glog.V(4).Infof("%s: set group to resize consistently\n", entityGroup.GroupId)
+				glog.V(2).Infof("%s: set group to resize consistently\n", groupId)
 				groupBuilder.ResizeConsistently()
```
enlinxu: I don't quite understand the code here as well as the comment. 
"if parent group resize policy is consistent resize, set it for the container sub-group"

I think the parent group should always be consistent resize false (if there're more than one container)

pallavidn: The consistent resize is set to true only for container sub-groups whose parent type are part of the  CONSISTENT_RESIZE_GROUP map.

PR #330, opened 2019-10-02
OM-51005 Pods in daemon sets are now controllable, in order to allow them to resize.

bcervenak:
Previous to this change, pods in daemon sets were flagged as daemons and were also marked non-controllable. This prevented generation of any actions on them. This change retains the flagging of the pods as daemons (i.e., disables suspension, cloning, and moving), but makes them resizable.

Please note that now that we are shipping the kubeturbo configuration with no daemonPodDetectors defined, pods in daemon sets are the only entities that will be marked as daemons. All pods in kube-system, for example, will now be movable, suspendable, and clonable, regardless of whether that is the correct behavior. This behavior can still be changed by adding daemonPodDetectors in kubeturbo or defining a server-side policy.


evat-pm:
this description satisfies the requirements in https://vmturbo.atlassian.net/wiki/spaces/PMTES/pages/1022493585/Containers+Resizing+daemonSet+vertical+scaling

PR #335, opened 2019-10-09
Open shift cluster api support

irfanurrehman:
This PR updates the cluster-api versions to map to the 4.1.x openshift machine api.
It also updates the usage and logic to comply with the many breaking api changes that has happened across the versions.


irfanurrehman:
@ading1977 @chlam4 @enlinxu This has been tested to work and needs your comments.

pkg/action/executor/machine_scalers.go
```diff
@@ -53,69 +54,30 @@ func (client *k8sClusterApi) verifyClusterAPIEnabled() error {
...
 		return nil, err
 	}
 	for _, machine := range machineList.Items {
-		machineSpecTmp := strings.Split(*machine.Spec.ProviderID, "/")
-		if len(nodeSpecTmp) < 2 {
-			return nil, fmt.Errorf("Machine " + machine.Name + " has no valid provider ID")
-		}
-		machineProviderID := machineSpecTmp[len(machineSpecTmp)-1]
-		if machineProviderID == nodeProviderID {
+		if nodeName == machine.Status.NodeRef.Name {
```
enlinxu: I remembered when Michael was implementing the AWS version, the provider ID is the only attr that matches. Can we make it match both on Name or ProviderID.

pkg/action/executor/machine_scalers.go
```diff
@@ -130,9 +131,32 @@ func (controller *machineSetController) checkPreconditions() error {
...
 // executeAction scales a MachineSet by modifying its replica count
 func (controller *machineSetController) executeAction() error {
-	desiredReplicas := controller.machineSet.Status.Replicas + controller.request.diff
+	diff := controller.request.diff
+	client := controller.request.client
+	desiredReplicas := controller.machineSet.Status.Replicas + diff
 	controller.machineSet.Spec.Replicas = &desiredReplicas
-	machineSet, err := controller.request.client.machineSet.Update(controller.machineSet)
+
+	if diff < 0 {
```
chlam4: Just wanted to confirm: When diff is negative, is it always -1?  Meaning scaling down by 1?  If so, then all this makes sense.

irfanurrehman: Yes, thats right. Its always -1.

pkg/action/executor/machine_scalers.go
irfanurrehman: @enlinxu I don't think its a good idea. I think the `NodeRef` is an authoritative link in terms of the API. Using the providerID was anyways a hacky mechanism earlier.

ading1977:
@irfanurrehman , I am not sure if you have answered this or not, have you tested your solution in both stitched and non-stitched environment? The host names passed to kubeturbo are different under these two circumstances, and we need to support both cases.

Thanks,
Meng

irfanurrehman:
> @irfanurrehman , I am not sure if you have answered this or not, have you tested your solution in both stitched and non-stitched environment? The host names passed to kubeturbo are different under these two circumstances, and we need to support both cases.

@ading1977 Thanks for the additional info. Matter of fact I have not tested this under stitched env. I will figure out a mechanism to test it in a stitched env also.

irfanurrehman:
Updated a gist for demo steps [here](https://gist.github.com/irfanurrehman/614f0250d25a026101dbb648206ca168).
Please suggest a placeholder for this.
Additionally this is tested in both stitched and non stitched env and works fine.

irfanurrehman:
Rebased.

PR #342, opened 2019-10-15
Set consistent resize flag for container groups without any sub-group

pallavidn:
Container sub groups are created for if there are more than one type of container in a deployment. The group members of these sub groups are set for consistent resize. 
We need to apply this setting for deployments with only one type of container.

### DIFF
--- a/pkg/discovery/dtofactory/group_dto_builder.go
+++ b/pkg/discovery/dtofactory/group_dto_builder.go
@@ -109,7 +109,6 @@ func (builder *groupDTOBuilder) createGroupsByEntityType(entityGroup *repository
 	createSubGroups := hasSubGroups(entityGroup)
 	resizeFlag := resizeEntityType && consistentResizeGroupType(entityGroup) && !createSubGroups
 
-	fmt.Printf("%s:%s -> %s\n", groupId, entityType, resizeFlag)
 	parentGroup := builder.createGroup(entityGroup, entityType, groupId, protoType, memberList, resizeFlag)
 	if parentGroup != nil {
 		result = append(result, parentGroup)
--- a/pkg/discovery/dtofactory/group_dto_builder.go
+++ b/pkg/discovery/dtofactory/group_dto_builder.go
@@ -25,6 +25,13 @@ var (
 )
 
 var (
+	GROUP_RESIZE_FLAG_MAP = map[metrics.DiscoveredEntityType]bool{
+		metrics.PodType:       false,
+		metrics.ContainerType: true,
+	}
+)
+
+var (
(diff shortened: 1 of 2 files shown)


pkg/discovery/dtofactory/group_dto_builder.go
```diff
@@ -63,77 +70,146 @@ func (builder *groupDTOBuilder) BuildGroupDTOs() []*proto.GroupDTO {
...
-			result = append(result, groupDTO)
+	// group id for parent group
+	groupId := fmt.Sprintf("%s", entityGroup.GroupId)
 
-			glog.V(4).Infof("groupDTO  : %++v", groupDTO)
-		}
+	// resize setting for container groups based on the parent type of the group
+	// and if sub groups will be created.
+	// If sub groups are created, the resize setting is applied on the sub-groups
+	resizeFlag := resizeGroup && consistentResize(entityGroup) && resizeParentGroup(entityGroup, entityType)
```
enlinxu: I see multiple places that we are checking if it's a container group or not, like GROUP_RESIZE_FLAG_MAP and in resizeParentGroup, can we merge the logic?

pkg/discovery/dtofactory/group_dto_builder.go
```diff
@@ -63,77 +70,146 @@ func (builder *groupDTOBuilder) BuildGroupDTOs() []*proto.GroupDTO {
...
+	groupDTO, err := groupBuilder.Build()
+	if err != nil {
+		glog.Errorf("Error creating group dto  %s::%s", id, err)
+		return nil
+	}
+
+	return groupDTO
+}
+
+func consistentResize(entityGroup *repository.EntityGroup) bool {
```
ading1977: Can we simplify this function? We don't need to check empty string, for example, we can do
```
func consistentResize(entityGroup *repository.EntityGroup) bool {
        _, resize := CONSISTENT_RESIZE_GROUPS[entityGroup.ParentKind]
        return resize
}
```

pkg/discovery/dtofactory/group_dto_builder.go
```diff
@@ -63,77 +70,146 @@ func (builder *groupDTOBuilder) BuildGroupDTOs() []*proto.GroupDTO {
...
-			if !foundType {
-				glog.Errorf("Invalid member entity type %s", etype)
-				continue
-			}
+	// member type
+	var protoType proto.EntityDTO_EntityType
+	protoType, foundType := ENTITY_TYPE_MAP[entityType]
+	if !foundType {
+		glog.Errorf("Invalid member entity type %s", entityType)
+		return []*proto.GroupDTO{}
```
ading1977: I believe there is no need to create and return an empty slice. Just return nil will work as Go allows you to append nil to a slice. There are several other places in the code that be treated the same way.

PR #351, opened 2019-11-07
Update kubernetes client lib to dynamic client

irfanurrehman:
This has been done for `replicationcontrollers` `replicasets` and `deployments`.

The code:
- Discovers the group version exposed by the given k8s server
- Uses the discovered group version of the resources with the dynamic client for all get and update operations.
- Modifies `pod utils`, pod `parent controller` and `executor` code to work with `unstructured` content rather then `typed` content. 

@enlinxu @chlam4 @ading1977

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	kubeturbo "github.com/turbonomic/kubeturbo/pkg"
+	"github.com/turbonomic/kubeturbo/pkg/util"
 	"github.com/turbonomic/kubeturbo/test/flag"
 
 	"github.com/golang/glog"
@@ -42,15 +43,10 @@ const (
 	defaultDiscoveryIntervalSec = 600
 	defaultValidationWorkers    = 10
 	defaultValidationTimeout    = 60
-	k8sExtensionsGroupName      = "extensions"
-	k8sAppsGroupName            = "apps"
 )
 
 var (
 	defaultSccSupport = []string{"restricted"}
-
-	// The API group version under which deployments and replicasets are exposed by the given k8s cluster
-	k8sAPIDeploymentReplicasetGV = &schema.GroupVersion{k8sAppsGroupName, "v1"}
 
 	// these variables will be deprecated. Keep it here for backward compatibility only
 	k8sVersion        = "1.8"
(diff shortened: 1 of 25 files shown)


cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -344,3 +367,64 @@ func checkServerVersion(restClient restclient.Interface) bool {
...
+			latestAppsVersion.Version = latestComparedVersion(version, latestAppsVersion.Version)
+		}
+	}
+
+	if latestAppsVersion.Version != "" {
+		return latestAppsVersion, nil
+	}
+	if latestExtensionsVersion.Version != "" {
+		return latestExtensionsVersion, nil
+	}
```
ading1977: Does this logic mean that we always prefer appsVersion over extensionsVersion, or it is not possible for a resource to belong to both appsGroup and extensionsGroup?

irfanurrehman: So the evolution of API resources from `extensions` to respective other groups has not been very elegant. There were many resources which were all collectively listed under group `extensions` and moved into their respective groups in the later releases, `deployments` moved to `apps` and `ingress` moved to `networking.k8s.io` being few examples.

It is possible for a resource to exist under multiple groups.
The strategy used here (which is kind of hard coding) is to simply prefer `apps` over `extensions`.

In case we move our code such that all other resource versions need to be discovered (I don't see the need for that as yet); a simple strategy would be to always not prefer choosing `extentions` if a resource is found in multiple groups.

irfanurrehman:
> * For all the `fmt.Errorf()`, suggest starting the error string with lower case, as the error string is a part of the message in the parent function.

Updated

> * Now that we are using `dynClient`, do we still need the static client? If not, can we remove the unused static client from the code?

I think there is no typed client reference which might be unused. I did update the usage of replicaset and deployment clients to dynamic. If there is an unused reference of any package`go mod tidy` should clean it.

> * Could you please update what tests have been done? For minimum, I think we need to test consistent resize which will update the deployment. It will be ideal if we can also test provision/suspension of pods in deployment.

I did test that scaling actions result in update of the parent spec (scale a pod updates the deployment spec). I did not test provision/suspension of pods. I can test that too once I am back to a stable working environment.. :-)

cmd/kubeturbo/app/kubeturbo_builder.go
ading1977: Thanks for the info. That makes sense to me.

irfanurrehman:
@ading1977 @enlinxu I did verify this today for both resize action and horizontal scale (suspend) action.
Both work as expected. 
Fixed the review comments. 
Fixed a non functional minor issue discovered during test.
I think this can be merged now.

pkg/discovery/util/pod_util.go
```diff
@@ -267,7 +270,7 @@ func GetPodParentInfo(pod *api.Pod) (string, string, error) {
 // GetPodGrandInfo gets grandParent (parent's parent) information of a pod: kind, name
 // If parent does not have parent, then return parent info.
 // Note: if parent kind is "ReplicaSet", then its parent's parent can be a "Deployment"
-func GetPodGrandInfo(kclient *client.Clientset, pod *api.Pod) (string, string, error) {
+func GetPodGrandInfo(kclient *client.Clientset, dynClient dynamic.Interface, pod *api.Pod) (string, string, error) {
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

ading1977:
Hi, @irfanurrehman 

Yes, please go ahead and merge.

PR #352, opened 2019-11-08
OM-50891 Handling of pods on unschedulable nodes

pallavidn:
Nodes in Kubernetes where new pods can be deployed are marked as 'schedulable'. This property is used to by Turbo to consider the node as a provider for moving pods.

We use Access commodity with a special key 'schedulable' to represent nodes that can serve as providers for pods.
- Nodes selling this commodity will accept pods.
- Pods buying this commodity can move to these nodes

Often a node is marked unschedulable for maintenance purposes or to prevent pods from being deployed to it. In this case, Turbo will not create the access commodity and the node will stop being a providers for pods, so there will be not be any recommendations to move pods to this node.

There is an issue on how to handle the pods that are already deployed on the unschedulable nodes. These pods are there for reasons such as 
- Pod is configured with toleration for unschedulable nodes
- Pod was deployed before the node was made unschedulable
- Specific node selectors led the pod to be deployed to this node

For these pods, it is desirable that the pods continue to stay on the node unless there is a resource(compute) constraint or policy violation. We don't want Turbo to give recommendations to move these pods.
In this change, we create the schedulable access commodity for pod when the underlying node is schedulable. When the node is unschedulable, the commodity is not created(or deleted) to prevent it from being moved out of the node.

### DIFF
--- a/pkg/discovery/dtofactory/node_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/node_entity_dto_builder.go
@@ -169,7 +169,7 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node) ([]*
 	// Access commodities: labels.
 	for key, value := range node.ObjectMeta.Labels {
 		label := key + "=" + value
-		glog.V(4).Infof("label for this Node is : %s", label)
+		//glog.V(2).Infof("label for this Node is : %s", label)
 
 		accessComm, err := sdkbuilder.NewCommodityDTOBuilder(proto.CommodityDTO_VMPM_ACCESS).
 			Key(label).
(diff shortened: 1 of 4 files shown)


pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -169,7 +169,7 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node) ([]*
 	// Access commodities: labels.
 	for key, value := range node.ObjectMeta.Labels {
 		label := key + "=" + value
-		glog.V(4).Infof("label for this Node is : %s", label)
+		//glog.V(2).Infof("label for this Node is : %s", label)
```
ading1977: Shall we bring back the V(4) logging here?

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -211,6 +211,7 @@ func (builder *podEntityDTOBuilder) getPodCommoditiesBought(pod *api.Pod, cpuFre
 	// Access commodities: selectors.
 	for key, value := range pod.Spec.NodeSelector {
 		selector := key + "=" + value
+		//glog.V(2).Infof("%s: node selector is %s", pod.Name, selector)
```
ading1977: Please remove commented out code

PR #362, opened 2019-12-19
OM:52890 Create anti-affinity policy for pods running on unschedulable nodes

pallavidn:
In Turbo, Schedulable nodes are marked by requiring to sell an access commodity 'schedulable'. Pods are required to buy this commodity so they can be moved to schedulable nodes
Unchedulable nodes do not sell this commodity, hence pods will not be moved to these nodes.
We do want the pods that are already running on the unschedulable nodes to not be moved out due to compliance issues, when the node stops selling the commodity. This is done by not requiring the pods already on the unschedulable nodes at the time of discovery to  buy the 'schedulable' commodity, so they can continue staying on the current node unless there are resource constraints or unless moved by the kubernetes controller.
This, however, created a situation where some pods on one unschedulable nodes are recommended to be moved to another unschedulable nodes. 
This PR resolves this by creating an anti-affinity policy for the pods on one unschedulable nodes with respect to other unschedulable nodes.


pkg/discovery/worker/compliance/schedulable_node_manager.go
```diff
@@ -1,43 +1,76 @@
...
+	// List of unschedulable nodes by node name
+	unSchedulableNodes []string
 }
 
-func NewSchedulableNodeManager(nodes []*api.Node) *SchedulableNodeManager {
-	nodeMap := make(map[string]*api.Node)
-	for _, node := range nodes {
-		nodeMap[node.Name] = node
+// Build a map for identifying the UIDs of the pods on the unschedulable nodes
+func (manager *SchedulableNodeManager) buildNodeToPodMap() map[string][]string {
```
irfanurrehman: Recommend naming this to `buildNodeToPodUIDMap` to avoid confusion. We already have `manager.cluster.NodeNameToPodMap` in use below.

pkg/discovery/worker/compliance/schedulable_node_manager.go
```diff
@@ -1,43 +1,76 @@
...
+	nodeToPodsMap := make(map[string][]string)
```
irfanurrehman: As above `nodeToPodUIDsMap`

pkg/discovery/worker/compliance/schedulable_node_manager.go
```diff
@@ -1,43 +1,76 @@
...
 	}
+
+	manager.checkNodes()
+	return manager
 }
 
 func (manager *SchedulableNodeManager) checkNodes() {
-	for _, node := range manager.nodes {
+	nodeList := manager.cluster.NodeList
+	for _, node := range nodeList {
```
irfanurrehman: Variable `nodeList` not needed additionally.

pkg/discovery/worker/compliance/schedulable_node_manager.go
```diff
@@ -1,43 +1,76 @@
...
+		// check for schedulable property
 		schedulable := util.NodeIsReady(node) && util.NodeIsSchedulable(node)
-		manager.schedulableNodes[node.Name] = schedulable
+		manager.schedulableStateMap[node.Name] = schedulable
+
+		if !schedulable {
+			manager.unSchedulableNodes = append(manager.unSchedulableNodes, node.Name)
```
irfanurrehman: I would clear `manager.unSchedulableNodes` at the beginning in this function to make multiple calls to `checknodes` possible, and so as to not add the same nodes multiple times in this set.

pkg/discovery/worker/compliance/taint_toleration_processor.go
```diff
@@ -30,41 +31,36 @@ type TaintTolerationProcessor struct {
 	// Map of node uid indexed by node name
 	nodeNameToUID map[string]string
 
+	cluster *repository.ClusterSummary
+
 	// Manager for schedulable nodes
-	nodesManager *SchedulableNodeManager
+	unSchedulableNodesManager *SchedulableNodeManager
```
irfanurrehman: This naming is totally off, (don't you think.. :)) with confusing readability.
Probably changing the name of the original structure to `nodesSchedulabilityManager` might solve this conflict as its dealing with both schedulable and unschedulable nodes. And then use the name `nodeManager` for the variable here.

pkg/discovery/worker/compliance/taint_toleration_processor.go
```diff
@@ -30,41 +31,36 @@ type TaintTolerationProcessor struct {
...
 }
 
-func NewTaintTolerationProcessor(nodeAndPodGetter NodeAndPodGetter) (*TaintTolerationProcessor, error) {
-	nodes, err := nodeAndPodGetter.GetAllNodes()
```
irfanurrehman: I have been looking at the code and at many places I find the use of interfaces and the original structure intertwined.
I think changes like this have made the code so.
Should we keep a guideline to not go back from the abstractions already defined in the code earlier? And if we choose to do so, cleanup the relevant code properly. [I guess, I would say no action needed here for now]

pkg/discovery/worker/compliance/taint_toleration_processor.go
```diff
@@ -30,41 +31,36 @@ type TaintTolerationProcessor struct {
...
-	if err != nil {
-		return nil, err
-	}
-	pods, err := nodeAndPodGetter.GetAllPods()
-	if err != nil {
-		return nil, err
-	}
+func NewTaintTolerationProcessor(cluster *repository.ClusterSummary,
+	unSchedulableNodesManager *SchedulableNodeManager) (*TaintTolerationProcessor, error) {
```
irfanurrehman: Same as previous to previous comment, and for all such references.

pkg/discovery/worker/compliance/taint_toleration_processor.go
```diff
@@ -125,7 +121,7 @@ func (t *TaintTolerationProcessor) createAccessCommoditiesBought(podDTOs []*prot
 		}
 
 		// Schedulable
-		schedulableComm, _ := createSchedulableBoughtComms(pod, t.nodesManager)
+		schedulableComm, _ := createSchedulableBoughtComms(pod, t.unSchedulableNodesManager)
```
irfanurrehman: I would log the error rathen then ignoring it.

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// In Turbo, Schedulable nodes are marked with an access commodity 'schedulable'. Pods are required to buy
+// this commodity so they can be moved to schedulable nodes
+// Unchedulable nodes do not sell this commodity, hence pods will not be moved to these nodes.
+// We want to address the issue that pods that are already running on the unschedulable nodes are not moved out
+// due to compliance issues. This is done by not requiring the pods already on the unschedulable nodes at the tine
```
irfanurrehman: nit: s/tine/time

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,102 @@
...
+// of discovery to not buy the 'schedulable' commodity so they can continue staying on the current node unless there
+// are resource constraints or unless moved by the kubernetes controller.
+// In addition, we need pods which do not buy the schedulable commodity to not move to other unschedulable nodes.
+// This is achieved by creating an anti-affinity policy for these pods to the other unschedulable nodes other than the
```
irfanurrehman: Nit: Probably replace to: `This is achieved by creating an anti-affinity policy for pods running on this unschedulable node against all other unschedulable nodes.`

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,102 @@
...
+// Creates DTOs for anti-affinity policy between the pods running on a unschedulable nodes to other unschedulable nodes.
+// This will prevent the pods running on a unschedulable nodes to be moved to other other unschedulable nodes.
+func (builder *UnschedulableNodesAntiAffinityGroupDTOBuilder) Build() []*proto.GroupDTO {
+	var groupDTOs []*proto.GroupDTO
+
+	// Map of unschedulable node to other unschedulable nodes
+	unSchedulableNodesGroups := builder.getUnschedulableNodesGroups()
+
+	// Map of unschedulable node name to list of UIDs of pods running on these unschedulable nodes
+	nodeToPodsMap := builder.nodesManager.buildNodeToPodMap()
```
irfanurrehman: This will be updated to `nodeToPodUIDsMap` as in an earlier comment.

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,102 @@
...
+
+	for unschedulableNodeName, nodeMembers := range unSchedulableNodesGroups {
```
irfanurrehman: Suggestion: A better way of doing this can be:

Rename `unSchedulableNodes` to `unSchedulableNodesByUID` [here]( https://github.com/turbonomic/kubeturbo/blob/a863d96886732eca173a6e249fa13675cdbdc080/pkg/discovery/worker/compliance/schedulable_node_manager.go#L17:2)
Change [this line](https://github.com/turbonomic/kubeturbo/blob/a863d96886732eca173a6e249fa13675cdbdc080/pkg/discovery/worker/compliance/schedulable_node_manager.go#L59) to below:
`manager.unSchedulableNodesByUID = append(manager.unSchedulableNodesByUID, node.UID)`
Basically curate a list of node IDs in place of node names.

Then use below code.
```
       var nodeUIDtoNameMap make(map[string]string)
       var nodeUIDs []string
	for _, nodeName := range builder.nodesManager.unSchedulableNodes {
		nodeUID := builder.cluster.NodeNameUIDMap[nodeName]
               nodeUIDtoNameMap[nodeUID] = nodeName
		nodeUIDs = append(nodeUIDs, nodeUID)
	}

	for indx, nodeUID := range nodeUIDs {
    		var otherNodeUIDs []string
    		otherNodeUIDs := append(otherNodeUIDs, unSchedulableNodesByUID[:indx]...)
    		otherNodeUIDs := append(otherNodeUIDs, nodeUIDs[indx+1:]...)
    		nodeName := nodeUIDtoNameMap[nodeUID]

	// rest of your code below


	}
```

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		glog.V(3).Infof("Unschedulable nodes anti affinity group: %+v", groupDTO)
+		groupDTOs = append(groupDTOs, groupDTO...)
+	}
+
+	return groupDTOs
+}
+
+// Helper method to map a unschedulable node to the UIDs of other unschedulable nodes
+func (builder *UnschedulableNodesAntiAffinityGroupDTOBuilder) getUnschedulableNodesGroups() map[string][]string {
```
irfanurrehman: This wont be needed if you use the suggestion above.

pkg/discovery/worker/compliance/taint_toleration_processor.go
pallavidn: Replaced the NodeAndPodGetter, that makes redundant api calls to get nodes and pods for this parsing the taints and tolerations, with the cluster summary. Cluster summary is created after node verification and holds the discovered nodes and pods objects.

irfanurrehman: Ack!

pkg/discovery/worker/dispatcher.go
```diff
@@ -81,7 +81,15 @@ func (d *Dispatcher) Dispatch(nodes []*api.Node, cluster *repository.ClusterSumm
 	for assignedNodesCount+perTaskNodeLength <= len(nodes) {
 		currNodes := nodes[assignedNodesCount : assignedNodesCount+perTaskNodeLength]
 
-		currPods := d.config.clusterInfoScraper.GetRunningAndReadyPodsOnNodes(currNodes)
+		var currPods []*api.Pod
+		//currPods := d.config.clusterInfoScraper.GetRunningAndReadyPodsOnNodes(currNodes)
```
irfanurrehman: Nit: Any reason of retaining this ? If not please remove (here an at all places below).

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+// Creates DTOs for anti-affinity policy between the pods running on a unschedulable nodes to other unschedulable nodes.
+// This will prevent the pods running on a unschedulable nodes to be moved to other other unschedulable nodes.
+func (builder *UnschedulableNodesAntiAffinityGroupDTOBuilder) Build() []*proto.GroupDTO {
+	var policyDTOs []*proto.GroupDTO
+
+	if len(builder.nodesManager.unSchedulableNodes) <= 1 {
+		glog.Info("Zero or one unschedulable node in the cluster, anti-affinity policies will not be created")
```
irfanurrehman: Nit: Recommend to increase verbosity level. In a lot of cases there would be just the master (in single master clusters) that will be marked unschedulable.

irfanurrehman:
@pallavidn thanks for the updates. Few more nits pending which LGTM.

irfanurrehman:
lgtm; Thanks @pallavidn.

pkg/discovery/worker/compliance/unschedulable_node_anti_affinity_group_dto_builder.go
```diff
@@ -0,0 +1,100 @@
...
+
+		// node members comprising of other unschedulable nodes for the policy
+		var otherNodeUIDs []string
+		otherNodeUIDs = append(otherNodeUIDs, nodeUIDs[:idx]...)
+		otherNodeUIDs = append(otherNodeUIDs, nodeUIDs[idx+1:]...)
+
+		nodeMembers := otherNodeUIDs
+
+		// policy id and display name
+		groupID := fmt.Sprintf("UnschedulableNodesAntiAffinity-%s-%s", nodeName, builder.targetId)
```
enlinxu: We may want to check if nodeMembers (podMembers as well) is empty and skip the rest of the code if it is.

pallavidn: Beginning of the method checks if there is only one unshcedulable node so policy creation can be skipped. Check for podMembers also exists.

PR #364, opened 2019-12-30
OM-52711: Fix for pod to pod affinity rule being ignored runtime

irfanurrehman:
In kubernetes the available scheduling policy for a pod to pod affinity
rules are requiredDuringSchedulingIgnoredDuringExecution or
preferredDuringSchedulingIgnoredDuringExecution which means that the
rescheduling does not happen if the rule is dynamically violated, meaning
the original set of pod(s), a given pod has affinity rule specified for,
moves to a different node. With this fix, we in kubeturbo will try to
ensure compliance for this affinity rule run time also.

### DIFF
--- a/pkg/discovery/worker/compliance/affinity_processor.go
+++ b/pkg/discovery/worker/compliance/affinity_processor.go
@@ -83,12 +83,13 @@ func (am *AffinityProcessor) processAffinityPerPod(pod *api.Pod, podsNodesMap ma
 		return
 	}
 
+	hostNode := podsNodesMap[pod]
 	for _, node := range am.nodes {
 		if matchesNodeSelector(pod, node) && matchesNodeAffinity(pod, node) {
-			am.addAffinityAccessCommodities(pod, node, nodeAffinityAccessCommoditiesSold, nodeAffinityAccessCommoditiesBought)
+			am.addAffinityAccessCommodities(pod, node, hostNode, nodeAffinityAccessCommoditiesSold, nodeAffinityAccessCommoditiesBought)
 		}
 		if interPodAffinityMatches(pod, node, podsNodesMap) {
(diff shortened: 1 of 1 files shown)


enlinxu:
@irfanurrehman  would you please add the test result?

irfanurrehman:
> @irfanurrehman would you please add the test result?

@enlinxu thanks for having a look. The test results are updated against the [jira issue](https://vmturbo.atlassian.net/browse/OM-52711).

pkg/discovery/worker/compliance/affinity_processor.go
```diff
@@ -83,14 +83,28 @@ func (am *AffinityProcessor) processAffinityPerPod(pod *api.Pod, podsNodesMap ma
...
-			am.addAffinityAccessCommodities(pod, node, podAffinityCommodityDTOsSold, podAffinityCommodityDTOsBought)
+			am.addAffinityAccessCommodities(pod, node, hostNode, podAffinityCommodityDTOsSold, podAffinityCommodityDTOsBought)
+		}
+	}
+}
+
+// findPodsNode is needed because we do not get the maps key as a single unique value consistently.
+// The key which is a pointer when passed around functions can result in a new pointer variable
+// pointing to the same value (*pod.Pod).
+func findPodsNode(pod *api.Pod, podsNodesMap map[*api.Pod]*api.Node) *api.Node {
```
ading1977: @irfanurrehman I am not sure I fully understand. Where in our code that you see this pointer key might get changed? 

https://stackoverflow.com/questions/34585018/is-it-safe-to-use-pointers-as-map-keys-in-go

irfanurrehman:
@ading1977 as talked, updated the code to add an example in comment and verified that hacky function is not needed. Now using the map directly. It works as expected.

PR #365, opened 2020-01-13
Add support for DeploymentConfig

irfanurrehman:
environments.


irfanurrehman:
This PR does not implement discovering the group and version of `deploymentconfig` as of now. That is the next step if we see a supported version different from `v1` in some cluster.

esara:
@irfanurrehman I thought deploymentconfig is a bit of legacy construct created in openshift before deployments were available, how far do we need to go supporting this?

irfanurrehman:
@esara FYI I have verified this to be working as expected against one of our openshift clusters.

PR #366, opened 2020-01-17
Group creation fixes

irfanurrehman:
We create groups (of containers and pods) by parent workload type while discovering the cluster and report the same to opsmgr.
These groups are available to set useful policies.
The reported groups so far were one per parent kind per parent and one single group per parent type (for example `pods by all deployments` across the cluster) was missing.
Interestingly and incidentally, some code to do that already did exist but had flaws.
This PR fixes that.
I am sure this is useful to any of our customers who actually uses the containers featureset.

### DIFF
--- a/pkg/discovery/dtofactory/group_dto_builder.go
+++ b/pkg/discovery/dtofactory/group_dto_builder.go
@@ -73,8 +73,8 @@ func (builder *groupDTOBuilder) BuildGroupDTOs() ([]*proto.GroupDTO, error) {
 		}
 
 		// create groups for each configured entity type
-		for etype, resizeEntityType := range ENTITY_RESIZE_FLAG_MAP {
-			groupDTOs := builder.createGroupsByEntityType(entityGroup, etype, resizeEntityType)
+		for etype, isEntityResizable := range ENTITY_RESIZE_FLAG_MAP {
+			groupDTOs := builder.createGroupsByEntityType(entityGroup, etype, isEntityResizable)
 
 			result = append(result, groupDTOs...)
 		}
@@ -84,7 +84,7 @@ func (builder *groupDTOBuilder) BuildGroupDTOs() ([]*proto.GroupDTO, error) {
 }
 
 func (builder *groupDTOBuilder) createGroupsByEntityType(entityGroup *repository.EntityGroup,
-	entityType metrics.DiscoveredEntityType, resizeEntityType bool) []*proto.GroupDTO {
+	entityType metrics.DiscoveredEntityType, isEntityResizable bool) []*proto.GroupDTO {
 
 	var result []*proto.GroupDTO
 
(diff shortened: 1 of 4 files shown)


irfanurrehman:
Usage example:- we generate resize actions on daemonset pods, but execution of such an action always fails because consistent resizing is on by default and we do not update the parent daemonset.
Having a group which filters all pods and all containers by all daemonsets across the cluster can help you create a policy to avoid generating these actions in the first place.
cc @esara

ading1977:
Hi, @irfanurrehman, can you post some screenshots of your test result? We want to make sure #326 is not broken, basically, we should be able to see, for example:

* A single group of containers from all deployments in a cluster
* One group of containers per deployment in a cluster
* One group of containers with the same name per deployment in a cluster

irfanurrehman:
@ading1977 verifications for consistent resize flag set on groups led into many finding and fixup of them all.

pkg/discovery/dtofactory/group_dto_builder.go
```diff
@@ -104,23 +104,29 @@ func (builder *groupDTOBuilder) createGroupsByEntityType(entityGroup *repository
...
+	}
 
 	// resize setting for entities is based on the parent type of the group and if sub groups will be created.
 	// If sub groups are to be created, the resize setting is applied on the sub-groups
 	// Else, the parent level container group has the consistent resize flag set to true
-	createSubGroups := hasSubGroups(entityGroup, entityType)
-
-	resizeFlag := resizeEntityType && consistentResizeGroupType(entityGroup) && !createSubGroups
+	needsSubGroups := needsSubGroups(entityGroup, entityType)
+	isSingleEntityInGroup := len(memberList) <= 1
```
ading1977: Why is this logic needed?

irfanurrehman: > Why is the `isSingleEntityInGroup` logic needed?

@ading1977 we get groups by parent (for example Pod by deployment, and containers by deployment) which have only one member (a Deployment which currently has only 1 replica which has only one container). The group containers by deployment, for this deployment, which has only one member with earlier code will have consistentResize=true. This fixes that. The other approach could be to not create a group at all if it has a single entity in it. We can choose either. I left creating the group as is to be consistent with creating a group for each parent.

irfanurrehman:
> Is this the summary of the primary fixes?
> 
> * add the entity group by kind from each group_metrics_collector to the list of groups that is returned by each worker
> * re-compute the consistentResize flag so it is not applied to the groups by kind

@pallavidn there is no new functionality that is added in this PR that we did not expect before. The existing grouping functionality is precisely summarised by @ading1977 in [this comment](https://github.com/turbonomic/kubeturbo/pull/366#issuecomment-575666075). The code meant to achieve creation of these groups sort of has been in place. Lets say some of it wasn't functioning as expected and had subtle bugs at some places. I have tried to create commits which address updates wrt fixing those subtlelities. Additionally I have updated some code to make it more maintainable, especially in [this commit](https://github.com/turbonomic/kubeturbo/pull/366/commits/155f6918190b242522ad63d6981f1adbb45ad11c).

pkg/discovery/dtofactory/group_dto_builder.go
ading1977: Just want to confirm, what is the original behaviour when there is only one pod/container in the deployment? If the original behaviour is to set consistentResize=true for deployment with one pod/container, I believe we need to keep that behaviour.

pallavidn: It would be less confusing to create groups per parent kind/parent instance even if they contain a single member. 
Meng is right,  since the consistent resize flag is read from the group level, the same rule would apply for group with single members as well for computing the consistentResize flag

irfanurrehman: Done!

irfanurrehman:
@ading1977 @pallavidn Updated! PTAL

PR #367, opened 2020-01-22
Fix for CN-4

irfanurrehman:
This PR is based on top of #366.
Please ensure that #366 is merged before this.
@ading1977 @evat-pm

### DIFF
--- a/pkg/discovery/worker/group_metrics_collector.go
+++ b/pkg/discovery/worker/group_metrics_collector.go
@@ -56,35 +56,43 @@ func (collector *GroupMetricsCollector) CollectGroupMetrics() ([]*repository.Ent
 				entityGroupList = append(entityGroupList, entityGroup)
 			}
 
+			// group1 = A group for each parent of this kind/type
 			entityGroup = entityGroups[ownerTypeString][ownerString]
-			// Add pod member to the group
+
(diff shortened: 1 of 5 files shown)


pkg/discovery/worker/group_metrics_collector.go
```diff
@@ -115,3 +114,11 @@ func (collector *GroupMetricsCollector) getGroupName(etype metrics.DiscoveredEnt
 
 	return ownerTypeString, ownerString, nil
 }
+
+func groupKey(kind, ns, name string) string {
```
ading1977: I could be wrong, but is this function being used anywhere?

pallavidn: same question

irfanurrehman: Apologies, I did intend to remove it but I published this PR probably half asleep late last night.. :)
Updated!

irfanurrehman:
> Can you confirm if the main change here to include the namespace in the group name ? The rest of the code is the restructuring and the fix that was already done in the PR#366.

@pallavidn yes, as listed in the title comment of this PR, this is on top of #366 so only the last commit in this PR is reviewable here.

PR #370, opened 2020-02-03
OM-51440 Use entity action settings changes for master nodes and daemon pods.

pallavidn:
OM-51440 provides a framework for action eligibility settings for entities. Entities can individually disable move, suspend or provision actions, which could be different from other entities of the same type. Kubeturbo has been updated with the latest turbo-go-sdk containing the entity action settings.
Kubeturbo has many use cases for using this framework.
- Disable suspend for master nodes
- Disable vertical resize of nodes
- Disable move for Daemon pods
- Disable suspend and provision for Daemon pods.

This PR makes use of the action eligibility settings for master nodes and daemon pods.

Testing was done using a locally built XL server with the changes outlined in two server side reviews found on RBCommons.
Targets used were dc11 Openshift cluster stitched with vCenter.


pkg/discovery/util/node_util.go
```diff
@@ -77,3 +79,43 @@ func NodeIsControllable(node *api.Node) bool {
...
+	// return all the role associated with the node
+	return allRoles
+}
+
+func DetectMasterRole(node *api.Node) bool {
+	nodeRoles := DetectNodeRoles(node)
+
+	isMasterNode := nodeRoles.Has(masterRoleName)
+
+	fmt.Printf("%s - %v [allRoles - %v] [masterRoles - %v]\n", node.Name, node.Labels, nodeRoles, isMasterNode)
```
chlam4: This is supposed to be a log statement instead of fmt.Printf?

pkg/registration/supply_chain_factory.go
```diff
@@ -136,11 +137,14 @@ func (f *SupplyChainFactory) buildNodeMergedEntityMetadata() (*proto.MergedEntit
 		builder.PropertyCapacity: {},
 	}
 	fieldsUsedCapacityPeak := map[string][]string{
```
chlam4: Nit: should we have a better name here, now that we're adding one more property resizable into the used, capacity and peak list? :)

pkg/registration/supply_chain_factory.go
```diff
@@ -136,11 +137,14 @@ func (f *SupplyChainFactory) buildNodeMergedEntityMetadata() (*proto.MergedEntit
...
-		builder.PropertyCapacity: {},
-		builder.PropertyPeak:     {},
+		builder.PropertyUsed:      {},
+		builder.PropertyCapacity:  {},
+		builder.PropertyPeak:      {},
+		builder.PropertyResizable: {},
 	}
 	mergedEntityMetadataBuilder := builder.NewMergedEntityMetadataBuilder()
+
+	mergedEntityMetadataBuilder.PatchField(ActionEligibilityField, []string{})
```
chlam4: Curious: why are we patching it as an empty list?

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -39,6 +35,13 @@ var (
 		metrics.CPURequestQuota,
 		metrics.MemoryRequestQuota,
 	}
+
+	resizableCommodities = map[proto.CommodityDTO_CommodityType]bool{
```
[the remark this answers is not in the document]
pallavidn: added comment

pkg/registration/supply_chain_factory.go
pallavidn: We are patching the action eligibility section in the DTO. The 2nd argument is the path to the section, it is empty  because the section is at the top level in the entity dto.

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -68,6 +68,9 @@ func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]*proto.E
 		podID := string(pod.UID)
 		entityDTOBuilder := sdkbuilder.NewEntityDTOBuilder(proto.EntityDTO_CONTAINER_POD, podID)
 
+		// determine if the pod is a daemon set pod as that determines the eligibility of the pod for different actions
+		daemon := util.Daemon(pod)
+
```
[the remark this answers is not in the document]
pallavidn: It is used for setting the eligibility of the movable and the suspendable actions. Movable action setting is per provider and suspendable is at the entity level.
I left this method which is used to determine if a pod is daemon set pod before we set the commodity bought and action eligibility sections, putting it closer to 104 will out it inside of the commodity bought setter code.

pkg/discovery/dtofactory/node_entity_dto_builder.go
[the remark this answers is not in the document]
pallavidn: added comment

PR #372, opened 2020-02-06
parse container id as entity metadata

esara:
parse container id as entity metadata, the resulting DTO looks like this:
```
entityDTO {
  entityType: CONTAINER
  id: "196e532e-624d-42b9-9018-c1ed4289bc36-0"
  displayName: "kube-system/coredns-6955765f44-qvnvx/coredns"
...
  entityProperties {
    namespace: "DEFAULT"
    name: "UUID"
    value: "d3910dbd1c204c6357ff368d76957105701d1c80fc200277b22580e8614fd13b"
  }
```


pkg/discovery/dtofactory/container_dto_builder.go
```diff
@@ -220,3 +241,14 @@ func (builder *containerDTOBuilder) addPodProperties(pod *api.Pod, index int) []
...
 	return properties
 }
+
+// Get the stitching property for Application.
+func getContainerStitchingProperty(pod *api.Pod, index int) string {
+	// Parse the container id from the status, it is either in the form of
+	// containerID: docker://986c8fc7247d1ff047f06e8e3487029b89baef1e908d3e334dd94aaa3bf4bc39
+	// or
+	// containerID: cri-o://81ab6978aa3252bc53cfaf7d23b8b1bf5f8f27ff4b9e79a0fbaffebeddaeeb4b
+	containerStatus := &(pod.Status.ContainerStatuses[index])
```
chlam4: Perhaps we can add a check here to ensure `ContainerStatuses` exists and it doesn't blow up, e.g.
```
if len(pod.Status.ContainerStatuses) == 0 {
  return ""
}
```
Or maybe even better this way?
```
if len(pod.Status.ContainerStatuses) <= index {
  return ""
}
```

pkg/discovery/dtofactory/container_dto_builder.go
```diff
@@ -220,3 +241,14 @@ func (builder *containerDTOBuilder) addPodProperties(pod *api.Pod, index int) []
...
+	cntiduri := strings.Split(containerStatus.ContainerID, "://")
+	return cntiduri[1]
```
chlam4: Along the same line to avoid any possibility of index out of range panic, we could add a check here too e.g.
```
if len(cntiduri) <= 1 {
  return ""
}
```

pkg/discovery/dtofactory/container_dto_builder.go
```diff
@@ -212,6 +214,25 @@ func (builder *containerDTOBuilder) getCommoditiesBought(podId, containerName, c
 	return result, nil
 }
 
+func (builder *containerDTOBuilder) getContainerProperties(pod *api.Pod, index int) []*proto.EntityDTO_EntityProperty {
+	var properties []*proto.EntityDTO_EntityProperty
+	podProperties := builder.addPodProperties(pod, index)
+
+	ns := stitching.DefaultPropertyNamespace
+	attr := string(stitching.UUID)
```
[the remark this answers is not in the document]
ading1977: Agreed. We should use a different attribute for the purpose of container stitching.

pkg/discovery/dtofactory/container_dto_builder.go
esara: thanks

pkg/discovery/dtofactory/container_dto_builder.go
esara: done, thanks

pkg/discovery/dtofactory/container_dto_builder.go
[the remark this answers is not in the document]
esara: moved to two new attributes (CNT for container id and CNTFULL for container full id)

pkg/discovery/dtofactory/container_dto_builder.go
```diff
@@ -220,15 +220,27 @@ func (builder *containerDTOBuilder) getContainerProperties(pod *api.Pod, index i
 	properties = append(properties, podProperties...)
 
 	ns := stitching.DefaultPropertyNamespace
-	attr := string(stitching.UUID)
+	idattr := string(stitching.CNTID)
+	fullidattr := string(stitching.CNTFULLID)
 	value := getContainerStitchingProperty(pod, index)
+	// Short containerID
+	if len(value) > 11 {
```
chlam4: Nit: I'd suggest to define a constant for this "11" number so we know what it means and the following number "12" can just be a "+1".

pkg/discovery/stitching/stitching_manager.go
```diff
@@ -16,6 +16,8 @@ import (
 const (
 	UUID StitchingPropertyType = "UUID"
 	IP   StitchingPropertyType = "IP"
+	CNTID StitchingPropertyType = "CNT"
+	CNTFULLID StitchingPropertyType = "CNTFULL"
```
chlam4: This `StitchingPropertyType` is actually a config enum for the users to indicate if we use UUID or IP to stitch VMs.  Now we have a lot of other entities being stitched, it's probably a good idea to change it to something like `VMStitchingPropertyType`.

In other words, the two container stitching properties newly introduced here aren't really belonging to the config enum and I'd thus suggest to just simply define them as a string like the `AppStitchingAttr` below.  For example,
```
ContainerShortIdStitchingAttr string = "CNT"
ContainerFullIdStitchingAttr string = "CNTFULL"
```

pkg/discovery/dtofactory/container_dto_builder.go
esara: I created a constant, thansk

pkg/discovery/stitching/stitching_manager.go
esara: good point, I just turned them into strings

PR #376, opened 2020-02-21
kubeturbo operator update 7.21

esara:


deploy/kubeturbo-operator/bundle/7.21.0/kubeturbo-operator.v7.21.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,210 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[{"apiVersion":"charts.helm.k8s.io/v1alpha1","kind":"Kubeturbo","metadata":{"name":"kubeturbo-release"},"spec":{"restAPIConfig":{"opsManagerPassword":"Turbo_password","opsManagerUserName":"Turbo_username"},"serverMeta":{"turboServer":"https://Turbo_server_URL","version":7.21}}}]'
```
chlam4: Nit: probably should include turbo credentials secret here too, now that you've introduced it?

deploy/kubeturbo/values.yaml
```diff
@@ -19,9 +19,10 @@ serverMeta:
   version: TURBONOMIC_SERVER_VERSION
   turboServer: https://Turbo_server_URL
 
-# Turbo server api user and password
+# Turbo server api user and password stored in a secret or optionally specified as username and password
 # The opsManagerUserName requires Turbo administrator role
 restAPIConfig:
+  turbonomicCredentials: "turbonomic-credentials"
```
chlam4: Suggest to name the var such that users could clearly tell this is a secret name, e.g. `turbonomicCredentialsSecretName`.

To be consistent, I'd also change the following two vars to something like:
- `turbonomicUserName`
- `turbonomicPlaintextPassword`

esara: thanks @chlam4 turbonomicCredentialsSecretName makes sense, I changed it
the other two values, I cannot change, otherwise we break backward compability
but we should prefer using the secret, so eventually they can be obsoleted

deploy/kubeturbo-operator/bundle/7.21.0/kubeturbo-operator.v7.21.0.clusterserviceversion.yaml
esara: this is just an example and I am removing all the credentials, I think we should encourage users to use the secret and by default they can just create the default secretname

PR #383, opened 2020-03-09
Unit conversion improvements

irfanurrehman:
These changes are towards better maintainability. Improves a bunch of code with conflicting names and readability issues. 
Additionally updates the storage values reported by node wrt the units expected by turbonomic server.

This is on top of https://github.com/turbonomic/kubeturbo/pull/382 and only last 2 commits are reviewable here. Please ensure #382 merges before this.

PR #384, opened 2020-03-09
OM-55951: Implement gathering relevant storage metrics per pod

irfanurrehman:
Implements code for https://vmturbo.atlassian.net/browse/OM-55951.
This PR currently picks the ephemeral storage details used by the pods. Future work will improve on this and include details of persistent volumes used by pods. 
This is on top of #383 and only last 2 commits are reviewable here.
Please ensure that #383 merges before this one.

### DIFF
--- a/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
+++ b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
@@ -163,7 +163,7 @@ func (m *KubeletMonitor) parseNodeStats(nodeStats stats.NodeStats) {
 	glog.V(4).Infof("Root File System size for node %s is %.3f Megabytes", nodeName, rootfsCapacity)
 	glog.V(4).Infof("Root File System used for node %s is %.3f Megabytes", nodeName, rootfsUsed)
 	m.genUsedMetrics(metrics.NodeType, key, cpuUsageCore, memoryWorkingSetKiloBytes)
-	m.genNodeFSMetrics(metrics.NodeType, key, rootfsCapacity, rootfsUsed)
+	m.genFSMetrics(metrics.NodeType, key, rootfsCapacity, rootfsUsed)
 }
 
 // Parse pod stats for every pod and put them into sink.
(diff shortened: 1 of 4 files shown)


irfanurrehman:
Rebased on latest from master and fixed the unit tests.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -172,13 +172,24 @@ func (m *KubeletMonitor) parsePodStats(podStats []stats.PodStats) {
...
 
+		ephemeralFsCapacity, ephemeralFsUsed := float64(0), float64(0)
+		if pod.EphemeralStorage.CapacityBytes != nil {
+			ephemeralFsCapacity = util.Base2BytesToMegabytes(float64(*pod.EphemeralStorage.CapacityBytes))
+		}
+		if pod.EphemeralStorage.UsedBytes != nil {
+			// OpsMgr server expects the reported size in megabytes
+			ephemeralFsUsed = util.Base2BytesToMegabytes(float64(*pod.EphemeralStorage.UsedBytes))
+		}
+
```
ading1977: Does the `pod.EphemeralStorage.CapacityBytes` represents the sum of the configured limits of ephemeral storage of all containers in a pod (i.e., `spec.containers[].resources.limits.ephemeral-storage`)? What if there is no configured limits (is `CapacityBytes` going to be `nil` in this case)? Shall we then default the limit to the node capacity?

irfanurrehman: @ading1977 I would also have expected what you are listing the behaviour as, however, in my tests, the ephemeral storage capacity of the pod is reported as the node rootfs capacity irrespective of the limits specified on pods containers. Also indicating that it is never nil. 
Let me cross verify the same on other clusters and report back my findings.
If this is indeed the case as to what kubelet reports, I would rather update this PR to get the pod capacity as the sum of limits  specified on the container(s) if the limits exist, and fall back to what is reported as the pod capacity by the kubelet (which is indeed the node rootfs capacity) if the limits are not there on the containers.




You can please hold off merging this PR until then.

irfanurrehman: Findings after further checks.
The resource `capacities` from kubelet are always node capacities. This is true for mem and cpu also. Infact, for all resources kubeturbo reports what kubelet reports and that is node capacity, irrespective of limits specified on the containers.
This is probably because, limits are something a user request's and not really a pods capacity. A pods capacity is only bound by the compute resources provided by the host, ie. the node.

Given above, I think the PR here is good to go.
@ading1977 
cc @enlinxu

ading1977:
@irfanurrehman Could you please confirm with @enlinxu if we need to merge these storage related PRs to 7.21.2 branch?

PR #385, opened 2020-03-09
Send a regular group of VMs that belong to the cluster

ading1977:
This pull request allows `kubeturbo` to send a regular group of VMs without cluster constraint. This is in addition to the existing cluster group. This new VM group is needed in `7.21.1` server to create user policies for VMs in a k8s cluster. It can be removed when the server is upgraded to `7.21.2+`.

Tests:

Verify that `VMs By Cluster [cluster name]` is showing up in the list of VM groups when creating a user policies for Virtual Machine:

image


irfanurrehman:
Looks good to me.

PR #389, opened 2020-03-23
Persistent volumes support

irfanurrehman:
This PR is now functionally complete.
Please check the code commitwise for better understanding of changes.

The overall tasks include:

Completed:
1. Discover persistent volumes and persistent volume claims from cluster.
2. Get the volume metrics from kubelet source and generate relevant metrices for consumption in kubeturbo.
3. Create a mapping between pods and volumes via pvs. This helps establish relationship between the pods DTOs and yet to come PersistentVolume DTOs.
4. Include the storage amount bought commodities in the POD DTOs.
5. Build Volume entity DTOs, complete with sold commodities.
6. Include volume and related commodity information in the supply chain.
7. Include volume entity metadata info in the probe registration.

Pending:
- Stiching properties.

### DIFF
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/turbonomic/kubeturbo/pkg/discovery/dtofactory/property"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/stitching"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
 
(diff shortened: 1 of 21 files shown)


pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -290,6 +299,30 @@ func (builder *podEntityDTOBuilder) getPodCommoditiesBoughtFromQuota(quotaUID st
...
+func (builder *podEntityDTOBuilder) getPodCommoditiesBoughtFromVolumes(pod *api.Pod, mounts []mountedVolumes) ([]*proto.CommodityDTO, error) {
+	var commoditiesBought []*proto.CommodityDTO
+	podKey := util.PodKeyFunc(pod)
+
+	for _, mount := range mounts {
+		mountName := mount.mountName
+		volEntityID := util.PodVolumeMetricId(podKey, mountName)
+		// We use <ns>/<pod-name>/<mounted-vol-name> as the commodity key
+		// to keep the relationship between pod and volume unique.
+		commKey := fmt.Sprintf("%s/%s", podKey, mountName)
```
enlinxu: StorageAmount is usually not key-ed. What's the purpose of having this unique key here?

irfanurrehman: As in the comment above. The intention is to keep the buy sell relationship between a pod and a volume unique and so actions are not generated for pods across volumes provising the same commodity (STORAGE_AMOUNT). However now that you already mention it; there are more ways of doing this (set movable to false on entities), and more scenarios that I need ways to handle (and am sort of struggling with).

For starters the mapping between pods and volumes seems to be many to many:
- The same volume can be mounted by multiple pods, depending on the readwrite modes and the support from underlying volume provider.
- One pod can mount multiple volumes.
- If the volume is mounted by multiple pods, the usage value is not supposed to be aggregated, rather each pod mounting the same volume would report the same usage amount.

I will reach out to you, if I can't arrive at a clean way of achieving the above.

irfanurrehman:
Rebased.

irfanurrehman:
closing in favour of #440

PR #393, opened 2020-03-30
Update go.mod and vendor to use new entity types and commodity types. And create new ContainerSpec entity dtos

marvelyue:
**Intent**
The intent is to
1. Update go.mod so that kubeturbo can use the new `ContainerSpec`, `WorkloadController` and `Namespace` entity types as well as new `VCPULimitQuota`, `VMemLimitQuota`, `VCPURequestQuota` and `VMemRequestQuota` commodity types.
2. Rename the corresponding quota commodity types in kubeturbo to avoid compilation errors.
3. Create new ContainerSpec entity dtos so that we can work on the changes in turbo platform side in parallel.

**Background**:
https://vmturbo.atlassian.net/browse/OM-56706
Kubeturbo should create ContainerSpec entities during discovery to model the shared portion of individual container replicas defined by the controller that manages the pods where the containers run. For example, consider the case of a Deployment that manages two types of containers: Foo and Bar. In this example, kubeturbo should create two ContainerSpec entities (one for Foo and one for Bar).

ContainerSpec entities should be marked as **NotMonitored** so that they do not participate in market analysis since we do not want the market to generate actions for ContainerSpec entities.

The purpose of ContainerSpec entity is to store historical utilization data of a group of containers of the same type so as to address the problem of loss of historical data for single container instances which are ephemeral.

**Implementation**:
1. Update go.mod by running `go get github.com/turbonomic/turbo-go-sdk@new-container-model`
2. Update vendor with latest turbo-go-sdk changes.
3. Use new xxQuota commodity types instead of previous xxALLOCATION commodities.
4. Rename metrics from `metrics.CPUQuota` and `metrics.MemoryQuota` to `metrics.CPULimitQuota` and `metrics.MemoryLimitQuota`
5. Create new `container_spec_entity_dto_builder.go` to build containerSpec entity dtos with "Monitored" as false. Displayname is container name, and entity ID is constructed by "controllerUID/containerName".
6. Collect controller UID in `getPodOwner` in `cluster_monitor.go` and store controller UID "metric" in `EntityMetricSink` in `genOwnerMetrics`.
7. Add proto.EntityDTO_CONTAINER_SPEC entity to supply chain in `supply_chain_factory.go` and `registration_client.go`

**Unit Test**:
Added unit test `container_spec_entity_dto_builder_test.go`

**Note**:
The changes here are only to create new ContainerSpec EntityDTOs with displayname and ID so that we can work on platform changes in parallel. Further changes to set up commodities sold by ContainerSpec and creation of utilization data will be implemented in different stories.

**Testing Done**:
Tested by adding kubeturbo target with the changes. ContainerSpec entity is stored in the dumped discovery dto file like:
```json
entityDTO {
  entityType: CONTAINER_SPEC
  id: "1e4495dd-37f0-11ea-bd96-005056803564/helloworld"
  displayName: "helloworld"
  monitored: false
  actionEligibility {
  }
}
```


pkg/discovery/dtofactory/general_builder.go
```diff
@@ -21,10 +21,10 @@ var (
 		metrics.CPUProvisioned:     proto.CommodityDTO_CPU_PROVISIONED,
 		metrics.MemoryProvisioned:  proto.CommodityDTO_MEM_PROVISIONED,
 		metrics.Transaction:        proto.CommodityDTO_TRANSACTION,
-		metrics.CPUQuota:           proto.CommodityDTO_CPU_ALLOCATION,
-		metrics.MemoryQuota:        proto.CommodityDTO_MEM_ALLOCATION,
-		metrics.CPURequestQuota:    proto.CommodityDTO_CPU_REQUEST_ALLOCATION,
-		metrics.MemoryRequestQuota: proto.CommodityDTO_MEM_REQUEST_ALLOCATION,
+		metrics.CPULimitQuota:      proto.CommodityDTO_VCPU_LIMIT_QUOTA,
+		metrics.MemoryLimitQuota:   proto.CommodityDTO_VMEM_REQUEST_QUOTA,
```
marvelyue: There's a typo here. Should be `CommodityDTO_VMEM_LIMIT_QUOTA` instead of `CommodityDTO_VMEM_REQUEST_QUOTA`. Will fix this in the next commit and will also update vendor with latest turbo-go-sdk changes to fix the failed CI build.

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -0,0 +1,81 @@
...
+
+// TODO consider passing []*repository.KubeContainer as argument which stores commodities data of each container instance
+func (builder *containerSpecDTOBuilder) BuildDTOs(pods []*api.Pod) ([]*proto.EntityDTO, error) {
+	var result []*proto.EntityDTO
+	controllerUIDToContainersMap := builder.getControllerUIDToContainersMap(pods)
+	for controllerUID, containersSet := range controllerUIDToContainersMap {
+		for containerName := range containersSet {
+			containerSpecID := util.ContainerSpecIdFunc(controllerUID, containerName)
+			entityDTOBuilder := sdkbuilder.NewEntityDTOBuilder(proto.EntityDTO_CONTAINER_SPEC, containerSpecID)
+			entityDTOBuilder.DisplayName(containerName)
```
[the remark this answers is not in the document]
marvelyue: Sure. containerName here is just the container name we specify in the k8s controller. 

For example, here's the yaml data snippet of the deployment "kube-state-metrics":
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  ...
  labels:
    app: kube-state-metrics
  name: kube-state-metrics
  namespace: openshift-monitoring
spec:
  ...
  replicas: 1
  ...
    spec:
      containers:
      - args:
        ...
        name: kube-rbac-proxy-main
        resources:
          limits:
            cpu: 20m
            memory: 40Mi
          requests:
            cpu: 10m
            memory: 20Mi
      - args:
        name: kube-rbac-proxy-self
        ...
      - args:
        name: kube-state-metrics
        ...
```
Then when doing a `kubectl get pods`, the pod looks something like this (1 replica with 3 containers):
```
kube-state-metrics-c4f86b5f8-97fq5             3/3     Running   3          75d
```
In the discovered entities from kubeturbo, the displayname of each entity is:
1. **Container** (existing logic): `namespace/pod.Name/container.Name` -- "openshift-monitoring/kube-state-metrics-c4f86b5f8-97fq5/kube-rbac-proxy-self"
2. **Pod** (existing logic): `namespace/pod.Name` -- "openshift-monitoring/kube-state-metrics-c4f86b5f8-97fq5"
3. **ContainerSpec** (changes in this PR). There will be 3 ContainerSpec entities created from this controller (deployment). The displayname of each ContainerSpec is just the specified `container.Name` like:
"kube-rbac-proxy-main"
"kube-rbac-proxy-self"
"kube-state-metrics"
(corresponding container.Name is same across different pod replicas)
4. **WorkloadController** (changes to be made): `controller.Name` -- "kube-state-metrics"

pkg/discovery/util/key_func.go
```diff
@@ -110,3 +110,7 @@ func NodeKeyFromPodFunc(pod *api.Pod) string {
 func VDCIdFunc(namespaceId string) string {
 	return fmt.Sprintf("%s-%s", vdcPrefix, namespaceId)
 }
+
+func ContainerSpecIdFunc(controllerUID string, containerName string) string {
+	return controllerUID + "/" + containerName
```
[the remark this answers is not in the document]
marvelyue: Will add a comment to the method here. containerName here is the name of the container "spec" we configure in the controller and it's actually also the name of a certain container instance. Name is the same for a certain type of container in different pod replicas. (example in the previous comment).

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -0,0 +1,81 @@
...
+func (builder *containerSpecDTOBuilder) getControllerUIDToContainersMap(pods []*api.Pod) map[string]map[string]struct{} {
+	controllerUIDToContainersMap := make(map[string]map[string]struct{})
+	for _, pod := range pods {
+		if pod.OwnerReferences == nil {
+			// If pod has no OwnerReferences, it is a bare pod without controller. Skip this.
+			continue
+		}
+		controllerUID, err := builder.getControllerUID(metrics.PodType, util.PodKeyFunc(pod))
+		if err != nil {
+			continue
```
marvelyue, line 55: Add a error log here

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -0,0 +1,81 @@
...
+		controllerUIDToContainersMap[controllerUID] = containerSet
+	}
+	return controllerUIDToContainersMap
+}
+
+func (builder *containerSpecDTOBuilder) getControllerUID(entityType metrics.DiscoveredEntityType, entityKey string) (string, error) {
+	ownerUIDMetricId := metrics.GenerateEntityStateMetricUID(entityType, entityKey, metrics.OwnerUID)
+	ownerUIDMetric, err := builder.metricsSink.GetMetric(ownerUIDMetricId)
+	if err != nil {
+		return "", fmt.Errorf("Error getting owner UID for pod %s --> %v\n", entityKey, err)
```
ading1977, line 73: There is no need to append new line "\n".

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	ownerUID := ownerUIDMetric.GetValue()
+	controllerUID, ok := ownerUID.(string)
+	if !ok {
+		return "", fmt.Errorf("Error getting owner UID for pod %s\n", entityKey)
```
ading1977, line 78: Same as above.

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
marvelyue: Thanks Meng. Will update based on your suggestion

dblinn:
@marvelyue 

As discussed, just a couple small asks:
1. Change the value of "infinity" to 1e12 from 1e9. Since memory is in kB this is a value of 1 terabyte. 1 TB of memory is a lot, but for the sum of all the memory available in, say, someone's production namespace, this is probably too small. I feel like a value of 1e12 (1 petabyte for memory) would be more likely to avoid running into issues.
2. Let's drop the 'k8s-namespace-' prefix on the namespace ID and keys. This appears in a lot of places and takes up a lot of memory and makes string comparisons more expensive unnecessarily. There's really no use in including the entity type in the ID field when we already have a separate dedicated field for explicitly identifying the entity type.

Overall looks very good thanks!

PR #394, opened 2020-03-31
Update deployment yaml

ading1977:
This PR update `kubeturbo` deployment yamls:

* Update the default image tag to `7.21.2`
* Support the `targetType` configuration in helm charts.
* Add instructions on how to enable registering `kubeturbo` as a probe without adding a target.

### DIFF
--- a/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
+++ b/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
@@ -5,7 +5,7 @@
 # Replace the image with desired version
 image:
   repository: turbonomic/kubeturbo
-  tag: 7.21.1
+  tag: 7.21.2
   pullPolicy: IfNotPresent
 
 #nameOverride: ""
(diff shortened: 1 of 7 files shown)


deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
```diff
@@ -27,9 +27,13 @@ restAPIConfig:
...
   opsManagerPassword: Turbo_password
 
-# Enabling targetConfig and providing a value will give a name to your cluster
-#targetConfig:
-#  targetName: Name_Each_Cluster
+# Specify a targetName only will register a probe with type Kubernetes-<targetName>, as well as adding your
+# cluster as a target with the name Kubernetes-<targetName>.
+# If you want to register a probe without adding your cluster as a target, specify a targetType only. The probe
+# will appear as a Cloud Native probe in the UI with a type Kubernetes-<targetType>.
+targetConfig:
```
esara: so the problem with this, that instead of a generated clustername - every kubeturbo instance/every cluster would be named the same - confusing - I think we should keep the default behaviour and let the customer name their clusters

deploy/kubeturbo/values.yaml
```diff
@@ -27,9 +27,13 @@ restAPIConfig:
...
   opsManagerPassword: Turbo_password
 
-# Enabling targetConfig and providing a value will give a name to your cluster
-#targetConfig:
-#  targetName: Name_Each_Cluster
+# Specify a targetName only will register a probe with type Kubernetes-<targetName>, as well as adding your
+# cluster as a target with the name Kubernetes-<targetName>.
+# If you want to register a probe without adding your cluster as a target, specify a targetType only. The probe
+# will appear as a Cloud Native probe in the UI with a type Kubernetes-<targetType>.
+targetConfig:
```
esara: same as before

PR #401, opened 2020-04-13
Create new WorkloadController entity for new container model

marvelyue:
**Intent**:
The intent is to
1. Explicitly represent k8s controller as an entity in the supply chain so as to show the aggregated data of a controller and in the future store the spec data.
2. Have WorkloadController entity sell VCPULimitQuota/VMemLimitQuotaand VCPURequestQuota/VMemRequestQuota commodities as a "_reseller_" to take K8s namespace resource quota into consideration for container resizing

**Background**:
The supply chain and resource commodities bought and sold by WorkloadController are shown in the screenshot below:
Screen Shot 2020-04-13 at 11 16 05
Screen Shot 2020-04-13 at 11 16 24
where WorkloadController sells commodities to Pod and buys commodities from Namespace.

**Implementation**:
1. Create `KubeController` type in `kube_cluster.go` to represent a k8s controller.
2. In `controller_metrics_collector.go`, 1) create `KubeController` entity by retrieving pod owner name, type and UID from metrics sink by calling `getKubeControllerInfo` func; 2) collect allocation resource metrics for K8s controllers from the pods by a given discovery worker, where usage values are aggregated from pods and capacity values are from namespace quota capacity. Note that for a pod deployed by Deployment controller, the created `KubeController` entity has controllerType as "Deployment" not "ReplicaSet".
3. Send list of `KubeController` discovered by a discovery worker in `k8s_discovery_worker::executeTask`
4. Collect all KubeControllers discovered by different discovery workers in `result_collector:: Collect` 
5. In `controller_discovery_worker.go`, merge the KubeControllers discovered by different discovery workers and invoke `workload_controller_entity_dto_builder` to create entity DTOs.
6. Create WorkloadController entity DTOs with commodities bought and sold as well as `WorkloadControllerData` with controller type data in `workload_controller_entity_dto_builder`. 
7. Add `EntityDTO_WORKLOAD_CONTROLLER` in `registration_client` and build `WorkloadControllerSupplyBuilder` in `supply_chain_factory`
8. Rename `k8sResourceQuotasDiscoveryWorker` to `k8sNamespaceDiscoveryWorker` to avoid confusion.

**Unit Test**:
Created new unit tests `controller_metrics_collector_test.go` and `workload_controller_entity_dto_builder_test.go`

**Testing Done**:
Added kubeturbo target (dc11) with the changes and dumped discovery dtos.
1. In the supply chain, WorkloadControllers are buying from Namespaces:
Screen Shot 2020-04-13 at 11 33 53
2. Here's a snippet of a WorkloadController entityDTO where quota is configured on the corresponding namespace:
```json
entityDTO {
  entityType: WORKLOAD_CONTROLLER
  id: "1d187381-0ba6-11ea-878a-005056803aed"
  displayName: "sc-testing"
  commoditiesSold {
    commodityType: VCPU_REQUEST_QUOTA
    key: "1d187381-0ba6-11ea-878a-005056803aed"
    used: 2397.4002
    capacity: 2663.778
    peak: 2397.4002
    resizable: false
  }
  ...
  commoditiesBought {
    providerId: "10c7cf65-0ba6-11ea-878a-005056803aed"
    bought {
      commodityType: VCPU_LIMIT_QUOTA
      key: "10c7cf65-0ba6-11ea-878a-005056803aed"
      used: 5061.178199999999
      peak: 5061.178199999999
      resizable: false
    }
    ...
    providerType: NAMESPACE
    actionEligibility {
      movable: false
    }
  }
  powerState: POWERED_ON
  actionEligibility {
    suspendable: false
    cloneable: false
  }
  workload_controller_data {
    deployment_data {
    }
  }
```
where
* WorkloadControlelr entity sells commodities with `resizable` as false
* entity buys commodities from Namespace with `resizable` as false and movable as false
* suspendable and cloneable are false
* entity has `deployment_data` in `workload_controller_data` which means this controller is a Deployment

3. If the corresponding Namespace has no quota configured and all containers in the pods have no compute resources (limits or requests) configured, the WorkloadController entity sells the commodity with capacity as "infinity" as 1e12 and usage/peak as 0:
```json
entityDTO {
  entityType: WORKLOAD_CONTROLLER
  id: "90e3886e-0489-11ea-878a-005056803aed"
  displayName: "kubeturbo-arsen-dep-6.4"
  commoditiesSold {
    commodityType: VCPU_REQUEST_QUOTA
    key: "90e3886e-0489-11ea-878a-005056803aed"
    used: 0.0
    capacity: 1.0E12
    peak: 0.0
    resizable: false
  }
```

4. Commodities of WorkloadController are correctly shown in "Capacity and Usage" widget:
Screen Shot 2020-04-13 at 11 44 15
where the first four are commodities sold and last four are commodities bought from the Namespace.

### DIFF
--- a/pkg/discovery/dtofactory/namespace_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/namespace_entity_dto_builder.go
@@ -40,6 +40,10 @@ func (builder *namespaceEntityDTOBuilder) BuildEntityDTOs() ([]*proto.EntityDTO,
 		}
 		entityDTOBuilder.SellsCommodities(commoditiesSold)
 
+		// Namespace entity cannot be provisioned or suspended by Turbonomic analysis
+		entityDTOBuilder.IsProvisionable(false)
+		entityDTOBuilder.IsSuspendable(false)
+
 		entityDTOBuilder.WithPowerState(proto.EntityDTO_POWERED_ON)
 
 		// build entityDTO.
@@ -91,6 +95,7 @@ func (builder *namespaceEntityDTOBuilder) getQuotaCommoditiesSold(kubeNamespace 
 
 		commSoldBuilder := sdkbuilder.NewCommodityDTOBuilder(cType)
 		commSoldBuilder.Used(usedValue)
+		commSoldBuilder.Peak(usedValue)
 		commSoldBuilder.Capacity(capacityValue)
 		commSoldBuilder.Resizable(false)
 		commSoldBuilder.Key(kubeNamespace.UID)
--- a/pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
@@ -1,0 +1,180 @@
+package dtofactory
+
+import (
+	"github.com/golang/glog"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
+	"github.com/turbonomic/kubeturbo/pkg/util"
+	sdkbuilder "github.com/turbonomic/turbo-go-sdk/pkg/builder"
+	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+)
+
+type workloadControllerDTOBuilder struct {
+	kubeControllersMap map[string]*repository.KubeController
+	namespaceUIDMap    map[string]string
+}
+
+func NewWorkloadControllerDTOBuilder(kubeControllersMap map[string]*repository.KubeController,
+	namespaceUIDMap map[string]string) *workloadControllerDTOBuilder {
+	return &workloadControllerDTOBuilder{
+		kubeControllersMap: kubeControllersMap,
+		namespaceUIDMap:    namespaceUIDMap,
+	}
+}
+
+// Build entityDTOs based on the given map from controller UID to KubeController entity.
+func (builder *workloadControllerDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error) {
+	var result []*proto.EntityDTO
+	for _, kubeController := range builder.kubeControllersMap {
+		// Id
+		workloadControllerId := kubeController.UID
+		entityDTOBuilder := sdkbuilder.NewEntityDTOBuilder(proto.EntityDTO_WORKLOAD_CONTROLLER, workloadControllerId)
+
+		// Display name
+		workloadControllerDisplayName := kubeController.Name
+		entityDTOBuilder.DisplayName(workloadControllerDisplayName)
+
+		// Resource commodities sold
+		commoditiesSold, err := builder.getCommoditiesSold(kubeController)
+		if err != nil {
+			glog.Errorf("Error creating commoditiesSold for %s: %v", kubeController.GetFullName(), err)
+			continue
+		}
+		entityDTOBuilder.SellsCommodities(commoditiesSold)
+
+		// Resource commodities bought from namespace
+		namespaceUID, exists := builder.namespaceUIDMap[kubeController.Namespace]
+		if exists {
+			commoditiesBought, err := builder.getCommoditiesBought(kubeController, namespaceUID)
+			if err != nil {
+				glog.Errorf("Error creating commoditiesBought for %s: %v", kubeController.GetFullName(), err)
+				continue
+			}
+			entityDTOBuilder.Provider(sdkbuilder.CreateProvider(proto.EntityDTO_NAMESPACE, namespaceUID)).BuysCommodities(commoditiesBought)
+			// Set movable false to avoid moving WorkloadController across namespaces
+			entityDTOBuilder.IsMovable(proto.EntityDTO_NAMESPACE, false)
+		} else {
+			glog.Errorf("Failed to get namespaceUID from namespace %s for controller %s", kubeController.Namespace,
+				kubeController.GetFullName())
+		}
+
+		// Create WorkloadControllerData to store controller type data
+		entityDTOBuilder.WorkloadControllerData(builder.createWorkloadControllerData(kubeController))
+
+		// WorkloadController cannot be provisioned or suspended by Turbonomic analysis
+		entityDTOBuilder.IsProvisionable(false)
+		entityDTOBuilder.IsSuspendable(false)
+
+		entityDTOBuilder.WithPowerState(proto.EntityDTO_POWERED_ON)
+
+		entityDTO, err := entityDTOBuilder.Create()
+		if err != nil {
+			glog.Errorf("failed to build WorkloadController[%s] entityDTO: %v", workloadControllerDisplayName, err)
+		}
+		result = append(result, entityDTO)
+	}
+	return result, nil
+}
+
+func (builder *workloadControllerDTOBuilder) getCommoditiesSold(kubeController *repository.KubeController) ([]*proto.CommodityDTO, error) {
+	var commoditiesSold []*proto.CommodityDTO
+	for resourceType, resource := range kubeController.AllocationResources {
+		commodityType, exist := rTypeMapping[resourceType]
+		if !exist {
+			glog.Errorf("ResourceType %s is not supported", resourceType)
+			continue
+		}
+		commSoldBuilder := sdkbuilder.NewCommodityDTOBuilder(commodityType)
+		commSoldBuilder.Used(resource.Used)
+		commSoldBuilder.Peak(resource.Used)
+		commSoldBuilder.Capacity(resource.Capacity)
+		commSoldBuilder.Resizable(false)
+		commSoldBuilder.Key(kubeController.UID)
+		commSold, err := commSoldBuilder.Create()
+		if err != nil {
+			glog.Errorf("%s: Failed to build commodity sold %s: %s", kubeController.GetFullName(), commodityType, err)
+			continue
+		}
+		commoditiesSold = append(commoditiesSold, commSold)
+	}
+	return commoditiesSold, nil
+}
+
+func (builder *workloadControllerDTOBuilder) getCommoditiesBought(kubeController *repository.KubeController,
+	namespaceUID string) ([]*proto.CommodityDTO, error) {
+	var commoditiesBought []*proto.CommodityDTO
+	for resourceType, resource := range kubeController.AllocationResources {
+		commodityType, exist := rTypeMapping[resourceType]
+		if !exist {
+			glog.Errorf("ResourceType %s is not supported", resourceType)
+			continue
+		}
+		commBoughtBuilder := sdkbuilder.NewCommodityDTOBuilder(commodityType)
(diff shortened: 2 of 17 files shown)


pkg/discovery/dtofactory/namespace_entity_dto_builder.go
```diff
@@ -40,6 +40,10 @@ func (builder *namespaceEntityDTOBuilder) BuildEntityDTOs() ([]*proto.EntityDTO,
 		}
 		entityDTOBuilder.SellsCommodities(commoditiesSold)
 
+		// Namespace entity cannot be provisioned or suspended by Turbonomic analysis
+		entityDTOBuilder.IsProvisionable(false)
+		entityDTOBuilder.IsSuspendable(false)
```
irfanurrehman, line 45: What actions are we actually expecting on the workloadController ?

pkg/discovery/k8s_discovery_client.go
```diff
@@ -224,15 +224,31 @@ func (dc *K8sDiscoveryClient) discoverWithNewFramework(targetID string) ([]*prot
...
+	if err != nil {
+		glog.Errorf("Failed to discover namespaces from current Kubernetes cluster with the new discovery framework: %s", err)
+	} else {
+		glog.V(2).Infof("There are %d namespace entityDTOs.", len(namespaceDtos))
+		entityDTOs = append(entityDTOs, namespaceDtos...)
+	}
+
+	// K8s workload controller discovery worker to create WorkloadController DTOs
+	controllerDiscoveryWorker := worker.NewK8sControllerDiscoveryWorker(clusterSummary)
+	workloadControllerDtos, _ := controllerDiscoveryWorker.Do(kubeControllerList)
```
irfanurrehman, line 245: Assign the variable `err` correctly here please.

pkg/discovery/worker/controller_metrics_collector.go
```diff
@@ -0,0 +1,171 @@
...
+				controllerUID)
+			kubeNamespace, namespaceExists := kubeNamespaceMap[namespace]
+			if !namespaceExists {
+				glog.Errorf("Namespace %s does not exist in cluster %s", namespace, collector.cluster.Name)
+				continue
+			}
+			// Add quota resources to KubeController entity with capacity as namespace quota capacity
+			for _, resourceType := range metrics.QuotaResources {
+				namespaceQuotaResource, err := kubeNamespace.GetAllocationResource(resourceType)
+				resourceCapacity := repository.DEFAULT_METRIC_CAPACITY_VALUE
```
irfanurrehman, line 67: I think setting the quota value to unlimited for namespace DTOs (if no quota exists) and accordingly the same for the workloadController is erroneous. I think it should be set to the clusters total usable resource capacity; atleast for the requests (as limits can overshoot the cluster/nodes capacity). The clusters capacity equal the value summed up for worker/schedulable nodes.
Let me know your thoughts!

pkg/discovery/worker/controller_metrics_collector.go
```diff
@@ -0,0 +1,171 @@
...
+					// from certain amount of nodes run by a k8s discovery worker. So at this point, there's no way to
+					// calculate average CPU frequency of all nodes in the cluster.
+					resourceCapacity = namespaceQuotaResource.Capacity
+				}
+				kubeController.AddAllocationResource(resourceType, resourceCapacity, repository.DEFAULT_METRIC_VALUE)
+			}
+			kubeControllersMap[controllerUID] = kubeController
+			kubeControllerList = append(kubeControllerList, kubeController)
+		}
+		kubeController.Pods = append(kubeController.Pods, pod)
```
irfanurrehman, line 84: Whats the reason of collecting these pods, I haven't seen these used so far.

pkg/discovery/dtofactory/namespace_entity_dto_builder.go
marvelyue: Right, we don't want to generate any action on Namespaces and WorkloadControllers.

pkg/discovery/k8s_discovery_client.go
marvelyue: Nice catch, thanks. Will add the missing `err` here

pkg/discovery/worker/controller_metrics_collector.go
marvelyue: This is exactly what @dblinn and I initially thought. It makes sense that if namespace quota is not configured, we can set the capacity as the total usable resource capacity of the whole cluster for requests, while the capacity of limits quota commodities should be much larger than cluster capacity to allow users to overcommit resource limits.

Then we considered minimizing the impact on Market analysis result if quota is not configured. In analysis we want to only resize down requests not resize up. So setting the capacity to infinity for not only limits but also requests will have the least effect on analysis result so that resource requests won't be resized from the missing namespace quota.

pkg/discovery/worker/controller_metrics_collector.go
marvelyue: I should have added a TODO here. I plan to use the pods of each controller to create ContainerSpec entity to simplify the logic in the committed changes in `container_spec_entity_dto_builder.go` (https://github.com/marvelyue/kubeturbo/blob/14ed3c34c23a28f257059d52a093e373d418c46e/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go)

ContainerSpec entity models a shared portion of container replicas defined by the K8s controller. Currently the logic in `container_spec_entity_dto_builder.go` is to get controller UID from all pods and create a map from controllerUID to container names. After creating KubeController entities with corresponding pods, we'll be able to simply the logic.

pkg/discovery/worker/controller_metrics_collector.go
irfanurrehman: Makes sense, thanks for the explaination!

pkg/discovery/worker/controller_metrics_collector.go
irfanurrehman: Ack!

pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
```diff
@@ -0,0 +1,180 @@
...
+			entityDTOBuilder.Provider(sdkbuilder.CreateProvider(proto.EntityDTO_NAMESPACE, namespaceUID)).BuysCommodities(commoditiesBought)
+			// Set movable false to avoid moving WorkloadController across namespaces
+			entityDTOBuilder.IsMovable(proto.EntityDTO_NAMESPACE, false)
+		} else {
+			glog.Errorf("Failed to get namespaceUID from namespace %s for controller %s", kubeController.Namespace,
+				kubeController.GetFullName())
+		}
+
+		// Create WorkloadControllerData to store controller type data
+		entityDTOBuilder.WorkloadControllerData(builder.createWorkloadControllerData(kubeController))
```
enlinxu, line 61: I am trying to find the answer by reading the review but I still couldn't figure out. My question here is , how are we dealing with pods under replicaset then under deployment. Are we recursively find the owner and only create one workloadcontroller in this case we we discussed? I couldn't find the related logic in this review.

marvelyue: Sorry I didn't explain this clearly in the description. Yes we create one WorkloadController with controller type as "Deployment" in this case.

We have already had existing logic to get the pod owner in `getPodOwner` func in `cluster_monitor.go` by calling `GetPodGrandInfo` func, which returns "Deployment" as the owner "type" of the pod in this case instead of "ReplicaSet". Then we store the pod owner metrics (owner name, type and UID) in the metrics sink by calling `genOwnerMetrics` in `cluster_monitor.go`.

In the changes here in this PR, `KubeController` entity is created by retrieving pod owner name, type and UID from metrics sink by calling `getKubeControllerInfo` func in `controller_metrics_collector.go`. The owner type (controller type) retrieved here is already "Deployment" in this case. Eventually, we convert `KubeController` entity to WorkloadController entity DTO in `workload_controller_entity_dto_builder.go`.

enlinxu: Thanks Yue, that makes a lot of sense.

pkg/discovery/worker/controller_discovery_worker.go
```diff
@@ -0,0 +1,80 @@
...
+	return &K8sControllerDiscoveryWorker{
+		cluster: cluster,
+	}
+}
+
+// Controller discovery worker collects KubeController entities discovered by different discovery workers.
+// It merges the pods belonging to the same controller but discovered by different discovery workers, and aggregates
+// allocation resources usage of the same KubeController from different workers.
+// Then it creates entity DTOs for the K8s controllers to be sent to the Turbonomic server.
+func (worker *K8sControllerDiscoveryWorker) Do(kubeControllers []*repository.KubeController) ([]*proto.EntityDTO, error) {
```
dblinn, line 27: nit: maybe document that this must be run after cluster namespaces have been populated.

pkg/discovery/worker/controller_discovery_worker.go
```diff
@@ -0,0 +1,80 @@
...
+			if !exists {
+				glog.Errorf("Namespace %s does not exist in cluster %s", kubeController.Namespace, worker.cluster.Name)
+				continue
+			}
+			for resourceType, resource := range kubeController.AllocationResources {
+				// For CPU resources, convert the capacity values expressed in number of cores to MHz.
+				// Skip the conversion if capacity value is repository.DEFAULT_METRIC_CAPACITY_VALUE (infinity), which
+				// means resource quota is not configured on the corresponding namespace.
+				if metrics.IsCPUType(resourceType) && resource.Capacity != repository.DEFAULT_METRIC_CAPACITY_VALUE &&
+					kubeNamespace.AverageNodeCpuFrequency > 0.0 {
```
dblinn, line 45: Under what circumstances would averageNodeCpuFrequency not be greater than zero? 

Since this value doesn't change inside the loop, would it make sense to move this check outside the loop and then log a warning if it is <= 0 instead of just silently skipping the conversion?

pkg/discovery/worker/controller_metrics_collector.go
```diff
@@ -0,0 +1,171 @@
...
+			// For CPU resources, convert the usage values expressed in cores to MHz based on the CPU frequency of the
+			// corresponding node where the given pod is located
+			cpuFrequency, err := collector.getNodeCPUFrequency(pod)
+			if err != nil {
+				glog.Errorf("Error getting node cpuFrequency from pod %s: %v", podKey, err)
+				continue
+			}
+			if cpuFrequency > 0.0 {
+				resourceUsed *= cpuFrequency
+			}
```
dblinn, line 148: nit: Do we want to log this conversion as well since we log the capacity conversions?

PR #405, opened 2020-04-20
Node cpu frequency getter via an ephemeral pod

irfanurrehman:
This code needs some cleanup and additional handling for `GetMachineCpuFrequency()`used by `ContainerResizer` but it gives the clear idea of what its trying to achieve.
It works on 1.18.

### DIFF
--- a/pkg/kubeclient/kubelet_client.go
+++ b/pkg/kubeclient/kubelet_client.go
@@ -3,19 +3,23 @@ package kubeclient
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/golang/glog"
-	cadvisorapi "github.com/google/cadvisor/info/v1"
-	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"io/ioutil"
-	"k8s.io/api/core/v1"
-	netutil "k8s.io/apimachinery/pkg/util/net"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/transport"
-	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 	"net/http"
 	"net/url"
 	"sync"
 	"time"
+
+	"github.com/golang/glog"
+	cadvisorapi "github.com/google/cadvisor/info/v1"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/types"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
+	v1 "k8s.io/api/core/v1"
+	netutil "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 )
 
 const (
@@ -33,15 +37,16 @@ type KubeHttpClientInterface interface {
 	ExecuteRequestAndGetValue(host string, endpoint string, value interface{}) error
 	GetSummary(host string) (*stats.Summary, error)
 	GetMachineInfo(host string) (*cadvisorapi.MachineInfo, error)
-	GetMachineCpuFrequency(host string) (uint64, error)
+	GetNodeCpuFrequency(node *v1.Node) (float64, error)
 }
 
 // Cache structure.
 // TODO(MB): Make sure the nodes that are no longer discovered are being removed from the cache!
 type CacheEntry struct {
 	statsSummary *stats.Summary
-	machineInfo  *cadvisorapi.MachineInfo
-	used         bool
+	// nodes cpu frequency in MHz as expected by server
+	nodeCpuFreq *float64
+	used        bool
 }
 
 // Cleanup the cache.
@@ -75,11 +80,13 @@ func (client *KubeletClient) CleanupCache(nodes []*v1.Node) int {
 // Since http.Client is thread safe (https://golang.org/src/net/http/client.go)
 // KubeletClient is also thread-safe if concurrent goroutines won't change the fields.
 type KubeletClient struct {
-	client    *http.Client
-	scheme    string
-	port      int
-	cache     map[string]*CacheEntry
-	cacheLock sync.Mutex
+	client              *http.Client
+	scheme              string
+	port                int
+	cache               map[string]*CacheEntry
+	cacheLock           sync.Mutex
+	fallbkCpuFreqGetter *NodeCpuFrequencyGetter
+	defaultCpuFreq      float64
 }
 
 func (client *KubeletClient) ExecuteRequestAndGetValue(host string, endpoint string, value interface{}) error {
@@ -137,6 +144,10 @@ func (client *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
 				return nil, err
 			}
 			glog.V(2).Infof("unable to retrieve machine[%s] summary: %v. Using cached value", host, err)
+			// TODO(irfanurrehman): Improve the node check [fn checknode()].
+			// This looks flawed. The same is also used as checknode;
+			// if ExecuteRequestAndGetValue() returns error, checknode should get error
+			// rather then a value from cache.
 			return entry.statsSummary, nil
 		} else {
 			glog.Errorf("failed to get machine[%s] summary: %v. No cache available", host, err)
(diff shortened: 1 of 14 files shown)


irfanurrehman:
This is now complete, tested and ready for review.
The flow is as below:
- Discovery queries the `/spec` endpoint for machineinfo, if found caches the cpufrequency of that node.
- If `/spec` returns error, it tries to get cpu frequency via an ephemeral pod, and caches it.
- If for some reason this also does not work for a node, cpu frequency of last queried node is used (assuming that most clusters are homogeneous).
- If that too isn't available a default of `2000 MHz` is set as cpufreq.

enlinxu:
I see a potential problem here. In the case of when fall back job also failed (for whatever reasons), we are falling back to the default capacilty. But in our supply chain definition of kubeturbo, we have 
		PatchSoldMetadata(proto.CommodityDTO_VCPU, fieldsUsedCapacityPeak).
Meaning for nodes, we are patching used, capacity and peak. This will be a problem since the capacity now is fake and we should override the real capacity discovered by IaaS. 
Irfan, can you verify if this is the case by testing in a stitched environment with default cpu capacity (you can just assign a random capacity and see if it overrides the real capacity on the vm discovered by the IaaS target.

pkg/kubeclient/kubelet_client.go
```diff
@@ -158,7 +170,73 @@ func (client *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
...
+	nodeFreq := client.defaultCpuFreq
+	minfo, err := client.GetMachineInfo(ip)
+	if err != nil {
+		glog.Errorf("Failed to get machine[%s] cpu.frequency from kubelet: %v. Will try pod based getter.", node.Name, err)
+		nodeIsActive := util.NodeIsReady(node) && util.NodeIsSchedulable(node)
+		if nodeIsActive {
+			nodeFreq, err = client.fallbkCpuFreqGetter.GetFrequency(node.Name)
+			if err != nil {
+				glog.Errorf("Failed to get cpufreq from getter %s: %s. Default value will be used.", node.Name, err)
+			}
```
enlinxu, line 203: maybe you should also do client.defaultCpuFreq = nodeFreq in case jobs are created successfully for some of the nodes, we may want to use this value for the failed ones.

irfanurrehman: Its exactly what is done on line 209 below.
So the default cpuFreq with be updated even if it could be discovered from only one node.

enlinxu: Line 209 is only invoked when machineinfo is available. My point is if we fall back to the job approach, and only some of the jobs are successfully deployed (others are not schedulable for example), then we should still override the default with the one we got from the successful job.

irfanurrehman:
> I see a potential problem here. In the case of when fall back job also failed (for whatever reasons), we are falling back to the default capacilty. But in our supply chain definition of kubeturbo, we have
> PatchSoldMetadata(proto.CommodityDTO_VCPU, fieldsUsedCapacityPeak).
> Meaning for nodes, we are patching used, capacity and peak. This will be a problem since the capacity now is fake and we should override the real capacity discovered by IaaS.
> Irfan, can you verify if this is the case by testing in a stitched environment with default cpu capacity (you can just assign a random capacity and see if it overrides the real capacity on the vm discovered by the IaaS target.

@enlinxu On second thoughts, I think the better mechanism would be to not fallback on default value at all, and fail kubeturbo in case it cannot discover the cpufreq from even (altleast) 1 node in the cluster. That way we will be able to catch environment issues, else kubeturbo will keep working even with env issues (and fake cpufreq).

pkg/kubeclient/kubelet_client.go
irfanurrehman: You are right. Will update.

irfanurrehman:
> > I see a potential problem here. In the case of when fall back job also failed (for whatever reasons), we are falling back to the default capacilty. But in our supply chain definition of kubeturbo, we have
> > PatchSoldMetadata(proto.CommodityDTO_VCPU, fieldsUsedCapacityPeak).
> > Meaning for nodes, we are patching used, capacity and peak. This will be a problem since the capacity now is fake and we should override the real capacity discovered by IaaS.
> > Irfan, can you verify if this is the case by testing in a stitched environment with default cpu capacity (you can just assign a random capacity and see if it overrides the real capacity on the vm discovered by the IaaS target.
> 
> @enlinxu On second thoughts, I think the better mechanism would be to not fallback on default value at all, and fail kubeturbo in case it cannot discover the cpufreq from even (altleast) 1 node in the cluster. That way we will be able to catch environment issues, else kubeturbo will keep working even with env issues (and fake cpufreq).

So, I did sleep over this thought without changing the code today and I am now more convinced that not falling back to preset default value should be the way to go.
1. Access issues (kubeturbo does not have access to create an unprivileged pod) wont be caught.
2. Image issues wont be caught (for example kubeturbo image is made internally avaliable and busybox image is not).
As a user, I would rather have a know to get configuration errors and correct them.

I would update the code accordingly.

ading1977:
@irfanurrehman, I do agree that `kubeturbo` should fail if we are not able to get cpu frequency through the fallback job. But I do think we need to make sure that relevant messages are logged (e.g., insufficient permission, or image pull failure), such that user knows what has happened and how to fix the problem. Are these messages being logged (e.g., we may have to get pod event to see certain errors)?

irfanurrehman:
> @irfanurrehman, I do agree that `kubeturbo` should fail if we are not able to get cpu frequency through the fallback job. But I do think we need to make sure that relevant messages are logged (e.g., insufficient permission, or image pull failure), such that user knows what has happened and how to fix the problem. Are these messages being logged (e.g., we may have to get pod event to see certain errors)?

@ading1977 yes all the errors are caught in code and are already logged. I am updating the code to do away with the fallback default value and fail discovery if kubeturbo cannot discover cpufreq from any node.

pkg/kubeclient/kubelet_client.go
```diff
@@ -150,59 +161,89 @@ func (client *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
...
+		nodeFreq = util.MetricKiloToMega(float64(minfo.CpuFrequency))
 	}
-	// Fill in the cache
+
+	if !freqDiscovered {
+		// We probably could not discover the cpufreq by any means and default
+		// value is what we need to stick to. However, we might be able to discover
+		// a valid one in the next discovery cycle.
+
+		return client.defaultCpuFreq, nil
```
ading1977, line 216: So just to confirm, in case we cannot discover cpufreq from any node, we still use the `DefaultCpuFreq`, right? If so, as @enlinxu has suggested, this will overwrite the used and capacity value of a node with the fake value in a stitched environment. In this situation, I still think we should fail the discovery (or the validation). Sending no value is still better than sending wrong values.

irfanurrehman: The issue of patching with wrong default in a stitched env is there ( I did verify it and the infra values are overridden by the ones sent from kubeturbo).

Having said that; I also had a chat with @esara before posting this yesterday. His point of view is that this is most likely going to be used for demo purposes in the demo envs (probably until we land a proper milicore based solution). Its less likely that we would have a real customer using 1.18+ cluster already. For demo envs, its also more likely that there are transient issues when we start off (for example image not available internally), where if we fail discovery we wont have anything to show in supply chain. It makes sense to show something, rather then nothing in such cases.

I have updated the code to catch all possible issues and they will be Error logged anyways.

If it might make more sense, I can also updated the default cpu freq to `2600 MHz` (in place of `2000`), which is what I have seen more often (2.6/2.7 GHz) across clusters.

pkg/kubeclient/kubelet_client.go
```diff
@@ -150,59 +161,89 @@ func (client *KubeletClient) GetSummary(host string) (*stats.Summary, error) {
...
+	}
+
+	// We optimistically set a cpu frequency of last discovered node as default.
+	// Clusters more often then not have similar nodes.
+	client.defaultCpuFreq = nodeFreq
```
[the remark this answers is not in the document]
irfanurrehman: I don't think there is any meaning to make this thread safe. Each node could be discovered in its own goroutine (say 4 cores assigned to kubeturbo discovering 4 nodes). Possible example scenario is that one node discovers its frequency and 3 others don't, where they need to fall back on default. The node that discovers it updates the default, before 2 (of the other 3) fail, so those 2 get the updated default and the third one doesn't (which probably failed earlier and picked the original default). This will stay so for this discovery, but eventually correct itself in the next.
Let me know if you think I am missing something.

pkg/kubeclient/kubelet_client.go
ading1977: Thanks for the explanation. Make senses for demo envs.

pkg/kubeclient/kubelet_client.go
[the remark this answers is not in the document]
ading1977: Thanks, I am ok with this.

PR #408, opened 2020-04-27
Set inactive commodities sold by ContainerSpec entities

marvelyue:
**Intent and Background**:
The VCPU and VMem commodities sold by ContainerSpecs are designed to model the shared utilization and usage data among all container replicas specified by the ContainerSpec. The shared utilization and usage data can be persisted in the platform so that even when the ephemeral replicas come and go, they can be supplied with that history.

The intent is to set VCPU and VMem commodities sold by ContainerSpec entities and implement a framework to define the container utilization and usage data aggregation.

**Implementation**:
1. Defined `ContainerSpec` type to model the shared portion of individual container replicas defined by the controller that manages the pods where the containers run.
2. Create ContainerSpecs with individual container replica commodity DTOs in `container_dto_builder.go` run by each discovery worker
3. Collect all ContainerSpecs with ContainerSpecs from different discovery workers in `result_collector.go`
4. In `container_spec_discovery_worker.go`, merge the commodities data of container replicas belonging to the same ContainerSpec but discovered by different discovery workers
5. In `container_spec_entity_dto_builder.go`, create inactive commodities sold by ContainerSpec and aggregate container utilization and usage data with the default data aggregator and then return ContainerSpec entity DTOs.
6. Create utilization and usage data aggregator using strategy pattern. 
6.1  In `container_utilization_data_aggregator.go`, there're `allUtilizationDataAggregator` and `maxUtilizationDataAggregator` to aggregate container utilization data based on the given commodity DTOs of a commodity Type from a set of container replicas for a ContainerSpec entity.
6.2 In `container_usage_data_aggregator.go`, there're `avgUsageDataAggregator` and `maxUsageDataAggregator` to aggregate container usage data (capacity, used and peak) based on the given commodity DTOs of a commodity type from a set of container replicas for a ContainerSpec entity 

The corresponding `Aggregate` function to populate the utilization and usage data will be implemented in a different story. Will post a following PR for this.

**Unit Tests**:
Update and create new unit tests. And all unit tests are passed

**Testing Done**:
Add a kubeturbo target with the changes, ContainerSpec entities can be correctly created with such discovery dto:
```json
entityDTO {
  entityType: CONTAINER_SPEC
  id: "e9384cfc-70e7-11ea-bda6-005056803aed/kubeturbo"
  displayName: "kubeturbo"
  commoditiesSold {
    commodityType: VCPU
    used: 0.0
    capacity: 0.0
    peak: 0.0
    active: false
    utilizationData {
      lastPointTimestampMs: 0
      intervalMs: 0
    }
  }
  commoditiesSold {
    commodityType: VMEM
    used: 0.0
    capacity: 0.0
    peak: 0.0
    active: false
    utilizationData {
      lastPointTimestampMs: 0
      intervalMs: 0
    }
  }
  monitored: false
  actionEligibility {
  }
}
```
There's no data point in utilizationData yet. The utilizationData, used, peak and capacity values will be populated in a different story.

### DIFF
--- a/pkg/discovery/worker/container_spec_discovery_worker.go
+++ b/pkg/discovery/worker/container_spec_discovery_worker.go
@@ -24,7 +24,7 @@ func NewK8sContainerSpecDiscoveryWorker() *k8sContainerSpecDiscoveryWorker {
 func (worker *k8sContainerSpecDiscoveryWorker) Do(containerSpecs []*repository.ContainerSpec,
 	utilizationDataAggStrategy, usageDataAggStrategy string) ([]*proto.EntityDTO, error) {
 	// Get data aggregators based on the given data aggregation strategies
-	utilizationDataAggregator, usageDataAggregator := worker.getContainerDataAggregator(utilizationDataAggStrategy, usageDataAggStrategy)
+	utilizationDataAggregator, usageDataAggregator := worker.getContainerDataAggregators(utilizationDataAggStrategy, usageDataAggStrategy)
 	// Create containerSpecs map from ContainerSpec ID to ContainerSpec with merge commodities data of container replicas
 	containerSpecMap := worker.createContainerSpecMap(containerSpecs)
 	containerSpecEntityDTOBuilder := dtofactory.NewContainerSpecDTOBuilder(containerSpecMap, utilizationDataAggregator,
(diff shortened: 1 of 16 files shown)


pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -343,27 +347,19 @@ func (worker *k8sDiscoveryWorker) buildDTOs(currTask *task.Task) ([]*proto.Entit
 
 	// 3. build entityDTOs for containers
```
dblinn, line 348: nit: maybe we should update the comment to say "build entityDTOs for containers and containerSpecs"

pkg/discovery/worker/container_spec_discovery_worker.go
```diff
@@ -0,0 +1,87 @@
...
+				commodities, exists := containerSpec.ContainerCommodities[commodityType]
+				if !exists {
+					glog.Errorf("%s commodities do not exist for ContainerSpec %s", commodityType, containerSpec.ContainerSpecId)
+					continue
+				}
+				existingCommodities = append(existingCommodities, commodities...)
+				existingContainerSpec.ContainerCommodities[commodityType] = existingCommodities
+			}
+			// Increment number of container replicas
+			existingContainerSpec.ContainerReplicas++
```
dblinn, line 83: Maybe we could log the number of container replicas for a spec. I'm not sure we're using this ContainerReplicas count anywhere.

PR #409, opened 2020-04-27
Populate shared utilization and usage values on VCPU and VMem commodities for ContainerSpec entity

marvelyue:
This is a continuous PR for https://github.com/turbonomic/kubeturbo/pull/408.

The intent here is to 
1. implement the `allUtilizationDataAggregator.Aggregate` and `maxUtilizationDataAggregator.Aggregate` functions in `container_utilization_data_aggregator.go` to populate shared utilization data on VCPU and VMem commodities for ContainerSpec entity.
2. implement the `avgUsageDataAggregator.Aggregate` and `maxUsageDataAggregator.Aggregate` functions in `container_usage_data_aggregator.go` to populate share used, peak and capacity data on VCPU and VMem commodities for ContainerSpec entity.

**Unit Tests**:
Create new unit tests to test the 4 functions.

**Testing Done**:
By default, we'll use "max utilization data strategy" to aggregate utilization data and "average usage data strategy" to aggregate usage data.
Tested with the changes https://github.com/turbonomic/kubeturbo/pull/408, an example of discovery DTO of ContainerSpec entity is:
```json
entityDTO {
  entityType: CONTAINER_SPEC
  id: "e933da8e-fb28-11e9-8fd6-005056803aed/openvswitch"
  displayName: "openvswitch"
  commoditiesSold {
    commodityType: VCPU
    used: 51.480816042460994
    capacity: 11987.000999999998
    peak: 51.480816042460994
    active: false
    utilizationData {
      point: 0.7080966750000001
      lastPointTimestampMs: 1588003737146
      intervalMs: 0
    }
  }
  commoditiesSold {
    commodityType: VMEM
    used: 65388.0
    capacity: 1.6266504E7
    peak: 65388.0
    active: false
    utilizationData {
      point: 0.5203962014667589
      lastPointTimestampMs: 1588003754218
      intervalMs: 0
    }
  }
  monitored: false
  actionEligibility {
  }
}
```

Note that the result is based on default "max utilization data strategy" and "average usage data strategy".

By programmatically changing the strategies to using "all utilization data strategy" to aggregate utilization data and "max usage data strategy" to aggregate usage data (capacity, used and peak), an example of discovery DTO of the same ContainerSpec entity is:
```json
entityDTO {
  entityType: CONTAINER_SPEC
  id: "e933da8e-fb28-11e9-8fd6-005056803aed/openvswitch"
  displayName: "openvswitch"
  commoditiesSold {
    commodityType: VCPU
    used: 54.239080467936
    capacity: 18646.446
    peak: 54.239080467936
    active: false
    utilizationData {
      point: 0.385514225
      point: 0.39761414999999994
      point: 0.5090428
      point: 0.30904732500000004
      point: 0.36527970000000004
      point: 0.1504104857142857
      lastPointTimestampMs: 1588008629978
      intervalMs: 0
    }
  }
  commoditiesSold {
    commodityType: VMEM
    used: 84636.0
    capacity: 1.6266604E7
    peak: 84636.0
    active: false
    utilizationData {
      point: 0.3749280426847966
      point: 0.37746084181882517
      point: 0.3730100006221342
      point: 0.3813949119312181
      point: 0.3845920238466727
      point: 0.5203224282598597
      lastPointTimestampMs: 1588008629978
      intervalMs: 0
    }
  }
  monitored: false
  actionEligibility {
  }
}
```

### DIFF
--- a/pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
+++ b/pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
@@ -1,0 +1,84 @@
+package aggregation
+
+import (
+	"fmt"
+	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+	"math"
+)
+
+var (
+	avgUsageDataStrategy = "avgUsageData"
+	maxUsageDataStrategy = "maxUsageData"
+
+	DefaultContainerUsageDataAggStrategy = avgUsageDataStrategy
+
+	// Map from the configured utilization data aggregation strategy to utilization data aggregator
+	ContainerUsageDataAggregators = map[string]ContainerUsageDataAggregator{
+		avgUsageDataStrategy: &avgUsageDataAggregator{aggregationStrategy: "average usage data strategy"},
+		maxUsageDataStrategy: &maxUsageDataAggregator{aggregationStrategy: "max usage data strategy"},
+	}
+)
+
+// ContainerUsageDataAggregator interface represents a type of container usage data aggregator
+type ContainerUsageDataAggregator interface {
+	// AggregationStrategy returns aggregation strategy of this data aggregator
+	AggregationStrategy() string
+	// Aggregate aggregates commodities usage data based on the given list of commodity DTOs of a commodity type and
+	// aggregation strategy, and returns aggregated capacity, used and peak values.
+	Aggregate(containerCommodities []*proto.CommodityDTO) (float64, float64, float64, error)
+}
+
+// ---------------- Average usage data aggregation strategy ----------------
+type avgUsageDataAggregator struct {
+	aggregationStrategy string
+}
+
+func (avgUsageDataAggregator *avgUsageDataAggregator) AggregationStrategy() string {
+	return avgUsageDataAggregator.aggregationStrategy
+}
+
+func (avgUsageDataAggregator *avgUsageDataAggregator) Aggregate(commodities []*proto.CommodityDTO) (float64, float64, float64, error) {
+	if len(commodities) == 0 {
+		err := fmt.Errorf("error to aggregate commodities using %s : commodities list is empty",
+			avgUsageDataAggregator.AggregationStrategy())
+		return 0.0, 0.0, 0.0, err
+	}
+	capacitySum := 0.0
+	usedSum := 0.0
+	peakSum := 0.0
+	for _, commodity := range commodities {
+		capacitySum += *commodity.Capacity
+		usedSum += *commodity.Used
+		peakSum += *commodity.Peak
+	}
+	avgCapacity := capacitySum / float64(len(commodities))
+	avgUsed := usedSum / float64(len(commodities))
+	avgPeak := peakSum / float64(len(commodities))
+	return avgCapacity, avgUsed, avgPeak, nil
+}
+
+// ---------------- Max usage data aggregation strategy ----------------
+type maxUsageDataAggregator struct {
+	aggregationStrategy string
+}
+
+func (maxUsageDataAggregator *maxUsageDataAggregator) AggregationStrategy() string {
+	return maxUsageDataAggregator.aggregationStrategy
+}
+
+func (maxUsageDataAggregator *maxUsageDataAggregator) Aggregate(commodities []*proto.CommodityDTO) (float64, float64, float64, error) {
+	if len(commodities) == 0 {
+		err := fmt.Errorf("error to aggregate commodities using %s : commodities list is empty",
+			maxUsageDataAggregator.AggregationStrategy())
+		return 0.0, 0.0, 0.0, err
+	}
+	maxCapacity := 0.0
+	maxUsed := 0.0
+	maxPeak := 0.0
+	for _, commodity := range commodities {
+		maxCapacity = math.Max(maxCapacity, *commodity.Capacity)
+		maxUsed = math.Max(maxUsed, *commodity.Used)
+		maxPeak = math.Max(maxPeak, *commodity.Peak)
+	}
+	return maxCapacity, maxUsed, maxPeak, nil
+}
--- a/pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
+++ b/pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
@@ -6,12 +6,13 @@ import (
 	"math"
 )
 
-var (
-	avgUsageDataStrategy = "avgUsageData"
-	maxUsageDataStrategy = "maxUsageData"
-
+const (
+	avgUsageDataStrategy                 = "avgUsageData"
+	maxUsageDataStrategy                 = "maxUsageData"
 	DefaultContainerUsageDataAggStrategy = avgUsageDataStrategy
+)
 
+var (
 	// Map from the configured utilization data aggregation strategy to utilization data aggregator
 	ContainerUsageDataAggregators = map[string]ContainerUsageDataAggregator{
 		avgUsageDataStrategy: &avgUsageDataAggregator{aggregationStrategy: "average usage data strategy"},
--- a/pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
+++ b/pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
@@ -6,12 +6,13 @@ import (
 	"math"
 )
 
-var (
-	maxUtilizationDataStrategy = "maxUtilizationData"
-	allUtilizationDataStrategy = "allUtilizationData"
-
+const (
+	maxUtilizationDataStrategy                 = "maxUtilizationData"
+	allUtilizationDataStrategy                 = "allUtilizationData"
 	DefaultContainerUtilizationDataAggStrategy = allUtilizationDataStrategy
+)
 
+var (
 	// Map from the configured utilization data aggregation strategy to utilization data aggregator
 	ContainerUtilizationDataAggregators = map[string]ContainerUtilizationDataAggregator{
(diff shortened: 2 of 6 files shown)


pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+	"math"
+)
+
+var (
+	avgUsageDataStrategy = "avgUsageData"
+	maxUsageDataStrategy = "maxUsageData"
+
+	DefaultContainerUsageDataAggStrategy = avgUsageDataStrategy
```
irfanurrehman, line 13: The above three are probably constants.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
```diff
@@ -0,0 +1,84 @@
...
+	ContainerUsageDataAggregators = map[string]ContainerUsageDataAggregator{
+		avgUsageDataStrategy: &avgUsageDataAggregator{aggregationStrategy: "average usage data strategy"},
+		maxUsageDataStrategy: &maxUsageDataAggregator{aggregationStrategy: "max usage data strategy"},
+	}
+)
+
+// ContainerUsageDataAggregator interface represents a type of container usage data aggregator
+type ContainerUsageDataAggregator interface {
+	// AggregationStrategy returns aggregation strategy of this data aggregator
+	AggregationStrategy() string
```
irfanurrehman, line 25: I am not sure how you plan to use the `AggregationStrategy()`. However with the current usage of this used only in print/log statements, you can implement a `String()` method instead and you can use the structure directly and the fmt (or other loggers) can decipher if the structure implements `String()`.
Example ref: https://play.golang.org/p/uQL_RMJ_Bj.

If you need `AggregationStrategy()` for a different reason, you can use it to store and return a string enum, something like what you have created a constant above `avgUsageDataStrategy = "avgUsageData"`.
After redefining it as:
```
type DataStrategy string
const (
    avgUsageDataStrategy DataStrategy = "avgUsageData"
    ............
)```

File: pkg/discovery/worker/aggregation/container_usage_data_aggregator.go [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+	if len(commodities) == 0 {
+		err := fmt.Errorf("error to aggregate commodities using %s : commodities list is empty",
+			avgUsageDataAggregator.AggregationStrategy())
+		return 0.0, 0.0, 0.0, err
+	}
+	capacitySum := 0.0
+	usedSum := 0.0
+	peakSum := 0.0
+	for _, commodity := range commodities {
+		capacitySum += *commodity.Capacity
```
irfanurrehman, line 50: I think one or more of these values could be nil. Commodity could be created with only used value. Nil check is necessary here.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
```diff
@@ -0,0 +1,84 @@
...
+	if len(commodities) == 0 {
+		err := fmt.Errorf("error to aggregate commodities using %s : commodities list is empty",
+			maxUsageDataAggregator.AggregationStrategy())
+		return 0.0, 0.0, 0.0, err
+	}
+	maxCapacity := 0.0
+	maxUsed := 0.0
+	maxPeak := 0.0
+	for _, commodity := range commodities {
+		maxCapacity = math.Max(maxCapacity, *commodity.Capacity)
```
irfanurrehman, line 79: Same as above!

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// ContainerUtilizationDataAggregator interface represents a type of container utilization data aggregator
+type ContainerUtilizationDataAggregator interface {
+	// AggregationStrategy returns aggregation strategy of this data aggregator
+	AggregationStrategy() string
+	// Aggregate aggregates commodities utilization data based on the given list of commodity DTOs of a commodity type
+	// and aggregation strategy, and returns aggregated utilization data which contains utilization data points, last
+	// point timestamp milliseconds and interval milliseconds
+	Aggregate(commodities []*proto.CommodityDTO, lastPointTimestampMs int64) ([]float64, int64, int32, error)
```
irfanurrehman, line 29: I dont see the 3rd argument used in the code.

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+type allUtilizationDataAggregator struct {
+	aggregationStrategy string
+}
+
+func (allDataAggregator *allUtilizationDataAggregator) AggregationStrategy() string {
+	return allDataAggregator.aggregationStrategy
+}
+
+func (allDataAggregator *allUtilizationDataAggregator) Aggregate(commodities []*proto.CommodityDTO,
+	lastPointTimestampMs int64) ([]float64, int64, int32, error) {
```
irfanurrehman, line 42: lastPointTimestampMs ? Whats the `M`?. Also it does not seem to be used in either interface implementers. Is there more functionality coming?

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
irfanurrehman: And at all similar places.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
marvelyue: I plan to use the `AggregationStrategy()` only for log statements to show the aggregation strategy.

Here I'm using strategy pattern to create 2 different aggregation strategies:

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
marvelyue: Thanks. Will make them constants

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
marvelyue: The commodities here are commodities sold which are supposed to have capacity set. But you are right, commodity could be created with only used value like commodities bought. Still it makes sense and safer to add nil check here.

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
marvelyue: The 3rd returned argument means the interval milliseconds of the data points we collect every discovery cycle. For example, we collect metrics data points every 1mins for Cloud probe. For kubeturbo for now, we collect only one set of data points every discovery cycle so the 3rd returned value is always 0. I reserved the 3rd argument here in case we want to collect multiple metrics data points within a discovery cycle in the future.

I think I can only return the first argument `[]float64` in this function to avoid confusion. If we need the interval milliseconds in the future, we can set it to wherever properly.

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
marvelyue: The "Ms" means milliseconds. Sorry for the confusion. I named "lastPointTimestampMs" based on the same name from CommodityDTO. For now it's always set to current unix milliseconds. Will remove this argument to be less confusing and keep the returned []float64 only.

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+	"math"
+)
+
+var (
+	maxUtilizationDataStrategy = "maxUtilizationData"
+	allUtilizationDataStrategy = "allUtilizationData"
+
+	DefaultContainerUtilizationDataAggStrategy = maxUtilizationDataStrategy
```
dblinn, line 13: As Endre mentioned in yesterday's meeting, I think that the agreement from our earlier meetings was that "all data" would be the default strategy based on preliminary feedback. We can do some testing in our lab between the two strategies and if we find that "max data" is better we can adjust the default, but I think for now we should use the "all data" as the default.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
irfanurrehman: @marvelyue Apologies, I think my usage of word structure confused you. I understand your code.
To be more clear.
What I meant is you can do this:
Update your interface to 
```
type ContainerUsageDataAggregator interface {
	String() string
	Aggregate(containerCommodities []*proto.CommodityDTO) (float64, float64, float64, error)
}
```
Implementation of `String()` by the structure does what `AggregationStrategy()` does.
and whereever you use the same to print you can use the interface variable directly, for example:
```
usageDataAggregator, _ := agg.ContainerUsageDataAggregators[usageDataAggStrategy]

glog.Infof("ContainerSpec will aggregate Containers usage data by %s", usageDataAggregator)
```

 and the fmt (or other logger) will decipher that your interface has a stringer and use it.

A much simpler reference [here](https://tour.golang.org/methods/17)

marvelyue: Ahh, I see what you meant. Yes it makes perfect sense to use `String()` function to replace `AggregationStrategy()` function for printing logs.

Thanks for the suggestion. I'll make the changes in this following PR https://github.com/turbonomic/kubeturbo/pull/411

PR #417, opened 2020-05-04
Use milicore values in place of Mhz for cpu commodities

irfanurrehman:
It uses the newly introduced commodity `VCPU_MILICORE` as in https://github.com/turbonomic/turbo-go-sdk/pull/98.
I think the sdk pr will need to be merged before the CI here can pass.


irfanurrehman:
Closing this as this is fairly outdated and would be considerable effort to rebase. Will keep the branch for reference , if and when needed in future.
cc @enlinxu

PR #419, opened 2020-05-05
Add support for DeploymentConfig - 7.21.4

irfanurrehman:
environments.


irfanurrehman:
This is validated with server version `7.21.5` and works fine.

PR #420, opened 2020-05-05
Add support for DeploymentConfig

irfanurrehman:
environments.


irfanurrehman:
Validated against `7.22.0-SNAPSHOT` and works fine.

PR #424, opened 2020-05-24
Update affinity access commodity key to a more user friendly string

irfanurrehman:
A hash is still appended to the string to ensure its expected uniqueness.
This is a partial fix for https://vmturbo.atlassian.net/browse/OM-56506


irfanurrehman:
Rule listing for node affinity/anti-affinity
image

Rule listing for pod affinity/antiaffinity
image

irfanurrehman:
> Hi, @irfanurrehman
> 
> Have you addressed the following comments from @dblinn ?
> 
> > David Blinn
> > @irfan Rehman I have some concerns about this. I believe we save commodity keys in the database with a max length. I don't know off the top of my head what the max length is but this is a very long key and seems like we cannot guarantee an upper bound on the length. Also in terms of human understandability this is obviously much better than a hash but it's still very hard to read.
> > May 18, 2020, 12:54 PM

@ading1977 thanks for having a look. Yes I have. The code now creates a shorter readable string using `getReadableNodeSelectorTermString` and `getReadablePodAffinityTermString`. The result is shown in the screen shots attached.

irfanurrehman:
Rebased.

PR #428, opened 2020-06-09
Merged actions executor

irfanurrehman:
<s>This is mainly for reference but reviewable.
The build here will fail because the sdk change is yet to be committed.</s>

This has been tested and can be merged.

cc @pallavidn @enlinxu


irfanurrehman:
This is complete and can be reviewed. The build will pass post sdk changes merged into master.

irfanurrehman:
This merges updated changes from #435 and is verified with merged actions successfully executed.

pkg/action/executor/resize_container_util.go
```diff
@@ -97,32 +97,36 @@ func checkLimitsRequests(container *k8sapi.Container) error {
...
+		}
+		container := &(podSpec.Containers[index])
 
-	//3. update Requests
-	if spec.NewRequest != nil && len(spec.NewRequest) > 0 {
-		changed = changed || updateRequests(container, spec.NewRequest)
-	}
+		//2. update Limits
+		if spec.NewCapacity != nil && len(spec.NewCapacity) > 0 {
+			changed = updateLimits(container, spec.NewCapacity)
```
marvelyue, line 113: If the global `changed` bool is updated in such way, the final returned `changed` bool (in line 132) only represents whether request is changed for the last container in the `specs` slice.

Maybe we can introduce a temporary bool at the start of each iteration (under line 102) like
```
containerResChanged := false
```
and update this local bool here:
```
containerResChanged = containerResChanged || updateLimits(container, spec.NewCapacity)
```
and same at line 118.

In line 127, we can just check `if !containerResChanged`.

Furthermore, at the end of each iteration, update the global `changed` bool
```
changed = changed || containerResChanged
```

irfanurrehman: Nice catch @marvelyue and thanks for having a look. Fixed!
Also apologies for a late reply. I somehow missed the notification for this comment and found it today when I came here to check as to what's missing for this PR to get merged.

pkg/registration/registration_client.go
```diff
@@ -161,10 +169,35 @@ func (rClient *K8sRegistrationClient) addActionPolicy(ab *builder.ActionPolicyBu
...
+
+	actionMergeTarget2 := builder.NewActionDeDuplicateAndAggregationTargetBuilder().
+		DeDuplicatedBy(builder.NewActionAggregationTargetBuilder(proto.EntityDTO_CONTAINER_SPEC,
+			proto.ConnectedEntity_AGGREGATED_BY_CONNECTION)).
+		AggregatedBy(builder.NewActionAggregationTargetBuilder(proto.EntityDTO_WORKLOAD_CONTROLLER,
+			proto.ConnectedEntity_OWNS_CONNECTION))
+
+	containerResizeMerge := builder.NewResizeMergeSpecBuilder().
+		ForEntityType(proto.EntityDTO_CONTAINER).
+		ForCommodity(proto.CommodityDTO_VCPU).
```
enlinxu, line 189: Maybe I am missing something since I am sure you tested it, what about the request commodity?

pallavidn: This list can be extended, i only used the two commodities during development

pallavidn:
LGTM

pkg/action/action_handler.go
```diff
@@ -38,6 +38,7 @@ var (
 	turboActionContainerResize  = turboActionType{proto.ActionItemDTO_RIGHT_SIZE, proto.EntityDTO_CONTAINER}
 	turboActionMachineProvision = turboActionType{proto.ActionItemDTO_PROVISION, proto.EntityDTO_VIRTUAL_MACHINE}
 	turboActionMachineSuspend   = turboActionType{proto.ActionItemDTO_SUSPEND, proto.EntityDTO_VIRTUAL_MACHINE}
+	turboActionControllerResize = turboActionType{proto.ActionItemDTO_RIGHT_SIZE, proto.EntityDTO_WORKLOAD_CONTROLLER}
```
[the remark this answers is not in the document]
irfanurrehman: No. As far as I understand we create a container spec only if there is an associated workload controller.

pkg/registration/registration_client.go
irfanurrehman: @pallavidn, I however get the request commodity actions merged even without adding them here. I was hoping to get some insight from you.
image


I have updated the PR anyways.

irfanurrehman:
Rebased.
@enlinxu @pallavidn

PR #440, opened 2020-07-06
Storage volumes support

irfanurrehman:
Creating a new PR with the work uptil now against a branch to retrigger review of this code.

This includes the code from #389 and #403 with additional stitching support for `Azure` volumes.
I have closed those PRs in favour of this one.

In summary, this includes:

1. Discover persistent volumes and persistent volume claims from cluster.
2. Get the volume metrics from kubelet source and generate relevant metrices for consumption in kubeturbo.
3. Create a mapping between pods and volumes via pvs. This helps establish relationship between the pods PersistentVolume.
4. Include the storage amount bought commodities in the POD DTOs.
5. Build Volume entity DTOs, complete with sold commodities.
6. Include volume and related commodity information in the supply chain.
7. Include volume entity metadata info in the probe registration.
8. Stiching support for AWS volumes.
9. Stitching support for Azure volumes.

Work in progress:

10. Stitching support for vsphere discovered vmdks/volumes.

### DIFF
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -387,9 +387,7 @@ func (builder *podEntityDTOBuilder) buyCommoditiesFromVolumes(pod *api.Pod, moun
 		dtoBuilder = dtoBuilder.Provider(provider)
 
 		// Each pod mounts any given volume only once
-		var singleCommoditySlice []*proto.CommodityDTO
-		singleCommoditySlice = append(singleCommoditySlice, commBought)
-		dtoBuilder.BuysCommodities(singleCommoditySlice)
+		dtoBuilder.BuysCommodities([]*proto.CommodityDTO{commBought})
 	}
 
 	return nil
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -10,6 +10,7 @@ import (
 	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/stitching"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
+	v1 "k8s.io/api/core/v1"
 
 	sdkbuilder "github.com/turbonomic/turbo-go-sdk/pkg/builder"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
@@ -164,7 +165,7 @@ func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod, podToVolsMa
 		}
 
 		// entities' properties.
-		properties, err := builder.getPodProperties(pod)
+		properties, err := builder.getPodProperties(pod, podToVolsMap[displayName])
 		if err != nil {
 			glog.Errorf("Failed to get required pod properties: %s", err)
 			continue
@@ -395,7 +396,7 @@ func (builder *podEntityDTOBuilder) buyCommoditiesFromVolumes(pod *api.Pod, moun
 }
 
 // Get the properties of the pod. This includes property related to pod cluster property.
-func (builder *podEntityDTOBuilder) getPodProperties(pod *api.Pod) ([]*proto.EntityDTO_EntityProperty, error) {
+func (builder *podEntityDTOBuilder) getPodProperties(pod *api.Pod, vols []repository.MountedVolume) ([]*proto.EntityDTO_EntityProperty, error) {
 	var properties []*proto.EntityDTO_EntityProperty
 	// additional node cluster info property.
 	podProperties := property.BuildPodProperties(pod)
@@ -412,6 +413,24 @@ func (builder *podEntityDTOBuilder) getPodProperties(pod *api.Pod) ([]*proto.Ent
 		return nil, fmt.Errorf("failed to build EntityDTO for Pod %s: %s", podClusterID, err)
 	}
 	properties = append(properties, stitchingProperty)
+
+	if len(vols) > 0 {
+		var apiVols []*v1.PersistentVolume
+		for _, vol := range vols {
+			apiVols = append(apiVols, vol.UsedVolume)
+		}
+
+		m := stitching.NewVolumeStitchingManager()
+		err := m.ProcessVolumes(apiVols)
+		if err == nil {
+			p, err := m.BuildDTOProperty(false)
+			if err == nil {
+				properties = append(properties, p)
+			} else {
+				glog.Errorf("failed to build Volume stitching properties for Pod %s: %s", podClusterID, err)
+			}
+		}
+	}
 
 	return properties, nil
 }
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -2,12 +2,14 @@ package dtofactory
 
 import (
 	"fmt"
+
 	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 
 	api "k8s.io/api/core/v1"
 
 	"github.com/turbonomic/kubeturbo/pkg/discovery/dtofactory/property"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/stitching"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
 
@@ -65,8 +67,9 @@ func NewPodEntityDTOBuilder(sink *metrics.EntityMetricSink, stitchingManager *st
 }
 
 // Build entityDTOs based on the given pod list.
-func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]*proto.EntityDTO, error) {
+func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod, podToVolsMap map[string][]repository.MountedVolume) ([]*proto.EntityDTO, error) {
 	var result []*proto.EntityDTO
+
 	for _, pod := range pods {
 		// id.
 		podID := string(pod.UID)
@@ -154,6 +157,12 @@ func (builder *podEntityDTOBuilder) BuildEntityDTOs(pods []*api.Pod) ([]*proto.E
 			entityDTOBuilder.BuysCommodities(commoditiesBoughtQuota)
 			// pods are not movable across WorkloadController
 			entityDTOBuilder.IsMovable(proto.EntityDTO_WORKLOAD_CONTROLLER, false)
+		}
+
+		// Commodities bought from volumes
+		err = builder.buyCommoditiesFromVolumes(pod, podToVolsMap[displayName], entityDTOBuilder)
+		if err != nil {
+			return nil, err
 		}
 
 		// entities' properties.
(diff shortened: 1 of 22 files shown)


pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -349,8 +357,46 @@ func (builder *podEntityDTOBuilder) getQuotaCommoditiesBought(providerUID string
...
+
+		providerVolUID := string(mount.UsedVolume.UID)
+
+		provider := sdkbuilder.CreateProvider(proto.EntityDTO_VIRTUAL_VOLUME, providerVolUID)
+		dtoBuilder = dtoBuilder.Provider(provider)
+
+		// Each pod mounts any given volume only once
+		var singleCommoditySlice []*proto.CommodityDTO
+		singleCommoditySlice = append(singleCommoditySlice, commBought)
+		dtoBuilder.BuysCommodities(singleCommoditySlice)
```
marvelyue, line 392: nit: would be simpler to just use one line:
```
dtoBuilder.BuysCommodities([]*proto.CommodityDTO{commBought})
```

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -368,6 +414,24 @@ func (builder *podEntityDTOBuilder) getPodProperties(pod *api.Pod) ([]*proto.Ent
...
 
+	if len(vols) > 0 {
+		var apiVols []*v1.PersistentVolume
+		for _, vol := range vols {
+			apiVols = append(apiVols, vol.UsedVolume)
+		}
+
+		m := stitching.NewVolumeStitchingManager()
+		err := m.ProcessVolumes(apiVols)
+		if err == nil {
```
marvelyue, line 425: Would be helpful to log the err if err != nil

pkg/discovery/dtofactory/volume_entity_dto_builder.go
```diff
@@ -0,0 +1,154 @@
...
+		entityDto, err := entityDTOBuilder.Create()
+		if err != nil {
+			glog.Errorf("Failed to build Volume: %s entityDTO: %s", displayName, err)
+			continue
+		}
+
+		volCapacity := float32(cap)
+		isEphemaral := false
+		virtualVolumeData := &proto.EntityDTO_VirtualVolumeData{
+			StorageAmountCapacity: &volCapacity,
```
marvelyue, line 75: Looks `StorageAmountCapacity` is deprecated in `EntityDTO_VirtualVolumeData`, and the comment of this field in `CommonDTO.pb.go` says:
> The field is deprecated because the value is now carried in the sold commodity

If  I understand correctly, we may not need to set volCapacity here, instead the StorageAmount commodity sold handles this capacity.

pkg/discovery/k8s_discovery_client.go
```diff
@@ -280,6 +281,15 @@ func (dc *K8sDiscoveryClient) discoverWithNewFramework(targetID string) ([]*prot
...
 	}
 
+	// 4. build entityDTOs for volumes
+	volumeEntityDTOBuilder := dtofactory.NewVolumeEntityDTOBuilder(podVolumeMetrics)
+	volumeEntityDTOs, err := volumeEntityDTOBuilder.BuildEntityDTOs(clusterSummary.VolumeToPodsMap)
+	if err != nil {
+		glog.Errorf("Error while creating volume entityDTOs: %v", err)
+	}
+	glog.V(3).Infof("There are %d Storage Volume entityDTOs.", len(volumeEntityDTOs))
+	entityDTOs = append(entityDTOs, volumeEntityDTOs...)
```
marvelyue, line 291: nit: V(2) looks better as this is an important message which can be shown by default and would be good to put these under else block (err == nil)

pkg/discovery/processor/cluster_processor_test.go
```diff
@@ -343,13 +345,28 @@ func (s *MockClusterScrapper) GetKubernetesServiceID() (string, error) {
...
 		return s.mockGetAllServices()
 	}
 	return nil, fmt.Errorf("GetAllServices Not implemented")
 }
 
+func (s *MockClusterScrapper) GetAllPVs() ([]*v1.PersistentVolume, error) {
+	if s.mockGetAllPVs != nil {
+		return s.mockGetAllPVs()
+	}
+	return nil, fmt.Errorf("GetAllServices Not implemented")
```
marvelyue, line 360: GetAllServices -> GetPVs

pkg/discovery/processor/cluster_processor_test.go
```diff
@@ -343,13 +345,28 @@ func (s *MockClusterScrapper) GetKubernetesServiceID() (string, error) {
...
+}
+
+func (s *MockClusterScrapper) GetAllPVCs() ([]*v1.PersistentVolumeClaim, error) {
+	if s.mockGetAllPVCs != nil {
+		return s.mockGetAllPVCs()
+	}
+	return nil, fmt.Errorf("GetAllServices Not implemented")
```
marvelyue, line 367: GetAllServices -> GetAllPVCs

pkg/discovery/processor/volume_processor.go
```diff
@@ -0,0 +1,115 @@
...
+	// TODO (irfanurrehman): We might need the volume name to map the metrics generated from pod
+	// and gathered via kubelet monitor to the individual volumes.
+	// TODO (irfanurrehman): (Next Step) To find out how are metrics generated for volumes which
+	// are used by multiple pods.
+	volumeToPodsMap := make(map[*v1.PersistentVolume][]repository.PodVolume)
+	for pv, pvc := range volumeToClaimMap {
+		if pvc == nil {
+			// Unused volume.
+			volumeToPodsMap[pv] = nil
+		}
```
marvelyue, line 74: Do you want to add a "continue" here to skip the whole iteration below when pvc == nil? And then in line 79 no need to check `if pvc != nil`

pkg/discovery/worker/metrics_collector.go
```diff
@@ -189,6 +190,55 @@ func collectContainersComputeResources(pod *v1.Pod) (float64, float64, float64,
...
+
+	var podToVolsMap map[string][]repository.MountedVolume
+	if collector.Cluster.PodToVolumesMap != nil {
+		podToVolsMap = collector.Cluster.PodToVolumesMap
+	}
+
+	metricsSink := collector.MetricsSink
+	//Iterate over all the pods in the collection
+	for _, pod := range collector.PodList {
+		podKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
```
marvelyue, line 204: There's a func `util.PodKeyFunc(pod)` which is equivalent to the logic here.

pkg/discovery/processor/volume_processor.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, pod := range allPods {
+			for _, vol := range pod.Spec.Volumes {
+				claim := vol.VolumeSource.PersistentVolumeClaim
+				if claim != nil {
+					if pvc != nil {
+						if claim.ClaimName == pvc.Name {
+							pVol := repository.PodVolume{
+								QualifiedPodName: fmt.Sprintf("%s/%s", pod.Namespace, pod.Name),
```
marvelyue, line 82: There's a func `util.PodKeyFunc(pod)` which is equivalent to the logic here.

pkg/registration/supply_chain_factory.go
```diff
@@ -220,6 +236,7 @@ func (f *SupplyChainFactory) buildNodeMergedEntityMetadata() (*proto.MergedEntit
 		PatchSoldMetadata(proto.CommodityDTO_VMEM_REQUEST_QUOTA, fieldsUsedCapacity).
 		PatchSoldMetadata(proto.CommodityDTO_NUMBER_CONSUMERS, fieldsUsedCapacity).
 		PatchSoldMetadata(proto.CommodityDTO_VSTORAGE, fieldsUsedCapacity).
+		PatchSoldMetadata(proto.CommodityDTO_STORAGE_AMOUNT, fieldsUsedCapacity).
```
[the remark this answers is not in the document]
irfanurrehman: I think this was a mistake. Updated!

pkg/registration/registration_client.go
```diff
@@ -149,6 +149,16 @@ func (rClient *K8sRegistrationClient) GetActionPolicy() []*proto.ActionPolicyDTO
 
 	rClient.addActionPolicy(ab, node, nodePolicy)
 
+	// 6. volumes
+	volume := proto.EntityDTO_VIRTUAL_VOLUME
```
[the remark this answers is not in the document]
irfanurrehman: I had a [document](https://vmturbo.sharepoint.com/:p:/s/CloudNative/ESSoEJbPAedCr1ioE_2QLv4BOKW9M-QY6IqvhFmPfsTTwQ?e=EqjJZ1) that I did create before starting implementation of this. Its a little tricky as to which probe should report policy and which probe should execute it. For example, please check slide 6 in the referred presentation. An action (for example size change) will need updates done via both probes. An action like a property change can be carried out only via the IaaS probe, but if any of those properties reflect in the pv spec/status then that needs to be updated by kubeturbo too.

pkg/discovery/worker/metrics_collector.go
```diff
@@ -189,6 +190,55 @@ func collectContainersComputeResources(pod *v1.Pod) (float64, float64, float64,
 	return totalCPULimits, totalCPURequests, totalMemLimits, totalMemRequests
 }
 
+func (collector *MetricsCollector) CollectPodVolumeMetrics() []*repository.PodVolumeMetrics {
+	var podVolumeMetricsCollection []*repository.PodVolumeMetrics
+
+	var podToVolsMap map[string][]repository.MountedVolume
+	if collector.Cluster.PodToVolumesMap != nil {
+		podToVolsMap = collector.Cluster.PodToVolumesMap
+	}
```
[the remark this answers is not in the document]
irfanurrehman: Yes, makes perfect sense. Updated!

pkg/discovery/worker/metrics_collector.go
```diff
@@ -189,6 +190,55 @@ func collectContainersComputeResources(pod *v1.Pod) (float64, float64, float64,
...
+
+			metricUID = metrics.GenerateEntityResourceMetricUID(metrics.PodType, volKey,
+				metrics.StorageAmount, metrics.Capacity)
+			metric, _ = metricsSink.GetMetric(metricUID)
+			if metric != nil && metric.GetValue() != nil {
+				podVolumeMetric.Capacity = metric.GetValue().(float64)
+				found = true
+			}
+
+			if found == true {
```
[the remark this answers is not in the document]
irfanurrehman: I wasn't (and still am not) aware of the requirement of the capacity being set. I am not sure what fails if a probe reports a commodity without capacity (and I think this can happen in kubeturbo too).
On the other hand I did write this code to handle cases where one or the other (for some reason) is not available. This is quite unlikely though, as the kubelet api, if it reports should ideally report both and thats the only point from where I might get one and not the other.

pkg/discovery/dtofactory/pod_entity_dto_builder.go
irfanurrehman: Nice catch. Updated!

pkg/discovery/dtofactory/pod_entity_dto_builder.go
irfanurrehman: Actually `ProcessVolumes` is logging the error if one is encountered internally. I am using the error returned from there only as a flag and retaining it as type error for any future use.

pkg/discovery/dtofactory/volume_entity_dto_builder.go
irfanurrehman: I am already setting the capacity in the sold commodity. I can't remember it completely now, but I think I did encounter some issue on an older server if this was not set and thats probably the reason I did introduce it. Is there any harm in leaving it as is, even if the value is unused by the server?

pkg/discovery/k8s_discovery_client.go
irfanurrehman: Agree, Updated!

pkg/discovery/processor/cluster_processor_test.go
irfanurrehman: Updated!

pkg/discovery/processor/cluster_processor_test.go
irfanurrehman: Updated!

pkg/discovery/processor/volume_processor.go
irfanurrehman: This is intentionally left this way to remember and handle the TODO in the else case in the loop below.
`// TODO (irfanurrehman): check other volume sources categorically.`

pkg/discovery/worker/metrics_collector.go
irfanurrehman: Updated!

pkg/discovery/processor/volume_processor.go
irfanurrehman: Updated!

irfanurrehman:
@ading1977 @marvelyue updated to fix code as per your comments. PTAL!

pkg/discovery/dtofactory/pod_entity_dto_builder.go
marvelyue: This makes sense. Sorry for missing the errors in `ProcessVolumes` func.

pkg/discovery/dtofactory/volume_entity_dto_builder.go
marvelyue: Actually I don't think there's any harm with this (it may be fine every from memory perspective). I was just wondering if we could avoid some redundancies since it's marked as deprecated. Still like you said, it could cause some issue on an older server. So looks It would be safer to keep this

pkg/discovery/processor/volume_processor.go
marvelyue: I see. Make sense

pkg/discovery/worker/metrics_collector.go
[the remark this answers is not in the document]
marvelyue: I remember there was an issue where a commodity capacity is not set and an exception occurred inside market analysis like this one (https://vmturbo.atlassian.net/browse/OM-59514). The fix Endre eventually put there was to set a large default capacity. Would be good to set a default large capacity just in case, although ideally kubelet should be able to return both used and capacity.

irfanurrehman:
@ading1977 please let me know if you have any additional comments.

PR #441, opened 2020-07-06
Operator managed resources control

marvelyue:
**Intent**:
Implement a new mechanism called OperatorResourceMapping to provide the capability to execute actions on the resources controlled by Operator in Kubeturbo.

**Background**:
When user configures Operator to manage the k8s application lifecycle, the Operator uses the declarative approach to manage the desired state of the application. In that case, any Turbonomic actions to manage the size of the deployment (horizontal and vertical) will be reverted by the operator.

This is applicable to manage our own XL as well. The baseline of this mechanism is to manage our own Xl cluster by automating Turbo actions.

**Design**:
Link to the slide deck for the detailed design: https://vmturbo-my.sharepoint.com/:p:/p/yue_wu/Eeg6SAysEDtFmdpIxDW6UVYBz3h9iFIBZBO_k-kNQUxilw?e=9IRl1T

* The basic idea of the ORM (OperatorResourceMapping) CR is to specify the resource mappings from the path of a K8s controller (like Deployment) to the path of the CustomResource (CR) managed by an Operator so that values can be updated into Operator managed CR instead of Deployment when executing actions.​
* One ORM CR is mapped to one CRD managed by an Operator.
* ORM CR is created per namespace to manage the operator resource mapping within a certain namespace.
* ORM CR focuses only on action execution in Kubeturbo. The supported actions include:
  * Container resizing in a K8s controller 
  * Pod provision/suspend in a K8s controller​
* Supported K8s controllers: (which match the existing action execution logic)​
  * Deployment​
  * ReplicaSet​
  * DeploymentConfig​
  * ReplicationController

A snippet of ORM CR for our Xl CRD is as follows:
```
apiVersion: turbonomic.com/v1alpha1 
kind: OperatorResourceMapping 
metadata: 
  name: xls.charts.helm.k8s.io 
spec: 
  resourceMappings: 
    - srcResourceSpec: 
        kind: Deployment 
        componentNames: 
          - action-orchestrator 
          - api 
          - auth 
          ... 
      resourceMappingTemplates:
        - srcPath: .spec.template.spec.containers[?(@.name=="{{.componentName}}")].resources
          destPath: .spec.{{.componentName}}.resources
        - srcPath: .spec.replicas
          destPath: .spec.{{.componentName}}.replicaCount
```
The `srcPath` and `destPath` are in [JSONPath](https://goessner.net/articles/JsonPath/) format.

ORM CRD is [turbo-operator-resource-mapping-crd.yaml](https://vmturbo-my.sharepoint.com/:u:/p/yue_wu/EWC3PcU-6eVBoib17oRzh4MBHazN6k3tU9FLTW4ea2_e8g?e=ybmEDL)
And the complete version of ORM CR based on our Xl CRD is [turbo-operator-resource-mapping-cr.yaml](https://vmturbo-my.sharepoint.com/personal/yue_wu_vmturbo_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fyue%5Fwu%5Fvmturbo%5Fcom%2FDocuments%2Fturbo%2Doperator%2Dresource%2Dmapping%2Dcr%2Eyaml&parent=%2Fpersonal%2Fyue%5Fwu%5Fvmturbo%5Fcom%2FDocuments&originalPath=aHR0cHM6Ly92bXR1cmJvLW15LnNoYXJlcG9pbnQuY29tLzp1Oi9wL3l1ZV93dS9FUkZodVYwdkkzaElqZUNDcWlxTjZJTUJmYjZFYXY2MWhiX0ZIdUdxRG8tRTRRP3J0aW1lPVAza21FT1FoMkVn)

**Implementation**:
1. Create an ORMClient which builds operator resource mapping templates fetched from OperatorResourceMapping CR in discovery client and provides the capability to update the corresponding CR for an Operator managed resource in action execution client.
2. In ORMClient, Discover all ORM CRs based on ORM CRD, where ORM CR name is the name of an Operator managed CR.
3. In ORMClient, Discover all Operator managed CRs from CRD name (ORM CR name)
4. In discovery client, store cached map data (`operatorReourceSpecMap`) in discovery client updated in each discovery. 
3.1 The map is from Operator managed CR UID to `ORMSpec` object. 
3.2 `ORMSpec` defines the spec data to update corresponding CR for an Operator managed resource in action execution client.
An example of `operatorResourceSpecMap` is:
```
{
  "b4ce6060-93be-11ea-9406-005056b83d00": {
    operatorGVResource  schema.GroupVersionResource
    "Deployment/group": {
      "componentName": "group",
      "resourceMappingTemplates":
        [
         {
           "srcPath": `.spec.template.spec.containers[?(@.name=="{{.componentName}}")].resources`
           "destPath": ".spec.{{.componentName}}.resources"
         },
        ]
    }
  }
}
```
4. In action execution client, when updating in `K8s_controller`, if ownerReference of a K8s controller is not nil, call ORMClient to do the update based on resource mapping instead of directly update K8s controller.
4.1 If an action is to change a field in K8s controller which is defined in ORM CR resourceMappingTemplate, update the CR accordingly
4.2 If nothing is updated to Operator CR, it means the action is trying to change a field in K8s controller which is NOT defined in ORM CR resourceMappingTemplate (meaning the change is not allowed), returns an error and make the action execution failed.

**Testing Done**:
Tested in a cluster with multi-tenancy (2 Xl are deployed in 2 different namespaces). Actions in 2 different namespaces can be successfully executed with following log messages:

Action1 in namespace `turbonomic`
```
I0713 14:32:15.367251       1 remote_mediation_client.go:215] [handleServerMessages][ServerRequestEndpoint] : received message with request type Action.
I0713 14:32:15.367367       1 action_handler.go:332] Received an action RIGHT_SIZE for entity CONTAINER [turbonomic/kafka-779d8f8589-b7ffp/kafka]
I0713 14:32:15.432850       1 resize_container_util.go:181] Begin to consistently resize Deployment of pod turbonomic/kafka-779d8f8589-b7ffp.
I0713 14:32:15.450312       1 resize_container_util.go:69] Try to update container kafka resource limit from map[memory:{i:{value:4294967296 scale:0} d:{Dec:<nil>} s:4Gi Format:BinarySI}] to map[memory:{{872415232 0} {<nil>}  BinarySI}]
I0713 14:32:15.491486       1 operator_resource_mapping.go:257] Updating CR Xl/xl-example for Deployment/kafka in namespace turbonomic
I0713 14:32:15.629329       1 operator_resource_mapping.go:310] Successfully updated CR Xl/xl-example '/spec/kafka/resources' from map[limits:map[memory:4Gi] requests:map[memory:768Mi]] to map[limits:map[memory:832Mi] requests:map[memory:768Mi]] for Deployment/kafka in namespace turbonomic
I0713 14:32:15.629413       1 k8s_controller_updater.go:117] Successfully updated Deployment of pod turbonomic/kafka-779d8f8589-b7ffp
```
Action2 in namespace `turbonomic2`:
```
I0713 14:32:22.404560       1 remote_mediation_client.go:215] [handleServerMessages][ServerRequestEndpoint] : received message with request type Action.
I0713 14:32:22.404634       1 action_handler.go:332] Received an action RIGHT_SIZE for entity CONTAINER [turbonomic2/kafka-5d46579f45-qhhkc/kafka]
I0713 14:32:22.432283       1 resize_container_util.go:181] Begin to consistently resize Deployment of pod turbonomic2/kafka-5d46579f45-qhhkc.
I0713 14:32:22.440663       1 resize_container_util.go:69] Try to update container kafka resource limit from map[memory:{i:{value:4294967296 scale:0} d:{Dec:<nil>} s:4Gi Format:BinarySI}] to map[memory:{{805306368 0} {<nil>}  BinarySI}]
I0713 14:32:22.464810       1 operator_resource_mapping.go:257] Updating CR Xl/xl-test for Deployment/kafka in namespace turbonomic2
I0713 14:32:22.523136       1 operator_resource_mapping.go:310] Successfully updated CR Xl/xl-test '/spec/kafka/resources' from map[limits:map[memory:4Gi] requests:map[memory:768Mi]] to map[limits:map[memory:768Mi] requests:map[memory:768Mi]] for Deployment/kafka in namespace turbonomic2
I0713 14:32:22.523258       1 k8s_controller_updater.go:117] Successfully updated Deployment of pod turbonomic2/kafka-5d46579f45-qhhkc
```

Furthermore tested on Endre's saas control instance and actions were executed successfully.


pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	return crs.Items, groupVersionRes, nil
+}
+
+func (ormClient *ORMClient) populateORMTemplateMap(ormCR unstructured.Unstructured) (map[string]ORMTemplate, error) {
+	ormCRName := ormCR.GetName()
+	ormTemplateMap := make(map[string]ORMTemplate)
+	resourceMappings, _, err := unstructured.NestedSlice(ormCR.Object, ormResourceMappingsPath...)
+	if err != nil {
+		return nil, fmt.Errorf("no data under '%s' in ORM CR %s: %v", util.JSONPath(ormResourceMappingsPath), ormCRName, err)
```
irfanurrehman, line 166: This should be based on the `bool` value return from the `NestedSlice` call. The flag tells if the field was found and `err` is generic error (which rarely happens on the methods on unstructured). 
Effectively if a field is not found, err would mostly be still nil.

pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+	for _, resourceMapping := range resourceMappings {
+		rm, ok := resourceMapping.(map[string]interface{})
+		if !ok {
+			glog.Errorf("ResourceMapping does not have the expected 'map[string]interface{}' structure in ORM CR %s", ormCRName)
```
irfanurrehman, line 171: This is unlikely, but as you are handling this case, better to use the index to give some kind of linking into the actual CR if the error does happen.

pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+			glog.Errorf("Value not found under '%s' under '%s' in ORM CR %s", util.JSONPath(resourceMappingTemplatesPath),
+				util.JSONPath(ormResourceMappingsPath), ormCRName)
+			continue
+		}
+
+		rmTemplates := make([]map[string]interface{}, 0, len(resourceMappingTemplates))
+		for _, resourceMappingTemplate := range resourceMappingTemplates {
+			rmTemplate, ok := resourceMappingTemplate.(map[string]interface{})
+			if !ok {
+				glog.Errorf("resourceMappingTemplate does not have the expected 'map[string]interface{}' structure in ORM CR %s", ormCRName)
```
irfanurrehman, line 197: Same as above, use the index. In this case it would be 2 indexes printed.

pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+		Resource: crd.Spec.Names.Plural,
+	}
+	crs, err := ormClient.dynClient.Resource(groupVersionRes).Namespace(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return nil, groupVersionRes, err
+	}
+	return crs.Items, groupVersionRes, nil
+}
+
+func (ormClient *ORMClient) populateORMTemplateMap(ormCR unstructured.Unstructured) (map[string]ORMTemplate, error) {
```
irfanurrehman, line 161: Its allright that you have manually curated the crd definition and read the definition into a code structure via this function for now. A more sustainable mechanism would be to use a project like [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) and bootstrap the crd type definition (ORM in our case) as golang structs.  
There are many benefits we would have with that. We can have typed clients (typed clients have all the type checking benefits) to deal with our API spec (ORM in our case) or directly use encode/decode helpers from unstructured libraries if we stick to dynamic client. Additionally any future changes could mean simply updating the golang struct definition and autogenerate the CRD spec, which can otherwise be daunting to manage manually.
Also, I remember that I did mention this to you earlier, but I must apologise to not have stressed its usage enough. I now feel that it should have made your work easier and certainly makes future maintainance easier.

irfanurrehman:
@marvelyue apologies for getting delayed in checking this. Had a look at the CR yaml that you published [here](https://vmturbo-my.sharepoint.com/personal/yue_wu_vmturbo_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fyue%5Fwu%5Fvmturbo%5Fcom%2FDocuments%2Fturbo%2Doperator%2Dresource%2Dmapping%2Dcr%2Eyaml&parent=%2Fpersonal%2Fyue%5Fwu%5Fvmturbo%5Fcom%2FDocuments&originalPath=aHR0cHM6Ly92bXR1cmJvLW15LnNoYXJlcG9pbnQuY29tLzp1Oi9wL3l1ZV93dS9FUkZodVYwdkkzaElqZUNDcWlxTjZJTUJmYjZFYXY2MWhiX0ZIdUdxRG8tRTRRP3J0aW1lPUVRNk9lUjRxMkVn). Is that the final CR definition. I am confused with `srcResourceSpec` under `resourceMappings`. I thought `resourceMappings` is a slice of object type `resourceMapping`.

Also where are you planning to maintain the CR definition and a CR sample/example for consumption. Should that also be part of this PR and maintained in this repo? Please have a look at [this comment](https://github.com/turbonomic/kubeturbo/pull/441/files#r456301664) and if at a point either this becomes a thing or we update our implementation we follow the `kubebuilder` way to do this. cc @enlinxu.

pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+		return fmt.Errorf("failed to get CR %s for %s in namespace %s: %v", operatorRes, componentKey, resourceNamespace, err)
+	}
+
+	glog.V(2).Infof("Updating CR %s for %s in namespace %s", operatorRes, componentKey, resourceNamespace)
+	updated := false
+	// Update based on each resourceMappingTemplate
+	for _, resourceMappingTemplate := range ormTemplate.resourceMappingTemplates {
+		// Set resourceMappingComponentName to resourceMappingTemplate so as to parse the srcPath and destPath based on text template
+		resourceMappingTemplate[resourceMappingComponentName] = ormTemplate.componentName
+		srcPath, destPath, err := ormClient.parseSrcAndDestPath(resourceMappingTemplate)
```
irfanurrehman, line 250: Would this work if the component name is skipped and the `src` and `dst` paths are actual paths. For example to control an operator which has only 1 component. Or alternatively for another CR writer (user) who specifies the paths in which nothing needs to be replaced.

pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+		if !needsUpdate {
+			glog.V(4).Infof("No changes in controller %s/%s src path '%s'. Skip updating Operator CR %s.",
+				updatedControllerObj.GetKind(), updatedControllerObj.GetName(), srcPath, operatorRes)
+			continue
+		}
+		fields := strings.Split(destPath, ".")
+		if len(fields) < 2 {
+			return fmt.Errorf("failed to update %v to CR %s for %s in namespace %s: '%s' is invalid path", newValue, operatorRes, componentKey, resourceNamespace, destPath)
+		}
+		origCRValue, _, err := util.NestedField(operatorCR, resourceMappingDestPath, destPath)
```
irfanurrehman, line 271: I could not understand the need of this function wrt to the `unstructured.NestedField()`. What am I missing?

pkg/resourcemapping/operator_resource_mapping.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		_, err = dynResourceClient.Update(operatorCR, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to update %v to CR %s '%s' for %s in namespace %s: %v", newValue, operatorRes, destPath, componentKey, resourceNamespace, err)
+		}
+		updated = true
+		glog.Infof("Successfully updated CR %s '%s' from %v to %v for %s in namespace %s", operatorRes, destPath, origCRValue, newValue, componentKey, resourceNamespace)
+	}
+	if !updated {
+		return fmt.Errorf("failed to update CR %s for %s in namespace %s: missing resource mapping template", operatorRes, componentKey, resourceNamespace)
```
irfanurrehman, line 287: This doesn't seem right. `updated` will remain false if all `resourceMappingTemplates` result in `!needsUpdate`. That seems like a valid case and not an error.

pkg/resourcemapping/operator_resource_mapping.go
irfanurrehman: Also, please validate the bool return.

pkg/resourcemapping/operator_resource_mapping.go
marvelyue: Thanks. Will use the index in the error message

marvelyue:
> @marvelyue apologies for getting delayed in checking this. Had a look at the CR yaml that you published [here](https://vmturbo-my.sharepoint.com/personal/yue_wu_vmturbo_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fyue%5Fwu%5Fvmturbo%5Fcom%2FDocuments%2Fturbo%2Doperator%2Dresource%2Dmapping%2Dcr%2Eyaml&parent=%2Fpersonal%2Fyue%5Fwu%5Fvmturbo%5Fcom%2FDocuments&originalPath=aHR0cHM6Ly92bXR1cmJvLW15LnNoYXJlcG9pbnQuY29tLzp1Oi9wL3l1ZV93dS9FUkZodVYwdkkzaElqZUNDcWlxTjZJTUJmYjZFYXY2MWhiX0ZIdUdxRG8tRTRRP3J0aW1lPUVRNk9lUjRxMkVn). Is that the final CR definition? I am confused with `srcResourceSpec` under `resourceMappings`. I thought `resourceMappings` is a slice of object type `resourceMapping`.
> 
> Also where are you planning to maintain the CR definition and a CR sample/example for consumption. Should that also be part of this PR and maintained in this repo? Please have a look at [this comment](https://github.com/turbonomic/kubeturbo/pull/441/files#r456301664) and if at a point either this becomes a thing or we update our implementation its better we follow the `kubebuilder` way to do this. cc @enlinxu.

Hi @irfanurrehman, thanks a lot for your review and comments!

For the CR, yes that's the latest version. Sorry for the confusion. I didn't explicitly use a `resourceMapping` object type under `resourceMappings`, instead, each element in `resourceMappings` slice is a map where the keys are "srcResourceSpec" and "resourceMappingTemplates":
```yaml
- srcResourceSpec:
    kind: Deployment
    componentNames:
      - action-orchestrator
      - api
      ...
  resourceMappingTemplates:
    - srcPath: .spec.template.spec.containers[?(@.name=="{{.componentName}}")].resources
      destPath: .spec.{{.componentName}}.resources
```

Or are your suggesting using a `resourceMapping` object? Something like:
```yaml
- resourceMapping:
     srcResourceSpec:
       kind: Deployment
       componentNames:
         - action-orchestrator
         - api
         ...
     resourceMappingTemplates:
       - srcPath: .spec.template.spec.containers[?(@.name=="{{.componentName}}")].resources
         destPath: .spec.{{.componentName}}.resources
```

And sorry, I forgot to attach the link to the CRD, which is [turbo-operator-resource-mapping-crd.yaml](https://vmturbo-my.sharepoint.com/:u:/p/yue_wu/EWC3PcU-6eVBoib17oRzh4MBHazN6k3tU9FLTW4ea2_e8g?e=ybmEDL) (I added to the description also)

My thought is to deploy the ORM CRD and CR during Xl installation. I planed to maintain the ORM CRD and CR example/sample in Xl project repo (like the ones under <xl_project>/build/kubernetes/operator/deploy/crds/) once kubeturbo implementation is finalized. I think need to sync up with @enlinxu on where (which repo) we want to maintain the ORM CRD and CR sample.

pkg/resourcemapping/operator_resource_mapping.go
marvelyue: Thanks @irfanurrehman ! Yes, I remember you mentioned the "kubebuilder". And sorry I didn't fully investigate and understand the usage. When taking an initial glance at that project, I thought it might complicate the process since it provides a whole set of customization to create our own controller and resources, where most of stuff we don't need. After playing with the "Hello Word". I agree that it'll actually be more sustainable to use this `kubebuilder`. 

If I understand correctly, we would need a new project using `kubebuilder` to maintain the ORM type definition and provide go struct with ORM specs. As you said, we can not only make use of the typed clients (or directly use `unstructured` libraries), but also autogenerate the CRD spec from the golang struct. Furthermore the repo of this project could be the place where we maintain the ORM CRD and CR sample/example.

If it's ok, I'll take this as a further improvement to current logic.

pkg/resourcemapping/operator_resource_mapping.go
marvelyue: Good point. Yes, this will still work if nothing needs to be replaced in the `srcPath` and `destPath`, for example:
```
.spec.template.spec.container.0.resources
```
or
```
.spec.resources
```
I'll add a new unit test for this case

pkg/resourcemapping/operator_resource_mapping.go
marvelyue: When calling func `unstructured.NestedFieldCopy(obj, fields ...string)`, the input `fields` can only be a certain key of a map but doesn't support filtering by a certain field like `containers[?(@.name=="api")]`. So in order to find the value of such path like `.spec.template.spec.containers[?(@.name=="api")].resources`, I introduced this new `util.NestedField` func.

Again it's valid case either `bool` as true (found the value to be replaced) or false (not found value so needs to be added). And the same logic below will handle both cases

pkg/resourcemapping/operator_resource_mapping.go
marvelyue: Good catch! So when the `operator_resource_mapping.Update` func is triggered, an action is executed from turbo server to make changes on the corresponding resource and there must be some changes (otherwise in `k8s_controller_updater` `update` func, `updated` bool would be true and the func would directly return)

Then if `needsUpdate` here is false, it means there are some changes turbo server wants to make but not defined in the ORM resource mapping templates, for example the fixed field defined in Operator CR but not allowed to be changed. I'm making use of the logic here to throw an error notification and make the action failed because nothing gets changed with this executed action. I'll add a comment here

Ideally, like we discussed before, it'll be helpful if we have a mechanism to avoid market analysis from generated such actions.

pkg/resourcemapping/operator_resource_mapping.go
irfanurrehman: As discussed offline, lets keep this item in the backlog and revisit this when we have real customer requirements on this and are sure that we will need to maintain this for long.

pkg/resourcemapping/operator_resource_mapping.go
irfanurrehman: Ack!

pkg/resourcemapping/operator_resource_mapping.go
irfanurrehman: Awsome!

pkg/resourcemapping/operator_resource_mapping.go
irfanurrehman: Ack!

pkg/resourcemapping/operator_resource_mapping.go
irfanurrehman: Its noway anybody looking at this code would know your original intent.. :)
Thanks for adding the comment.. !

irfanurrehman:
> Or are your suggesting using a `resourceMapping` object? Something like:
> 
> ```yaml
> - resourceMapping:
>      srcResourceSpec:
>        kind: Deployment
>        componentNames:
>          - action-orchestrator
>          - api
>          ...
>      resourceMappingTemplates:
>        - srcPath: .spec.template.spec.containers[?(@.name=="{{.componentName}}")].resources
>          destPath: .spec.{{.componentName}}.resources
> ```

yes.

PR #444, opened 2020-07-09
Fix affinity processor error

irfanurrehman:
Fix for observed error msg while processing affinities
Fixes [OM-60545](https://vmturbo.atlassian.net/browse/OM-60545)    
    
This happens because the affinity processor considers all pods wrt only the ready and running pods on which the original pod entityDTOs are created. There is no functionality impact except with or without the fix.

### DIFF
--- a/pkg/cluster/cluster_info_scraper.go
+++ b/pkg/cluster/cluster_info_scraper.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+
 	"k8s.io/client-go/dynamic"
 
 	api "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/cluster/cluster_info_scraper.go
```diff
@@ -205,13 +206,32 @@ func (s *ClusterScraper) GetRunningAndReadyPodsOnNodes(nodeList []*api.Node) []*
...
 	fieldSelector, err := fields.ParseSelector("spec.nodeName=" + nodeName + ",status.phase=" +
 		string(api.PodRunning))
 	if err != nil {
 		return nil, err
 	}
+
+	return s.findSelectedPods(fieldSelector)
+}
+
+func (s *ClusterScraper) findSelectedPods(fieldSelector fields.Selector) ([]*api.Pod, error) {
```
ading1977, line 234: Can you use the existing `GetPods()` function?

irfanurrehman: Updated!

irfanurrehman:
@ading1977 thanks for having a look. Updated as per your comment. PTAL!

PR #446, opened 2020-07-14
Make HA nodes non suspendable

irfanurrehman:
This PR updates the node role groups as below:
1. Add suspendable = false for entities (VMs) that are part of the HA Node consumer group.
2. All node role groups have a common naming.
3. There is no inbuilt check for `master` as a node label to make the node non suspendable. This is replaced by 1 above.

### DIFF
--- a/pkg/discovery/dtofactory/node_roles_group_dto_builder.go
+++ b/pkg/discovery/dtofactory/node_roles_group_dto_builder.go
@@ -37,8 +37,10 @@ func (builder *NodeRolesGroupDTOBuilder) Build() []*proto.GroupDTO {
 		var err error
 		roleName := role.roleName
 		if role.isHA {
-			groupID := fmt.Sprintf("NodeRole[HA]-%s-%s", roleName, builder.targetId)
-			displayName := fmt.Sprintf("NodeRole[HA]::%s [%s]", roleName, builder.targetId)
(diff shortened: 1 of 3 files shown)


irfanurrehman:
cc @enlinxu @evat-pm

irfanurrehman:
Groups created with common naming scheme:
image

Policy name still carries the string `HA` (else `DRS Segmentation` can cause confusions):
image

pkg/discovery/dtofactory/node_roles_group_dto_builder.go
```diff
@@ -37,8 +37,10 @@ func (builder *NodeRolesGroupDTOBuilder) Build() []*proto.GroupDTO {
 		var err error
 		roleName := role.roleName
 		if role.isHA {
-			groupID := fmt.Sprintf("NodeRole[HA]-%s-%s", roleName, builder.targetId)
-			displayName := fmt.Sprintf("NodeRole[HA]::%s [%s]", roleName, builder.targetId)
+			groupID := fmt.Sprintf("NodeRole-%s-%s", roleName, builder.targetId)
+			displayName := fmt.Sprintf("[HA]NodeRole::%s [%s]", roleName, builder.targetId)
```
enlinxu, line 41: I don't think we should have [HA] here based on our discussion

PR #449, opened 2020-07-20
Add VM stitching support for tanzu

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-60758
Verified the stitching working fine on:
- `dc11` with `ae-cluster2`
- `azure` with `aks-virtual`
Should ideally work allright on aws and gce too.
cc @enlinxu @esara


pkg/discovery/stitching/node_uuid.go
```diff
@@ -203,6 +205,50 @@ func (gce *gceNodeUUIDGetter) GetUUID(node *api.Node) (string, error) {
...
+
+	//1. Get the suffix, this is one of our ids:
+	//   4200c244-1e27-8473-ab44-999195de924c
+	stitchingID := providerId[len(vspherePrefix):]
+
+	//2. Check if node.Status.NodeInfo.SystemUUID is filled by provider
+	//   If it is, append that as another id.
+	//   4200c244-1e27-8473-ab44-999195de924c,4200c244-1e27-8473-ab44-999195de924c
+	if node.Status.NodeInfo.SystemUUID != "" {
+		stitchingID = fmt.Sprintf("%s,%s", stitchingID, node.Status.NodeInfo.SystemUUID)
```
chlam4, line 246: A minor suggestion: maybe we can abstract out the delimiter to a constant definition and share between here and in the `supply_chain_factory`?

pkg/discovery/stitching/node_uuid.go
```diff
@@ -203,6 +205,50 @@ func (gce *gceNodeUUIDGetter) GetUUID(node *api.Node) (string, error) {
...
+	}
+
+	//1. Get the suffix, this is one of our ids:
+	//   4200c244-1e27-8473-ab44-999195de924c
+	stitchingID := providerId[len(vspherePrefix):]
+
+	//2. Check if node.Status.NodeInfo.SystemUUID is filled by provider
+	//   If it is, append that as another id.
+	//   4200c244-1e27-8473-ab44-999195de924c,4200c244-1e27-8473-ab44-999195de924c
+	if node.Status.NodeInfo.SystemUUID != "" {
```
chlam4, line 245: Nit: Are there a lot of cases where the `SystemUUID` is the same as the `ProviderID`, just like the example here?  If yes, then it may be worth an extra check to make sure they are different before adding it?

irfanurrehman: Thanks makes sense. Updated!

pkg/discovery/stitching/node_uuid.go
irfanurrehman: Ignnoring this one... :) !

irfanurrehman:
@chlam4 thanks for having a look. Updated as per your suggestion. PTAL!

PR #453, opened 2020-07-26
Update metrics and metrics sink to support multipoint metrics

irfanurrehman:
@ading1977 @marvelyue 
Metrics sink support needed for https://vmturbo.atlassian.net/browse/OM-61095
Use [Points](https://github.com/turbonomic/kubeturbo/compare/master...irfanurrehman:multiple-datapoint-metrics?expand=1#diff-8f39f854b9577e1f9584f770bdaa16caR168) to update a single or multiple points of metric value to create and update the multipoint metric.

Please let me know if adding unit tests make sense here and I will add some.

### DIFF
--- a/pkg/discovery/metrics/metric.go
+++ b/pkg/discovery/metrics/metric.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"github.com/golang/glog"
 	v1 "k8s.io/api/core/v1"
 )
 
@@ -153,6 +154,7 @@ type Metric interface {
 	GetMetricProp() MetricProp
 	GetUID() string
 	GetValue() interface{}
+	UpdateValue(existing interface{})
 }
 
 type MetricFilterFunc func(m Metric) bool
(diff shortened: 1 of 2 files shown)


pkg/discovery/metrics/metric.go
```diff
@@ -153,39 +154,51 @@ type Metric interface {
...
 	entityID   string
+}
 
+type EntityResourceMetric struct {
+	EntityDetails
 	ResourceMetric
 }
 
 func NewEntityResourceMetric(eType DiscoveredEntityType, id string, rType ResourceType, mProp MetricProp,
 	v float64) EntityResourceMetric {
```
marvelyue, line 195: It would be helpful to have this input `v` as `interface{}` instead of `float64`

irfanurrehman: Fixed!

irfanurrehman:
@marvelyue fixed your finding. Please feel free to merge if this suits your need!

PR #454, opened 2020-07-27
Cont storage volumes fixes

irfanurrehman:
Fixes for the known issues with the storage volumes work, which are:
- Duplicate commodities show up if the same volume mount names and pvcs used across workloads (multiple XLs is a typical example). [Fixed]
- Solve reconfigure actions. In discussion with storage team, until volumes are added as providers in market all probes disable actions on them.[ Fixed]
- Stitch storage_amount with actual usage from pods and no keys. Actions based on usage will show up on volumes. [Fixed]
- Patch only used storage amount and let the cloud probe drive the capacity for the same.

### DIFF
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -364,11 +364,8 @@ func (builder *podEntityDTOBuilder) buyCommoditiesFromVolumes(pod *api.Pod, moun
 	for _, mount := range mounts {
 		mountName := mount.MountName
 		volEntityID := util.PodVolumeMetricId(podKey, mountName)
-		// We use <ns>/<pod-name>/<mounted-vol-name> as the commodity key
-		// to keep the relationship between pod and volume unique.
-		commKey := fmt.Sprintf("%s/%s", podKey, mountName)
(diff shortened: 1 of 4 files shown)


irfanurrehman:
cc @enlinxu @esara

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -385,6 +382,9 @@ func (builder *podEntityDTOBuilder) buyCommoditiesFromVolumes(pod *api.Pod, moun
 
 		provider := sdkbuilder.CreateProvider(proto.EntityDTO_VIRTUAL_VOLUME, providerVolUID)
 		dtoBuilder = dtoBuilder.Provider(provider)
+		dtoBuilder.IsMovable(proto.EntityDTO_VIRTUAL_VOLUME, false).
```
enlinxu, line 385: Do you think we can just make it controllable false?

irfanurrehman: I did replicate what other [probes are doing](https://rbcommons.com/s/VMTurbo/r/39397/diff/1/#0) wrt to the volumes as of now. I think we better stick to the storage teams recommendation.

PR #455, opened 2020-07-28
Flag to enable or disble action merge feature. 

pallavidn:
Enabled by default.

### DIFF
--- a/deploy/kubeturbo_yamls/step5_turbo_kubeturboDeploy.yaml
+++ b/deploy/kubeturbo_yamls/step5_turbo_kubeturboDeploy.yaml
@@ -30,6 +30,7 @@ spec:
           # change to https=false and port=10255 if unsecure kubelet read only is configured
           - --kubelet-https=true
           - --kubelet-port=10250
+          - --mergeactions=false
           # Uncomment to stitch using IP, or if using Openstack, Hyper-V/VMM
           #- --stitch-uuid=false
         volumeMounts:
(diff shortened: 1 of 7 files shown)


deploy/kubeturbo_yamls/step5_turbo_kubeturboDeploy.yaml
```diff
@@ -30,6 +30,7 @@ spec:
           # change to https=false and port=10255 if unsecure kubelet read only is configured
           - --kubelet-https=true
           - --kubelet-port=10250
+          - --mergeactions=false
```
ading1977, line 33: Maybe remove this?

deploy/kubeturbo_yamls/turbo_kubeturbo_full.yaml
```diff
@@ -105,6 +105,8 @@ spec:
           # change to https=false and port=10255 if unsecure kubelet read only is configured
           - --kubelet-https=true
           - --kubelet-port=10250
+          # Arg to enable or diable the action merge feature
+          - --mergeactions=false
```
ading1977, line 109: Same with above.

deploy/kubeturbo_yamls/step5_turbo_kubeturboDeploy.yaml
ading1977: Maybe this is better:

ading1977: This change is only against 7.22.5 branch, so it should be ok to leave it there.

PR #456, opened 2020-08-02
Fix crash in updated metrics interface

irfanurrehman:
@marvelyue not sure if you noticed this yet. However I did get a crash (which this change fixes) when I was testing kubeturbo with another feature on the latest.


marvelyue:
Thanks @irfanurrehman !

PR #459, opened 2020-08-03
Create new data sampling discovery workers

marvelyue:
**Intent**:
Create a new set of workers to discover more usage data samples from kubelet and store the multi-points data into a global metrics cache.

**Background**:
The main idea is to have these new workers run more frequently in parallel with the existing main discovery workers with the given configurable number of samples and sampling interval. We don't need to discover limit/request capacity more frequently, because 1) it’s very heavy to talk to API server (cluster monitor), and 2) container limit/request does not change very often.

**Implementation**:
1. Introduced 2 configurable variables:
  - `discovery-samples` - this defines the max size of the data stored in global metrics cache. Since our new workers keep running in Kubeturbo, this would avoid out of memory when Turbo server side crashes and no discovery is triggered. Default value is 10.
  - `discovery-sample-interval` - this defines how frequently our new workers scrap kubelet. Default value is 60 (seconds).
2. Create `SamplingDispatcher` to schedule assigning sampling discoveries based on configured samplingInterval. Set the timeout of one set of sampling discovery to the configured samplingInterval to avoid goroutine pile up.
3. Create a `globalEntityMetricSink` in `K8sDiscoveryClient` to store collected multi-points metrics samples discovered by sampling discovery workers.
4. Clear the `globalEntityMetricSink` and reschedule sampling discoveries after finishing a full discovery.
5. In `kubelet_monitor`, create new `scrapeKubeletForSampling` func to collect only resource usage data during sampling discoveries.

**Note**:
There will be 2 following changes in different PR to:
1. Merge sampled data from `globalEntityMetricSink` into the metrics sink of each individual main discovery worker
2. Construct multi-points UtilizationData on ContainerSpec based on collected data samples.

**Testing Done**:
1. Main discovery still works correctly.
2. Sampling discoveries are triggered every given samplingInterval. By debugging the code, `globalEntityMetricSink` store at most given samples of data for each collected metrics.
3. Will do more end-to-end tests once finishing the 2 following changes mentioned above.

### DIFF
--- a/pkg/discovery/k8s_discovery_client.go
+++ b/pkg/discovery/k8s_discovery_client.go
@@ -284,7 +284,7 @@ func (dc *K8sDiscoveryClient) discoverWithNewFramework(targetID string) ([]*prot
 	taskCount := dc.dispatcher.Dispatch(nodes, clusterSummary)
 	result := dc.resultCollector.Collect(taskCount)
 
-	// Clear globalEntityMetricSink cache after collecting main discovery results
+	// Clear globalEntityMetricSink cache after collecting full discovery results
 	dc.globalEntityMetricSink.ClearCache()
 	// Reschedule dispatch sampling discovery tasks for newly discovered nodes
 	dc.samplingDispatcher.ScheduleDispatch(nodes)
--- a/pkg/discovery/k8s_discovery_client.go
+++ b/pkg/discovery/k8s_discovery_client.go
@@ -2,6 +2,7 @@ package discovery
 
 import (
 	"fmt"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
 	"strings"
 	"time"
 
@@ -24,12 +25,14 @@ const (
 )
 
 type DiscoveryClientConfig struct {
-	probeConfig          *configs.ProbeConfig
-	targetConfig         *configs.K8sTargetConfig
-	ValidationWorkers    int
-	ValidationTimeoutSec int
-	DiscoveryWorkers     int
-	DiscoveryTimeoutSec  int
+	probeConfig                *configs.ProbeConfig
+	targetConfig               *configs.K8sTargetConfig
+	ValidationWorkers          int
+	ValidationTimeoutSec       int
+	DiscoveryWorkers           int
+	DiscoveryTimeoutSec        int
+	DiscoverySamples           int
+	DiscoverySampleIntervalSec int
 	// Strategy to aggregate Container utilization data on ContainerSpec entity
 	containerUtilizationDataAggStrategy string
 	// Strategy to aggregate Container usage data on ContainerSpec entity
@@ -43,7 +46,7 @@ func NewDiscoveryConfig(probeConfig *configs.ProbeConfig,
 	targetConfig *configs.K8sTargetConfig, ValidationWorkers int,
 	ValidationTimeoutSec int, containerUtilizationDataAggStrategy,
 	containerUsageDataAggStrategy string, ormClient *resourcemapping.ORMClient,
-	discoveryWorkers int, discoveryTimeoutMin int) *DiscoveryClientConfig {
+	discoveryWorkers, discoveryTimeoutMin, discoverySamples, discoverySampleIntervalSec int) *DiscoveryClientConfig {
 	if discoveryWorkers < minDiscoveryWorker {
 		glog.Warningf("Invalid number of discovery workers %v, set it to %v.",
 			discoveryWorkers, minDiscoveryWorker)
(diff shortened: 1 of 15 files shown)


pkg/discovery/k8s_discovery_client.go
```diff
@@ -257,12 +275,20 @@ func (dc *K8sDiscoveryClient) discoverWithNewFramework(targetID string) ([]*prot
...
-	// Collect the kubePod, kubeNamespace metrics, groups and kubeControllers from all the discovery workers
+	// Merge collected usage data samples from globalEntityMetricSink into the metric sink of each individual discovery worker.
+	// Collect the kubePod, kubeNamespace metrics, groups and kubeControllers from all the discovery workers.
 	taskCount := dc.dispatcher.Dispatch(nodes, clusterSummary)
 	result := dc.resultCollector.Collect(taskCount)
 
+	// Clear globalEntityMetricSink cache after collecting main discovery results
+	dc.globalEntityMetricSink.ClearCache()
+	// Reschedule dispatch sampling discovery tasks for newly discovered nodes
+	dc.samplingDispatcher.ScheduleDispatch(nodes)
```
irfanurrehman, line 290: [Probably no action needed but worth taking note]: This seems to mean that the nodes are refreshed only when discovery is trigerred. So if a node leaves in between discoveries, or a new one joins in during that time, the sampling data would be incoherent. This will especially be exacerbated when the probe runs disconnected from server, and then eventually joins in, or a patchy network, where probe keeps running online/offline over a period of time.

pkg/discovery/metrics/metric.go
```diff
@@ -154,7 +154,7 @@ type Metric interface {
 	GetMetricProp() MetricProp
 	GetUID() string
 	GetValue() interface{}
-	UpdateValue(existing interface{})
+	UpdateValue(existing interface{}, maxMetricPointsSize int) Metric
```
irfanurrehman, line 157: Suggestion: Consider setting `maxMetricPointsSize` while creating the metrics as part of `NewEntityResourceMetric()`, rather then in each `UpdateValue()` call.

pkg/discovery/metrics/metric_sink.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 type EntityMetricSink struct {
-	data *turbostore.Cache
+	data                *turbostore.Cache
+	maxMetricPointsSize int
```
irfanurrehman, line 14: You might not need this if you take the previous suggestion.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -126,20 +135,42 @@ func (m *KubeletMonitor) scrapeKubelet(node *api.Node) {
...
-	m.parsePodStats(summary.Pods)
+	currentMilliSec := time.Now().UnixNano() / int64(time.Millisecond)
+	m.parseNodeStats(summary.Node, currentMilliSec)
+	m.parsePodStats(summary.Pods, currentMilliSec)
 
 	glog.V(4).Infof("Finished scrape node %s.", node.Name)
 }
 
+// scrapeKubeletForSampling collects resource usage data samples for node, and containers and pods in the node.
+func (m *KubeletMonitor) scrapeKubeletForSampling(node *api.Node) {
```
irfanurrehman, line 146: Suggestion: This method duplicates code, consider writing another commond method which can be used here and in `scrapeKubelet()`

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -126,20 +135,42 @@ func (m *KubeletMonitor) scrapeKubelet(node *api.Node) {
...
+	}
+	// get summary information about the given node and the pods running on it.
+	summary, err := kc.GetSummary(ip)
+	if err != nil {
+		glog.Errorf("Failed to get resource metrics summary from %s: %s", node.Name, err)
+		return
+	}
+
+	currentMilliSec := time.Now().UnixNano() / int64(time.Millisecond)
+	m.parseNodeStats(summary.Node, currentMilliSec)
```
irfanurrehman, line 161: I thought the only data we did want to discover in this time based sampling is the container utilisation data useful for driving actions via constainerSpecs. Am I missing something?

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -223,31 +254,62 @@ func (m *KubeletMonitor) parseContainerStats(pod *stats.PodStats) (float64, floa
...
+	var memMetric metrics.EntityResourceMetric
+	// TODO Yue simplify if/else conditions when implementing merging global metrics sink to each individual main discovery worker
+	if m.isMainDiscovery {
+		cpuMetric = metrics.NewEntityResourceMetric(etype, key, metrics.CPU, metrics.Used, cpu)
+		memMetric = metrics.NewEntityResourceMetric(etype, key, metrics.Memory, metrics.Used, memory)
+	} else {
+		cpuMetric = metrics.NewEntityResourceMetric(etype, key, metrics.CPU, metrics.Used,
+			metrics.Points{
+				Values:    []float64{cpu},
+				Timestamp: timestamp,
```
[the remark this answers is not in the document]
marvelyue: Discussed with @ading1977, since the metrics are collected at the same time, we want to make sure the timestamp is the same for them all.

pkg/discovery/k8s_discovery_client.go
marvelyue: Thanks @irfanurrehman, this is a good point. As we discussed, it would be expensive to get all nodes every sampling interval from cluster. And I'll keep the current logic, and add a comment and additional warn message if there's error when getting metrics summary from an unavailable node during sampling discovery.

pkg/discovery/metrics/metric.go
marvelyue: Discussed with Irfan, will keep current logic which will be cleaner and less code changes.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
marvelyue: Current resource peak and used values are always the same. We also want to discover more usage data samples for pod and node, and then use max value as "peak" and average as "used" so as to better extract and show the peak number.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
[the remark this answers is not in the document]
irfanurrehman: @marvelyue please remember to put a comment here for this. Else somebody looking at this code later might think of improving it.. :)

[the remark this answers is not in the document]
marvelyue: Thanks @irfanurrehman. I added a comment here in the new PR: https://github.com/turbonomic/kubeturbo/pull/462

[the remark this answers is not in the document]
irfanurrehman: Thanks @marvelyue

PR #465, opened 2020-08-10
Update utilization data for container spec entity based on collected multiple data samples

marvelyue:
JIRA: https://vmturbo.atlassian.net/browse/OM-61197

**Intent**:
Update utilization data for container spec entity based on collected multiple usage data samples.

**Background**:
1. New sampling discoveries have been introduced to collect more usage data samples for VM, Pod, Container and ApplicationComponent by scraping kubelet more frequently.
2. Resource usage and peak data for the entity types above are calculated based the additional data samples.

This PR is the last major change to support more frequent data samples in kubeturbo to update utilization data of container spec entities by making using of the collected multiple data samples so as to provide more precise container resizing recommendations.

**Implementation**:
1. Change `ContainerSpec` struct to `ContainerSpecMetrics` struct with `ContainerMetrics` struct to represent the object for collecting container spec metrics with multiple container replicas metrics data points.
2. Remove the logic of creating `ContainerSpec` objects in `container_dto_builder.go`.
3. Create new `container_spec_metrics_collector.go` to collect resource capacity value and multiple resource usage data points for container replicas in each full discovery worker.
4. Combine `ContainerMetrics` used data points which belong to the same ContainerSpecMetrics discovered from different full discovery workers in `container_spec_discovery_worker.go`.
5. Update the logic in `container_spec_entity_dto_builder.go` to construct utilization data of container spec entity dto with multiple data samples in stead of one data sample.
6. Update `container_usage_data_aggregator.go` and `container_utilization_data_aggregator.go` to pass in `ContainerMetrics` with multiple usage data samples and aggregate usage/utilization data. By default usage data is the average of all collected container replicas usage of the same container spec, and utilization data points contains all container replicas usage data samples since last full discovery.
7. Change the resource metric value from `metrics.Points` to `[]metrics.Point` where each `metrics.Point` contains usage value and timestamp when the usage value is collected.

**Unit Tests**:
Updated unit tests and all tests are passed.

**Testing Done**:
**1. For **VIrtualMachine** commodity data**,

- 1.1 Before the whole changes of data sampling, here's a VM entity dto:
```
entityDTO {
  entityType: VIRTUAL_MACHINE
  id: "3c2910f0-2c62-4d57-a13d-569081f6aa17"
  displayName: "pt-k8s-master-2"
  commoditiesSold {
    commodityType: VCPU
    used: 1287.164931884286
    capacity: 10655.112
    peak: 1287.164931884286
    resizable: false
  }
  commoditiesSold {
    commodityType: VMEM
    used: 3230584.0
    capacity: 1.6398004E7
    peak: 3230584.0
    resizable: false
  }
```
where used and peak are always the same for resource commodities.

- 1.2 After the whole changes of data sampling, here's the corresponding VM entity dto:
```
entityDTO {
  entityType: VIRTUAL_MACHINE
  id: "3c2910f0-2c62-4d57-a13d-569081f6aa17"
  displayName: "pt-k8s-master-2"
  commoditiesSold {
    commodityType: VCPU
    used: 1196.5503725508838
    capacity: 10655.112
    peak: 1611.33108609876
    resizable: false
  }
  commoditiesSold {
    commodityType: VMEM
    used: 3234180.8
    capacity: 1.6398004E7
    peak: 3235072.0
    resizable: false
  }
```
where peak is differentiated from used value for resource commodities.

---
**2. For ContainerPod commodity data**,

- 2.1 Before the whole changes of data sampling, here's a pod entity dto:
```
entityDTO {
  entityType: CONTAINER_POD
  id: "3eea40ac-1b85-48e7-9e94-9f4e02c91c29"
  displayName: "turbo/kubeturbo-arsen-upgrade-644fc4db7c-w2fkj"
  commoditiesSold {
    commodityType: VCPU
    used: 0.814953577542
    capacity: 15982.667999999998
    peak: 0.814953577542
    resizable: false
  }
  commoditiesSold {
    commodityType: VMEM
    used: 20588.0
    capacity: 1.6397588E7
    peak: 20588.0
    resizable: false
  }
```
where used and peak are always the same.

- 2.2 After the changes, here's the corresponding pod entity dto:
```
entityDTO {
  entityType: CONTAINER_POD
  id: "3eea40ac-1b85-48e7-9e94-9f4e02c91c29"
  displayName: "turbo/kubeturbo-arsen-upgrade-644fc4db7c-w2fkj"
  commoditiesSold {
    commodityType: VCPU
    used: 2.8786322067906
    capacity: 15982.667999999998
    peak: 20.492433498887998
    resizable: false
  }
  commoditiesSold {
    commodityType: VMEM
    used: 20508.8
    capacity: 1.6397588E7
    peak: 20516.0
    resizable: false
  }
```
where used and peak values are different.

---
**3. For Container commodity data**,

- 3.1 Before changes:
```
entityDTO {
  entityType: CONTAINER
  id: "0f65d4d3-2c32-4f69-9f32-a4597ccc5c95-1"
  displayName: "monitoring/node-exporter-s5ntq/kube-rbac-proxy"
  commoditiesSold {
    commodityType: VCPU
    used: 1.0836168990659998
    capacity: 53.27556
    peak: 1.0836168990659998
    resizable: true
  }
  commoditiesSold {
    commodityType: VMEM
    used: 16696.0
    capacity: 40960.0
    peak: 16696.0
    resizable: true
  }
  commoditiesSold {
    commodityType: VCPU_REQUEST
    used: 1.0836168990659998
    capacity: 26.63778
    peak: 1.0836168990659998
    resizable: true
  }
  commoditiesSold {
    commodityType: VMEM_REQUEST
    used: 16696.0
    capacity: 20480.0
    peak: 16696.0
    resizable: true
  }
```

- 3.2 After changes:
```
entityDTO {
  entityType: CONTAINER
  id: "0f65d4d3-2c32-4f69-9f32-a4597ccc5c95-1"
  displayName: "monitoring/node-exporter-s5ntq/kube-rbac-proxy"
  commoditiesSold {
    commodityType: VCPU
    used: 1.2692060416151998
    capacity: 53.27556
    peak: 2.958932593734
    resizable: true
  }
  commoditiesSold {
    commodityType: VMEM
    used: 16559.2
    capacity: 40960.0
    peak: 16604.0
    resizable: true
  }
  commoditiesSold {
    commodityType: VCPU_REQUEST
    used: 1.2692060416151998
    capacity: 26.63778
    peak: 2.958932593734
    resizable: true
  }
  commoditiesSold {
    commodityType: VMEM_REQUEST
    used: 16559.2
    capacity: 20480.0
    peak: 16604.0
    resizable: true
  }
```

---
**4. For ApplicationComponent commodity data**,

- 4.1 Before changes:
```
entityDTO {
  entityType: APPLICATION_COMPONENT
  id: "App-e5516272-4240-431a-8d0e-e1b83a996e7e-2"
  displayName: "App-monitoring/prometheus-k8s-1/rules-configmap-reloader"
  commoditiesBought {
    providerId: "e5516272-4240-431a-8d0e-e1b83a996e7e-2"
    bought {
      commodityType: VCPU
      used: 0.006081405173999999
      peak: 0.006081405173999999
    }
    bought {
      commodityType: VMEM
      used: 5168.0
      peak: 5168.0
    }
```

- 4.2 After changes:
```
entityDTO {
  entityType: APPLICATION_COMPONENT
  id: "App-e5516272-4240-431a-8d0e-e1b83a996e7e-2"
  displayName: "App-monitoring/prometheus-k8s-1/rules-configmap-reloader"
  commoditiesSold {
    commodityType: APPLICATION
    key: "cb248341-ffc4-4a80-8513-1152977ba611-2"
    capacity: 100000.0
  }
  commoditiesBought {
    providerId: "e5516272-4240-431a-8d0e-e1b83a996e7e-2"
    bought {
      commodityType: VCPU
      used: 0.014935004112599999
      peak: 0.14370815932199998
    }
    bought {
      commodityType: VMEM
      used: 5186.4
      peak: 5192.0
    }
```

---
**5. For ContainerSpec entity**

- 5.1 Before changes, here's entity dto of a ContainerSpec with one container replica:
```
entityDTO {
  entityType: CONTAINER_SPEC
  id: "0761a65a-7f2f-49df-9f2f-5b86024791f4/prometheus"
  displayName: "prometheus"
  commoditiesSold {
    commodityType: VCPU
    used: 162.77071941093598
    capacity: 15982.667999999998
    peak: 162.77071941093598
    active: true
    resizable: true
    utilizationData {
      point: 1.0184202
      lastPointTimestampMs: 1597196174149
      intervalMs: 0
    }
  }
  commoditiesSold {
    commodityType: VMEM
    used: 637196.0
    capacity: 1.6397572E7
    peak: 637196.0
    active: true
    resizable: true
    utilizationData {
      point: 3.8859167686533103
      lastPointTimestampMs: 1597196174149
      intervalMs: 0
    }
  }
```
where used and peak values are same and utilizationData has only one point.

- 5.2 After changes, here's the corresponding entity dto of the ContainerSpec with one container replica:
```
entityDTO {
  entityType: CONTAINER_SPEC
  id: "0761a65a-7f2f-49df-9f2f-5b86024791f4/prometheus"
  displayName: "prometheus"
  commoditiesSold {
    commodityType: VCPU
    used: 213.1093339071918
    capacity: 15982.667999999998
    peak: 380.551845511266
    active: true
    resizable: true
    utilizationData {
      point: 0.8913513
      point: 0.86388245
      point: 1.8296162833333334
      point: 1.2034350333333335
      point: 2.2074865999999997
      point: 1.1176684
      point: 1.0257308833333334
      point: 0.9222266499999999
      point: 2.3810282833333334
      point: 0.8913513
      lastPointTimestampMs: 1597195063836
      intervalMs: 53305
    }
  }
  commoditiesSold {
    commodityType: VMEM
    used: 633227.6
    capacity: 1.6397572E7
    peak: 637604.0
    active: true
    resizable: true
    utilizationData {
      point: 3.8884049419023743
      point: 3.8496918934095854
      point: 3.8509359800341176
      point: 3.85071643533567
      point: 3.8504724967818404
      point: 3.8506188599141384
      point: 3.8511067370217984
      point: 3.8503261336495425
      point: 3.8864778273271186
      point: 3.8884049419023743
      lastPointTimestampMs: 1597195063836
      intervalMs: 53305
    }
  }
```
where used and peak values are different, and there are 10 data points in utilization data with intervalMs 53331 milliseconds (53 seconds which is around 1 mins)

### DIFF
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -78,22 +78,18 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 		}
 		commSoldBuilder := sdkbuilder.NewCommodityDTOBuilder(commodityType)
 
-		// Aggregate container replicas utilization data
-		utilizationDataPoints, lastPointTimestamp, samplingDuration, err := builder.containerUtilizationDataAggregator.Aggregate(resourceMetrics)
+		// Aggregate container replicas utilization data.
+		// Note that the returned dataInterval is not the real sampling interval because there could be multiple data
+		// points from container replicas discovered at the same time in one set of data samples. This is just a calculated
+		// equivalent interval to be fed into percentile based algorithm in Turbo server side.
+		utilizationDataPoints, lastPointTimestamp, dataInterval, err := builder.containerUtilizationDataAggregator.Aggregate(resourceMetrics)
 		if err != nil {
 			glog.Errorf("Error aggregating %s utilization data for ContainerSpec %s, %v",
 				resourceType, containerSpecMetrics.ContainerSpecId, err)
 			continue
 		}
-		// Calculate interval between data points. Note that this is not the real sampling interval because there could
-		// be multiple data points from container replicas discovered at the same time in one set of data samples. This
-		// is to just calculate an equivalent interval to be fed into percentile based algorithm in Turbo server side.
-		var interval int32
-		if len(utilizationDataPoints) > 1 {
-			interval = samplingDuration / int32(len(utilizationDataPoints)-1)
-		}
 		// Construct UtilizationData with multiple data points, last point timestamp in milliseconds and interval in milliseconds
-		commSoldBuilder.UtilizationData(utilizationDataPoints, lastPointTimestamp, interval)
+		commSoldBuilder.UtilizationData(utilizationDataPoints, lastPointTimestamp, dataInterval)
 
 		// Aggregate container replicas usage data (capacity, used and peak)
 		aggregatedCap, aggregatedUsed, aggregatedPeak, err :=
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -79,17 +79,17 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 		commSoldBuilder := sdkbuilder.NewCommodityDTOBuilder(commodityType)
 
 		// Aggregate container replicas utilization data.
-		// Note that the returned dataInterval is not the real sampling interval because there could be multiple data
+		// Note that the returned dataIntervalMs is not the real sampling interval because there could be multiple data
 		// points from container replicas discovered at the same time in one set of data samples. This is just a calculated
-		// equivalent interval to be fed into percentile based algorithm in Turbo server side.
-		utilizationDataPoints, lastPointTimestamp, dataInterval, err := builder.containerUtilizationDataAggregator.Aggregate(resourceMetrics)
+		// equivalent interval between 2 data points to be fed into percentile based algorithm in Turbo server side.
+		utilizationDataPoints, lastPointTimestamp, dataIntervalMs, err := builder.containerUtilizationDataAggregator.Aggregate(resourceMetrics)
 		if err != nil {
 			glog.Errorf("Error aggregating %s utilization data for ContainerSpec %s, %v",
 				resourceType, containerSpecMetrics.ContainerSpecId, err)
 			continue
 		}
 		// Construct UtilizationData with multiple data points, last point timestamp in milliseconds and interval in milliseconds
-		commSoldBuilder.UtilizationData(utilizationDataPoints, lastPointTimestamp, dataInterval)
+		commSoldBuilder.UtilizationData(utilizationDataPoints, lastPointTimestamp, dataIntervalMs)
 
 		// Aggregate container replicas usage data (capacity, used and peak)
 		aggregatedCap, aggregatedUsed, aggregatedPeak, err :=
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -2,37 +2,36 @@ package dtofactory
 
 import (
 	"github.com/golang/glog"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/worker/aggregation"
 	sdkbuilder "github.com/turbonomic/turbo-go-sdk/pkg/builder"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
-	"time"
 )
 
 var (
-	ContainerSpecCommoditiesSold = []proto.CommodityDTO_CommodityType{
-		proto.CommodityDTO_VCPU,
-		proto.CommodityDTO_VMEM,
-		proto.CommodityDTO_VCPU_REQUEST,
-		proto.CommodityDTO_VMEM_REQUEST,
+	ContainerSpecResourceTypes = []metrics.ResourceType{
+		metrics.CPU,
+		metrics.Memory,
+		metrics.CPURequest,
+		metrics.MemoryRequest,
 	}
 )
 
 type containerSpecDTOBuilder struct {
-	// Map from ContainerSpec ID to ContainerSpec entity which contains list of individual container replica commodities
-	// data to be aggregated
-	containerSpecMap map[string]*repository.ContainerSpec
+	// Map from ContainerSpec ID to ContainerSpecMetrics which contains list of container replicas usage metrics data to be aggregated
+	containerSpecMetricsMap map[string]*repository.ContainerSpecMetrics
 	// Aggregator to aggregate container replicas commodity utilization data
 	containerUtilizationDataAggregator aggregation.ContainerUtilizationDataAggregator
 	// Aggregator to aggregate container replicas commodity usage data (used, peak and capacity)
 	containerUsageDataAggregator aggregation.ContainerUsageDataAggregator
 }
 
-func NewContainerSpecDTOBuilder(containerSpecMap map[string]*repository.ContainerSpec,
+func NewContainerSpecDTOBuilder(containerSpecMetricsMap map[string]*repository.ContainerSpecMetrics,
 	containerUtilizationDataAggregator aggregation.ContainerUtilizationDataAggregator,
 	containerUsageDataAggregator aggregation.ContainerUsageDataAggregator) *containerSpecDTOBuilder {
 	return &containerSpecDTOBuilder{
-		containerSpecMap:                   containerSpecMap,
+		containerSpecMetricsMap:            containerSpecMetricsMap,
 		containerUtilizationDataAggregator: containerUtilizationDataAggregator,
 		containerUsageDataAggregator:       containerUsageDataAggregator,
 	}
@@ -40,7 +39,7 @@ func NewContainerSpecDTOBuilder(containerSpecMap map[string]*repository.Containe
 
 func (builder *containerSpecDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error) {
 	var result []*proto.EntityDTO
-	for containerSpecId, containerSpec := range builder.containerSpecMap {
+	for containerSpecId, containerSpec := range builder.containerSpecMetricsMap {
 		entityDTOBuilder := sdkbuilder.NewEntityDTOBuilder(proto.EntityDTO_CONTAINER_SPEC, containerSpecId)
 		entityDTOBuilder.DisplayName(containerSpec.ContainerSpecName)
 		commoditiesSold, err := builder.getCommoditiesSold(containerSpec)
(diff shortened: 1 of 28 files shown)


pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -62,37 +61,46 @@ func (builder *containerSpecDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error)
...
-		currentMs := time.Now().UnixNano() / int64(time.Millisecond)
-		// Currently we only collect one set of metrics data points from Kubernetes within a discovery cycle so the
-		// lastPointTimestampMs of utilizationData is current timestamp in milliseconds and intervalMs is 0.
-		commSoldBuilder.UtilizationData(utilizationDataPoints, currentMs, 0)
+		// Calculate interval between data points. Note that this is not the real sampling interval because there could
+		// be multiple data points from container replicas discovered at the same time in one set of data samples. This
+		// is to just calculate an equivalent interval to be fed into percentile based algorithm in Turbo server side.
+		var interval int32
+		if len(utilizationDataPoints) > 1 {
+			interval = samplingDuration / int32(len(utilizationDataPoints)-1)
```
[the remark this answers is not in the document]
marvelyue: Good point. We can just do the calculation inside the `Aggregate()` method above and return the interval, which will be cleaner here. I'll make the change

pkg/discovery/dtofactory/general_builder.go
```diff
@@ -194,14 +194,14 @@ func (builder generalBuilder) metricValue(entityType metrics.DiscoveredEntityTyp
 
 	value := metric.GetValue()
 	switch value.(type) {
-	case metrics.Points:
+	case []metrics.Point:
```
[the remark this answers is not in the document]
marvelyue: Yes, I realized that it would be helpful to still keep the timestamp on each point collected to calculate the interval above

marvelyue:
> A better way would be to use the time stamp attached to the discovered metric updated by the kubelet itself for example [this one for cpu stats](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/apis/stats/v1alpha1/types.go#L201). This is a more realistic time stamp that can be attached with the metric. It does not at all need to be part of this PR and can be taken an a follow on.

Thanks @irfanurrehman, this is a very good suggestion! I didn't notice the stats returned from kubelet has timestamp. The changes may not be very small in `kubelet_monitor.go` and besides we don't need very precise timestamp and current manually created timestamp should be closed to the real timestamp. Like you said, I'll add a TODO in the comment for now.

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -62,37 +61,42 @@ func (builder *containerSpecDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error)
...
 		}
 		commSoldBuilder := sdkbuilder.NewCommodityDTOBuilder(commodityType)
 
-		// Aggregate container replicas utilization data
-		utilizationDataPoints, err := builder.containerUtilizationDataAggregator.Aggregate(commodities)
+		// Aggregate container replicas utilization data.
+		// Note that the returned dataInterval is not the real sampling interval because there could be multiple data
+		// points from container replicas discovered at the same time in one set of data samples. This is just a calculated
+		// equivalent interval to be fed into percentile based algorithm in Turbo server side.
+		utilizationDataPoints, lastPointTimestamp, dataInterval, err := builder.containerUtilizationDataAggregator.Aggregate(resourceMetrics)
```
dblinn, line 85: nit: could we label the time units on the interval (ie dataIntervalMs)?

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -62,37 +61,42 @@ func (builder *containerSpecDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error)
...
+			glog.Errorf("Error aggregating %s utilization data for ContainerSpec %s, %v",
+				resourceType, containerSpecMetrics.ContainerSpecId, err)
 			continue
 		}
-		currentMs := time.Now().UnixNano() / int64(time.Millisecond)
-		// Currently we only collect one set of metrics data points from Kubernetes within a discovery cycle so the
-		// lastPointTimestampMs of utilizationData is current timestamp in milliseconds and intervalMs is 0.
-		commSoldBuilder.UtilizationData(utilizationDataPoints, currentMs, 0)
+		// Construct UtilizationData with multiple data points, last point timestamp in milliseconds and interval in milliseconds
+		commSoldBuilder.UtilizationData(utilizationDataPoints, lastPointTimestamp, dataInterval)
```
dblinn, line 92: Maybe we could talk about this, but if I'm reading the code correctly I don't think we are providing a good dataInterval. As I understand the platform's handling of the data interval, this value should be the time interval between individual samples in the utilization data array that we return in the discovery results. This means, for example, if we have 100 samples, a data interval of 10 minutes, and a last timestamp of 12 PM, the platform will interpret that to mean the last sample in the array is at 12 PM, the second to last sample is at 12 PM - 10 minutes, the third to last sample is at 12 PM - 20 minutes, etc. and the first sample is at 12 PM - 99 * 10 minutes. This means that the first data point will be over half a day before the first one, and I don't think that's what we want.

I think the data interval we actually want to set here is `dataIntervalMs / len(utilizationDataPoints)` so that, in the example above, the last data point will be interpreted as being at 12 PM and the first data point will be at 12 PM - 99 * (10 minutes / 100), or roughly 10 minutes before the last data point.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
```diff
@@ -65,18 +61,16 @@ func (maxUsageDataAggregator *maxUsageDataAggregator) String() string {
...
 	maxUsed := 0.0
-	maxPeak := 0.0
-	for _, commodity := range commodities {
-		maxCapacity = math.Max(maxCapacity, *commodity.Capacity)
-		maxUsed = math.Max(maxUsed, *commodity.Used)
-		maxPeak = math.Max(maxPeak, *commodity.Peak)
+	peak := 0.0
+	for _, usedPoint := range resourceMetrics.Used {
+		maxUsed = math.Max(maxUsed, usedPoint.Value)
+		peak = math.Max(peak, usedPoint.Value)
```
dblinn, line 73: nit: it looks like maxUsed and peak are guaranteed to always be exactly the same. Is there a good reason to calculate the same value twice?

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
```diff
@@ -37,24 +38,33 @@ func (allDataAggregator *allUtilizationDataAggregator) String() string {
...
-		utilization := used / capacity * 100
+	var firstTimestamp int64
+	var lastTimestamp int64
+	for _, usedPoint := range resourceMetrics.Used {
+		utilization := usedPoint.Value / capacity * 100
 		utilizationDataPoints = append(utilizationDataPoints, utilization)
+		if firstTimestamp == 0 || usedPoint.Timestamp < firstTimestamp {
+			firstTimestamp = usedPoint.Timestamp
+		}
+		lastTimestamp = int64(math.Max(float64(lastTimestamp), float64(usedPoint.Timestamp)))
```
dblinn, line 60: I think it's really not a good idea to cast to float64 first because these timestamps are pretty large numbers and so if they are close together the loss in precision can actually have a significant impact on the calculation.

It's unfortunate there's no nice way to take the max of two integer types in go, but it would be better to use regular comparison operators here instead of casting to float64 and back.

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
dblinn: Yue and I discussed. The data interval is actually correct. My understanding of it was based on an out-of-date comment in the code but the implementation is actually correct.

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
irfanurrehman: @marvelyue @dblinn What would be your comment about keeping this as the golangs origin type `Time.time`? The comparisons could safely be done, using the comparisons available in `Time` package, on that then. In case server does not agree with it, we can cast the final value as the server accepted type after all this processing is complete.

marvelyue: Thanks @irfanurrehman for the suggestion. I just updated the code by simply using normal comparison on int64 here per @dblinn's suggestion:
```
                if usedPoint.Timestamp > lastTimestamp {
			lastTimestamp = usedPoint.Timestamp
		}
```
I think using the comparison in `Time` package (like `After` `Before`) would be similar as comparing int64 values above. Either should work

PR #468, opened 2020-08-17
Add pod move support for pods with persistent volumes

irfanurrehman:
This has turned out to be a large peice of work and also almost barely/hackably doable.
There are couple of things this PR doesn't do as of now and I intend to follow on:
- XL pods with volumes manged by operator are still not movable as this mechanism needs to update the parent and parents parent (both replicaset and deployment), in which deployment is controlled by operator and thus can be updated only via the operator. This is also doable but involves much more work where we will need to expose the deployments `spec.paused` field also from the operator in addition to setting up the code to update the operator via the operator resource mapping. Which also means that it won't work if the ORM is not already installed and CR setup with resource mapping together with the operator.
- The get pods parent and parent's parent methods are not neat and can be improved, I will at some point want to update them to something like a `podInfoGetter` in utils, which will support getting relevent info to relevant callers.

I would not want this to be this hacky, but I have this as the only feasible way of doing podmoves for a volume attached pod as of now (tried mutiple things). I might be missing something, so suggestions are certainly welcome. 

cc @enlinxu @esara

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -1,13 +1,19 @@
 package executor
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/golang/glog"
 	"github.com/turbonomic/kubeturbo/pkg/action/util"
+	"github.com/turbonomic/kubeturbo/pkg/cluster"
 	podutil "github.com/turbonomic/kubeturbo/pkg/discovery/util"
+	commonutil "github.com/turbonomic/kubeturbo/pkg/util"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 
 	api "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -71,23 +77,34 @@ func genNewPodName(oldPod *api.Pod) string {
 }
 
 // Move pod to node nodeName in four steps:
-//  step1: create a clone pod of the original pod (without labels)
-//  step2: wait until the cloned pod is ready
-//  step3: delete the original pod
-//  step4: add the labels to the cloned pod
-func movePod(client *kclient.Clientset, pod *api.Pod, nodeName string, retryNum int) (*api.Pod, error) {
-	podClient := client.CoreV1().Pods(pod.Namespace)
+//  stepA1: create a clone pod of the original pod (without labels)
+//  stepA2: wait until the cloned pod is ready
+//  stepA3: delete the original pod
+//  stepA4: add the labels to the cloned pod
+// If a pod has a persistent volume attached the steps are different:
+//  stepB1: create a clone pod of the original pod (without labels)
+//  stepB2: if the parent has parent (rs has deployment) pause the rollout
+//  stepB3: change the scheduler of parent to non-default (turbo-scheduler)
+//  stepB4: delete the original pod
+//  stepB5: wait until the cloned pod is ready
+//  stepB6: add the labels to the cloned pod
+//  stepB7: change the scheduler of parent back to to default-scheduler
+//  stepB8: if the parent has parent, unpause the rollout
+// TODO: add support for operator controlled parent or parent's parent.
+func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName,
+	parentKind, parentName string, retryNum int) (*api.Pod, error) {
+	podClient := clusterScraper.Clientset.CoreV1().Pods(pod.Namespace)
 	//NOTE: do deep-copy if the original pod may be modified outside this function
 	labels := pod.Labels
 
-	//1. create a clone pod--podC of the original pod--podA
-	npod, err := createClonePod(client, pod, nodeName)
+	//step A1/B1. create a clone pod--podC of the original pod--podA
+	npod, err := createClonePod(clusterScraper.Clientset, pod, nodeName)
 	if err != nil {
 		glog.Errorf("Move pod failed: failed to create a clone pod: %v", err)
 		return nil, err
 	}
 
-	//delete the clone pod if this action fails
+	//delete the clone pod if any of the below action fails
 	flag := false
 	defer func() {
 		if !flag {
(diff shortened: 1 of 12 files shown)


irfanurrehman:
@enlinxu this now has a flag `--fail-volume-pod-moves` for which the default value will be `true`.
Additionally I did some testing and find that the pod moves can happen for XL without the need to pause the deployment via the operator CR. It seems that this particular operator ignores the deployments `paused` field while reconciling the resources. I am therefor removing the check to fail the move if it finds that the parent resource is managed by an operator as in this [last commit](https://github.com/turbonomic/kubeturbo/pull/468/commits/dce1d711475e22552b00780e1ed50c801cee50b8). We can do a more foolproof implementation when we get a customer or an env which has a different resource (eg a replicaset) which is managed by an operator.

pkg/action/executor/move_util.go
```diff
@@ -97,22 +121,76 @@ func movePod(client *kclient.Clientset, pod *api.Pod, nodeName string, retryNum
...
+		invalid := false
+		// step B7: (via defer)
+		defer func() {
+			glog.V(3).Infof("Updating scheduler of %s for pod %s back to default.", parentKind, podQualifiedName)
+			changeScheduler(pClient, parent, valid)
+
+		}()
+		// step B3:
+		glog.V(3).Infof("Updating scheduler of %s for pod %s to unknown (turbo-scheduler).", parentKind, pod.Name)
+		changeScheduler(pClient, parent, invalid)
```
ading1977, line 161: Do we need to check error here and return if there is anything wrong?

pkg/action/executor/move_util.go
```diff
@@ -97,22 +121,76 @@ func movePod(client *kclient.Clientset, pod *api.Pod, nodeName string, retryNum
...
+		// via a config either in kubeturbo or driven from server UI.
+		delOpt := &metav1.DeleteOptions{}
+		if err := podClient.Delete(pod.Name, delOpt); err != nil {
+			glog.Errorf("Move pod warning: failed to delete original pod: %v", err)
+			return nil, err
+		}
+	}
+
+	//step A2/B5: wait until podC gets ready
+	err = podutil.WaitForPodReady(clusterScraper.Clientset, npod.Namespace, npod.Name, nodeName,
```
ading1977, line 177: I can think of a few scenarios that this might run into problems (although chances are rare):

* If the previous Delete of the original pod does not complete for whatever reason (e.g., stuck in Terminating), then the cloned pod won't start. After timeout, the cloned pod is deleted, and scheduler/rollout is reenabled, at which time, a new pod will be scheduled to bump up the replica, but the new pod will probably still fail to start because the original pod is stuck in Terminating
* The original pod is deleted successfully, but the cloned pod fails to start for whatever reason. After timeout, the cloned pod is deleted, and scheduler/rollout is reenabled, at which time, a new pod will be scheduled to bump up the replica, but the new pod could run anywhere in the cluster.

I do not have a good solution (if you agree that the above can happen), but at least we can keep them in mind in case we run into them and we can give an explanation.

pkg/action/executor/move_util.go
irfanurrehman: Thanks for catching this. Updated!

pkg/action/executor/move_util.go
irfanurrehman: As discussed, I will compile a proactive readme which can list possible corner cases, which although rare can happen.

irfanurrehman:
@ading1977 updated. PTAL!

pkg/action/executor/rescheduler.go
```diff
@@ -149,36 +151,48 @@ func (r *ReScheduler) preActionCheck(pod *api.Pod, node *api.Node) error {
...
 
 func (r *ReScheduler) reSchedule(pod *api.Pod, node *api.Node) (*api.Pod, error) {
+	reScheduled := false
+	var finalErr error
+	defer func() {
+		if !reScheduled {
+			glog.V(2).Infof("Move action aborted: %v.", finalErr)
+		}
+	}()
+
```
marvelyue, line 161: These seem to be redundant. We may not need this defer func, `reScheduled` bool and `finalErr`. Whenever there's error below, this `reSchedule` func will return the error and its parent func `Execute` will print the error and fail the action execution:
```
func (r *ReScheduler) Execute(input *TurboActionExecutorInput) (*TurboActionExecutorOutput, error) {
        ...
        npod, err := r.reSchedule(pod, node)
	if err != nil {
		glog.Errorf("Failed to execute pod move: %v.", err)
		return &TurboActionExecutorOutput{}, err
	}
```

pkg/cluster/cluster_info_scraper.go
```diff
@@ -286,26 +287,31 @@ func (s *ClusterScraper) GetPVCs(namespace string, opts metav1.ListOptions) ([]*
...
-// GetPodGrandparentInfo gets grandParent (parent's parent) information of a pod: kind, name and uid
-// If pod's parent does not have parent, then return pod's parent info.
+// GetPodGrandparentInfo gets grandParent (parent's parent) information of a pod: kind, name, uid
+// If parent does not have parent, then return parent info.
 // Note: if parent kind is "ReplicaSet", then its parent's parent can be a "Deployment"
 //       or if its a "ReplicationController" its parent could be "DeploymentConfig" (as in openshift).
-func (s *ClusterScraper) GetPodGrandparentInfo(pod *api.Pod) (string, string, string, error) {
-	// Get pod controller info from cache if exists
+// The function also returns the retrieved parent and parents crud interface for use by the callers.
+func (s *ClusterScraper) GetPodGrandparentInfo(pod *api.Pod, ignoreCache bool) (string, string,
```
marvelyue, line 295: I might be missing something here for the `ignoreCache` input argument. Looks all the places to call this `GetPodGrandparentInfo` func pass `ignoreCache` as false. And as I understand, the parent or grand parent info of a pod wouldn't be changed for the same pod with the same UID (the `podControllerInfoKey` is created from pod.UID). Maybe we could remove this argument?

pkg/cluster/cluster_info_scraper.go
```diff
@@ -318,30 +324,33 @@ func (s *ClusterScraper) GetPodGrandparentInfo(pod *api.Pod) (string, string, st
 			Resource: commonutil.ReplicationControllerResName}
 	case commonutil.KindReplicaSet:
 		res = schema.GroupVersionResource{
-			Group:    commonutil.K8sAPIReplicasetGV.Group,
-			Version:  commonutil.K8sAPIReplicasetGV.Version,
+			Group:    commonutil.K8sAPIDeploymentGV.Group,
+			Version:  commonutil.K8sAPIDeploymentGV.Version,
```
marvelyue, line 328: Looks we need `K8sAPIReplicasetGV` here instead of `K8sAPIDeploymentGV`. Or is there a reason to use Deployment GroupVersion to get ReplicaSet resource?

pkg/cluster/cluster_info_scraper.go
```diff
@@ -318,30 +324,33 @@ func (s *ClusterScraper) GetPodGrandparentInfo(pod *api.Pod) (string, string, st
...
+		return "", "", "", nil, nil, err
 	}
 	//2.2 get parent's parent info by parsing ownerReferences:
 	rsOwnerReferences := obj.GetOwnerReferences()
 	if rsOwnerReferences != nil && len(rsOwnerReferences) > 0 {
 		gkind, gname, guid := util.ParseOwnerReferences(rsOwnerReferences)
-		if len(gkind) > 0 && len(gname) > 0 && len(guid) > 0 {
-			s.cacheControllerInfo(podControllerInfoKey, gkind, gname, guid)
-			return gkind, gname, guid, nil
+		if len(gkind) > 0 && len(gname) > 0 {
```
marvelyue, line 346: Nit: Although it's unlikely `guid` is empty when `gkind` and `gname` exist,  I think it would be safer to keep the previous `&& len(guid) > 0` condition.

pkg/action/executor/rescheduler.go
irfanurrehman: Updated!

pkg/cluster/cluster_info_scraper.go
irfanurrehman: Thanks for catching this. This being false from all invocations is a miss. Actually the intention is to be able to retrieve the resources always, even if the cache exists in the case of volume based pod moves. The flag will be true [here](https://github.com/turbonomic/kubeturbo/pull/468/files#diff-623e1452732b5bf67c1efd4dd5ebb66aR246) now.

pkg/cluster/cluster_info_scraper.go
irfanurrehman: Right. I guess this was introduced while rebasing my changes on latest. Updated!

pkg/cluster/cluster_info_scraper.go
irfanurrehman: Updated. I think this too was introduced while rebasing these changes on latest.

ading1977 (approved):

irfanurrehman:
@enlinxu I have done enough testing and error case handling in this one and I beleive it should be fine to be run in our SaaS env.
@ading1977 @marvelyue I would appreciate your review on the last commit.

pkg/action/executor/move_util.go
```diff
@@ -132,6 +250,219 @@ func movePod(client *kclient.Clientset, pod *api.Pod, nodeName string, retryNum
...
+				return podClient.Delete(pod.Name, &metav1.DeleteOptions{})
+			})
+		if err != nil {
+			glog.Errorf("Failed to delete pending pod: %s.", pod.Name)
+		}
+	}
+
+}
+
+func newPendingInvalidPods(list1, list2 *api.PodList) []*api.Pod {
```
marvelyue, line 280: Nit: Maybe it's more readable to name the arguments more meaningful to something like `newPodList` and `existingPodList`

pkg/action/executor/move_util.go
```diff
@@ -132,6 +250,219 @@ func movePod(client *kclient.Clientset, pod *api.Pod, nodeName string, retryNum
...
+	var diffList []*api.Pod
+	for _, pod1 := range list1.Items {
+		found := false
+		for _, pod2 := range list2.Items {
+			if pod1.Name == pod2.Name {
+				found = true
+				break
+			}
+		}
```
marvelyue, line 289: Nit: Not something critical. Would it be better to convert `list2` into a HashSet of pod names like `existingPodSet` first at the start of this function, and in each iteration of `list1` check if `pod1` name exists in `existingPodSet`? This could improve the time complexity of this func from O(N^2) to O(N) which would be more efficient especially when there are lots of pods we care about here.

marvelyue (approved):
Only some minor comments nothing related to functionality. Overall looks good! And thanks a lot @irfanurrehman, really lots of corner cases you have handled here!

pkg/action/executor/move_util.go
irfanurrehman: That might actually be more costly in terms of conversion (from original slice into the set, which is a map) in general. Typically a parent has a few replicas (for example 4) its quite uncommon to see double digit replicas of a workload.

pkg/action/executor/move_util.go
irfanurrehman: Thanks for the suggestion, skipping this for now... :)

PR #469, opened 2020-08-17
Improve Kubeturbo discovery performance by caching pod ownerReferences during cluster monitoring

marvelyue:
JIRA: https://vmturbo.atlassian.net/browse/OM-61188

**Intent**:
In cluster monitoring, we invoke api call in every discover for pod owner info, which is actually very expensive. We should introduce cache for those metadata kind of information, to avoid repeated api call without new information.

**Implementation**:
1. In `cluster_info_scraper.go`, create a `cache`(`turbostore.ITurboCache` type) in `ClusterScraper` struct with expiration duration as `defaultCacheTTL` (24 hours). `TurboCache` is an existing struct which makes use of cache library https://github.com/patrickmn/go-cache. The cached data will be cleaned up after `defaultCacheTTL` (24 hours). Meanwhile the lock inside go-cache library is a `sync.RWMutex`, which is very efficient in the case where there are more reads than writes.
2. Move the main logic from `GetPodGrandInfo` func in `pod_util` to `GetPodGrandparentInfo` in `cluster_info_scraper`. And use cache data inside `GetPodGrandparentInfo` func if exists.
3. There are 2 places which call `GetPodGrandparentInfo` func: 1) `ClusterMonitor::getPodOwner` in `cluster_monitor` and 2)`newK8sControllerUpdaterViaPod` func in `k8s_controller_updater`. Refactor the code to use the same `clusterScraper` (with the same cache) in both places.

**Testing Done**:

1. After the changes, controller kind is still shown correctly in UI:
Screen Shot 2020-08-17 at 01 43 04
And action on WorkloadController can still be executed successfully.

---
2. As for performance improvement, used [pprof](https://golang.org/pkg/net/http/pprof/) tool to profile CPU performance by running kubeturbo of cluster `K8s 1.18.5 in DC11` locally

**Environment**:
macOS
CPU: 2.5 GHz Intel Core i7

Before and after changes run such command:
```
go tool pprof http://127.0.0.1:10265/debug/pprof/profile?seconds=10
```
and then trigger a rediscovery to profile the runtime of each functions during discovery.

**Evidence**:
By repeating the above steps multiple times,
- before the change, `ClusterMonitor::Do` func runs around 30ms, where getPodOwner takes almost the whole 30ms:
```
      flat  flat%   sum%        cum   cum%
         0     0%   100%       30ms  7.32%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).Do
         0     0%   100%       30ms  7.32%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).RetrieveClusterStat
         0     0%   100%       30ms  7.32%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).findNodeStates
         0     0%   100%       30ms  7.32%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).genNodePodsMetrics
         0     0%   100%       30ms  7.32%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).genNodeResourceMetrics
         0     0%   100%       30ms  7.32%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).getPodOwner
```

- after the change, `ClusterMonitor::Do` func runs much faster with no more than 10 ms:
```
      flat  flat%   sum%        cum   cum%
         0     0%   100%       10ms  3.12%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).Do
         0     0%   100%       10ms  3.12%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).RetrieveClusterStat
         0     0%   100%       10ms  3.12%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).findNodeStates
         0     0%   100%       10ms  3.12%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).genNodePodsMetrics
         0     0%   100%       10ms  3.12%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).genNodeResourceMetrics
         0     0%   100%       10ms  3.12%  github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/master.(*ClusterMonitor).genOwnerMetrics
```
where `getPodOwner` func runs much faster than 10ms (looks the functions less than 10 ms are not shown in the profiling results)

### DIFF
--- a/pkg/k8s_tap_service.go
+++ b/pkg/k8s_tap_service.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/turbonomic/kubeturbo/pkg/action"
+	"github.com/turbonomic/kubeturbo/pkg/cluster"
 	"github.com/turbonomic/kubeturbo/pkg/discovery"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/configs"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/detectors"
@@ -125,7 +126,8 @@ func createProbeConfigOrDie(c *Config) *configs.ProbeConfig {
 	kubeletMonitoringConfig := kubelet.NewKubeletMonitorConfig(c.KubeletClient, c.KubeClient)
 
 	// Create cluster monitoring
-	masterMonitoringConfig := master.NewClusterMonitorConfig(c.KubeClient, c.DynamicClient)
+	clusterScraper := cluster.NewClusterScraper(c.KubeClient, c.DynamicClient)
+	masterMonitoringConfig := master.NewClusterMonitorConfig(clusterScraper)
 
 	// TODO for now kubelet is the only monitoring source. As we have more sources, we should choose what to be added into the slice here.
 	monitoringConfigs := []monitoring.MonitorWorkerConfig{
@@ -136,8 +138,7 @@ func createProbeConfigOrDie(c *Config) *configs.ProbeConfig {
 	probeConfig := &configs.ProbeConfig{
 		StitchingPropertyType: c.StitchingPropType,
 		MonitoringConfigs:     monitoringConfigs,
-		ClusterClient:         c.KubeClient,
-		DynamicClient:         c.DynamicClient,
+		ClusterScraper:        clusterScraper,
 		NodeClient:            c.KubeletClient,
 	}
 
@@ -162,8 +163,8 @@ func NewKubernetesTAPService(config *Config) (*K8sTAPService, error) {
 		config.containerUsageDataAggStrategy, config.ORMClient, config.DiscoveryWorkers, config.DiscoveryTimeoutSec,
 		config.DiscoverySamples, config.DiscoverySampleIntervalSec)
 
-	actionHandlerConfig := action.NewActionHandlerConfig(config.CAPINamespace, config.CAClient, config.KubeClient,
-		config.KubeletClient, config.DynamicClient, config.SccSupport, config.ORMClient)
+	actionHandlerConfig := action.NewActionHandlerConfig(config.CAPINamespace, config.CAClient, config.KubeletClient,
+		probeConfig.ClusterScraper, config.SccSupport, config.ORMClient)
 
 	// Kubernetes Probe Registration Client
 	registrationClient := registration.NewK8sRegistrationClient(registrationClientConfig, config.tapSpec.K8sTargetConfig)
--- a/pkg/action/action_handler.go
+++ b/pkg/action/action_handler.go
@@ -2,7 +2,6 @@ package action
 
 import (
 	"fmt"
-	"github.com/turbonomic/kubeturbo/pkg/resourcemapping"
 	"strings"
 	"time"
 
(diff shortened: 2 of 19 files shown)


pkg/k8s_tap_service.go
```diff
@@ -125,7 +126,8 @@ func createProbeConfigOrDie(c *Config) *configs.ProbeConfig {
 	kubeletMonitoringConfig := kubelet.NewKubeletMonitorConfig(c.KubeletClient, c.KubeClient)
 
 	// Create cluster monitoring
-	masterMonitoringConfig := master.NewClusterMonitorConfig(c.KubeClient, c.DynamicClient)
+	clusterScraper := cluster.NewClusterScraper(c.KubeClient, c.DynamicClient)
```
irfanurrehman, line 129: I like that you are using cluster scraper to carry the cache and store the pod's parent info.
I however think that we can use a lighter version of the same. That would mean that we can break the cluster scraper into two parts:
`PodInfoGetter`  which implements GetPodGrandparentInfo() or any other similar utility that is needed by some other consumer, for example what I have alluded in [this PRs description, point 2](https://github.com/turbonomic/kubeturbo/pull/468#issue-468579603).
`ClusterScraper` which can nest this in (only if needed). 
Then pass only `PodInfoGetter` to k8s_controller_updater` or to whichever consumer that needs only that functionality.
Passing the whole `ClusterScrapper` to consumers which need only pod related functionality will carry that little bit of extra memory in terms of remaining method signatures.

Having said this, its not necessary to be done as part of this PR and I would be more then happy to take it up as part of #468.

irfanurrehman:
Good work @marvelyue. 
lgtm

pkg/k8s_tap_service.go
marvelyue: Thanks @irfanurrehman! I get your point. To get pod info, we just need the `cache` and `dynamicClient` in this `ClusterScraper`. And there's an extra `Clientset` in `ClusterScraper` which is used to get cluster resources.

Actually by taking a further look at the code, in current logic I think it would be simpler to just have one `ClusterScraper` to carry all the objects. 

First before the changes, there were 2 places to initiate `ClusterScraper` in [k8s_discovery_client](https://github.com/turbonomic/kubeturbo/pull/469/files#diff-a529c27731d7f231654165f9cc0674d7L102) and [config](https://github.com/turbonomic/kubeturbo/pull/469/files#diff-be0cc81aebf545c5a85e77bbc5a79fecL16). Now in this PR, I create the `ClusterScraper` only once here and the usage of `ClusterScraper` in each consumer is just a pointer so it actually helps reduce memory consumption. Then almost all the places where `ClusterScraper` is passed as input argument need the `Clientset` and `dynamicClient`. 

And in the changes you are making in #468, we could move the `getParentAndGrandParentInfo` func to `cluster_info_scraper` as well and pass`clusterScraper` as input to replace `clientset` and `dynClient` in this [movePod func](https://github.com/turbonomic/kubeturbo/pull/468/files#diff-623e1452732b5bf67c1efd4dd5ebb66aR93). So maybe we don't even need a new `PodInfoGetter`.

PR #471, opened 2020-09-07
Add volume to vm relationship to show connection in unstitched env

irfanurrehman:
This partly fixes https://vmturbo.atlassian.net/browse/OM-62195.
The commodity can also be stitched to reflect the used value against those reported for the virtual machine too. It does not work though. So if unstitched the vm will report right values, but after stitching vm shows used same as capacity. I think the stitching is useful and will try to figure out some way of doing that.

### DIFF
--- a/pkg/discovery/dtofactory/node_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/node_entity_dto_builder.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/turbonomic/kubeturbo/pkg/discovery/dtofactory/property"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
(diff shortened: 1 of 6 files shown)


irfanurrehman:
Rebased!
I am waiting for review on this one.
cc @ading1977 @enlinxu @marvelyue

enlinxu:
Discussed with @irfanurrehman , this review will help populating the VM storage amount bought when the stitching is working. Right now we still need to figure out how to make the stitching work, due to the key inconsistency. Will re-visit once we can make the stitching work.

enlinxu:
@irfanurrehman  based on our discussion, can we close this PR?

irfanurrehman:
ok @enlinxu closing this for now. I will get back to this again when I can find a way to stitch the `STORAGE_AMOUNT` commodity on VMs.

enlinxu:
@irfanurrehman I thought we decided to fix this through supply chain navigation and you already had code for that, is that not the case?

PR #473, opened 2020-09-11
Fix error handling while discovering volumes

irfanurrehman:
Fix for https://vmturbo.atlassian.net/browse/OM-62546

### DIFF
--- a/pkg/discovery/dtofactory/volume_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/volume_entity_dto_builder.go
@@ -33,11 +33,14 @@ func (builder *volumeEntityDTOBuilder) BuildEntityDTOs(volToPodsMap map[*api.Per
 		entityDTOBuilder.DisplayName(displayName)
 
 		commoditiesSold, cap, err := builder.getVolumeCommoditiesSold(vol, podVolumes)
+		// A pod can have multiple volumes and we might process one even when
+		// we see an error on another
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ading1977 (approved):

enlinxu (approved):

pkg/discovery/dtofactory/volume_entity_dto_builder.go
```diff
@@ -103,7 +106,7 @@ func (builder *volumeEntityDTOBuilder) getVolumeCommoditiesSold(vol *api.Persist
 		commBuilder.Capacity(capacity)
 		commodityPerPod, err := commBuilder.Create()
 		if err != nil {
-			return nil, volumeCapacity, err
+			return commoditiesSold, volumeCapacity, err
 		}
 
 		// TODO(irfanurrehman): Set resisable depending on node properties
```
marvelyue, line 112: Nit: Although it's less likely an error could occur here, would it be better to just skip appending this commoditySold if there's error here and continue on the next podVol? Something like:
```
if err != nil {
    glog.Errorf("Error creating commoditySold by volume %s: %v ", podVol.MountName, err)
} else {
    // TODO(irfanurrehman): Set resisable depending on node properties
    commoditiesSold = append(commoditiesSold, commodityPerPod)
}
```

marvelyue (approved):

pkg/discovery/dtofactory/volume_entity_dto_builder.go
irfanurrehman: Thanks @marvelyue. Updated!

PR #475, opened 2020-09-15
Skip adding those volumes for which there are no stats

irfanurrehman:
This change skips creation of volumes if the volume metrics are not discovered in the cluster. This avoids having any dangling volumes.


esara (approved):
LGTM

PR #476, opened 2020-09-18
Feature gates

irfanurrehman:
This adds a feature gating mechanism to kubeturbo.
Additionally keeps the persistent volumes feature gated, but enabled by default.
I have used the config in place of a command line flag as I personally feel that the command line flags are overflowing. We should probably think of moving more flags to config.

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -11,7 +11,6 @@ import (
 	"syscall"
 
 	agg "github.com/turbonomic/kubeturbo/pkg/discovery/worker/aggregation"
-	"github.com/turbonomic/kubeturbo/pkg/features"
 	"github.com/turbonomic/kubeturbo/pkg/resourcemapping"
 
 	clusterclient "github.com/openshift/cluster-api/pkg/client/clientset_generated/clientset"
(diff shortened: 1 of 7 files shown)


irfanurrehman:
@evat-pm This will need additional section in the kubeturbo config as below:
```
      "featureGates": {
        "features": [
          {
            "name": "PersistentVolumes",
            "configuration": "Disabled"
          }
        ]
      }
```
Please let me know where exactly to update this to be useful. 
Should I add this configuration to the helm chart too?

ading1977:
> @evat-pm to disable (or enable a feature disabled by default) an additional section in the kubeturbo config will be needed as below:
> 
> ```
>       "featureGates": {
>         "features": [
>           {
>             "name": "PersistentVolumes",
>             "configuration": "Disabled"
>           }
>         ]
>       }
> ```
> 
> Please let me know where exactly to update this to be useful.
> Should I add this configuration to the helm charts too?

What are the main use cases for the feature gates? If I understand correctly, we only need to **explicitly** call out a feature if we want to disable it, so maybe the following wording is more intuitive?

```
         "features": [
           {
             "name": "PersistentVolumes",
             "disabled": "true"
           }
```
Or maybe even simpler:
```
        "disabledFeatures": ["PersistentVolumes"]
```

irfanurrehman:
> What are the main use cases for the feature gates? If I understand correctly, we only need to **explicitly** call out a feature if we want to disable it? If so, maybe the following wording is more intuitive?
> 
> ```
>          "features": [
>            {
>              "name": "PersistentVolumes",
>              "disabled": "true"
>            }
> ```
> 
> Or maybe even simpler:
> 
> ```
>         "disabledFeatures": ["PersistentVolumes"]
> ```

Thanks @ading1977 for taking a look at this.
I have updated the config as below:
```
      "featureGates": {
        "disabledFeatures": ["PersistentVolumes"]
      }
```
PTAL.

irfanurrehman:
@enlinxu @ading1977 Does this look like a feature to keep? I ask because we have had some features after this and we sort of did not use gating for them?
May be keep it to have an option to use, in case we have a feature in future which might be tricky (or alpha) so that if needed the gating coule be used.

enlinxu (approved):
LGTM, let's rebase and merge

PR #477, opened 2020-09-21
Copy pod spec from parent in pod move actions

irfanurrehman:
Solves https://vmturbo.atlassian.net/browse/OM-61699.
Additionally this copies podSpec always from the parents `spec/template/spec` into the new pods spec, as the pods are supposed to be created from the parents podSpec anyways.

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -26,7 +26,7 @@ import (
 //TODO: if pod is from controller, then copy pod in the way as
 // kubernetes/pkg/controller/controller_utils.go#GetPodFromTemplate
 // https://github.com/kubernetes/kubernetes/blob/0c7e7ae1d9cccd0cca7313ee5a8ae3c313b72139/pkg/controller/controller_utils.go#L553
-func copyPodInfo(oldPod, newPod *api.Pod) {
+func copyPodInfo(oldPod, newPod *api.Pod, copySpec bool) {
 	//1. typeMeta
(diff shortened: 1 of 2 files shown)


enlinxu (approved):
LGTM, Let's have Meng or Yue also taking a look

ading1977 (approved):
LGTM for the move use case.

However this probably breaks the resize of single container: `clonePodWithNewSize`, could you please double check? I am not sure if we should still support resize of single container use case, if not, we should probably disable it in the code.

irfanurrehman:
> However this probably breaks the resize of single container: `clonePodWithNewSize`, could you please double check? I am not sure if we should still support resize of single container use case, if not, we should probably disable it in the code.

@ading1977 Do you mean we need to do something similar for individual pod resizes too? We can and I think we do need single pod resizes for daemonsets (The solution to daemonsets is not complete yet). Although I think the single pod resize and the pod move solution in this PR are orthogonal and I would probably tream them as such.

irfanurrehman:
@ading1977 updated this to ensure the single container resize does not break.

ading1977:
Discussed with @irfanurrehman, with the latest change, resize single pod will not copy spec from parent (if exists), it only makes sure that the existing functionality does not break. In addition, there are some known issues right now with regard to the interaction of daemonset pod resize and merged actions, which @irfanurrehman is trying to address.

irfanurrehman:
Thanks @ading1977 for taking a look. So this is good to be merged as of now. I have rebased on latest.

PR #478, opened 2020-09-22
Fix failed action merge execution

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-62724.
The original pod did not consider possibility of pods which won't have a valid nodename assigned, for example pending pods.


ading1977 (approved):
LGTM

marvelyue (approved):

PR #479, opened 2020-09-23
Replicaset move action fixed

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-62758


enlinxu (approved):
LGTM

ading1977 (approved):
So the problem is that getting the member pods of a controller via labels of that controller may bring in pods from other controller if the labels are misconfigured.

All these additional logic is due to the need to convert between MHz and millicores. When we properly support millicores, these logic won't be necessary anymore.

PR #480, opened 2020-09-23
Correct aggregated commodity capacity on ContainerSpec.

marvelyue:
**Intent**:
This is to fix part of the issue in JIRA: https://vmturbo.atlassian.net/browse/OM-62598.

**Problem and Diagnosis**:
CPU commodity capacity and usage are converted from milli-cores to MHz based on node CPU frequency. When there are multiple container replicas belonging to the same container spec and those container replicas are on different nodes with different CPU frequency, aggregated CPU commodity capacity on container spec varies from different discoveries.

This is because in previous logic, when aggregating CPU commodity capacity, we always use the resource capacity from the first container. And order is not guaranteed for container replicas of the same container spec, so CPU commodity capacity could be different from different discoveries for the same container spec.

**Solution**:
1. Instead of using the resource capacity from the first container, collect resource capacity value of each individual container instance in `ContainerMetrics` of `ContainerSpecMetrics` in `container_spec_metrics_collector.go`.  Note that in previous logic, there was:
```
containerMetrics.Used = append(containerMetrics.Used, containerResourceMetrics.Used...)
```
We don't actually need to append used data points to `containerMetrics` of `containerSpecMetric` because it just simply collects each individual container capacity and used data points data here.

2. In `container_spec_discovery_worker.go`, append capacity values from container replicas belonging to the same container spec discovered from different nodes.

3. In `container_usage_data_aggregator.go` and `container_utilization_data_aggregator.go`, use max of capacity values from all container replicas as the aggregated capacity value on container spec. Max is better than average here because when any container is rescheduled on a different node with different CPU frequency, average CPU capacity value would be changed. Note that it's still not guaranteed CPU capacity will always be consistent because max capacity could still be changed if container is rescheduled on a different node with larger CPU frequency. Using max value here will be more consistent. 

4. Updated unit tests.

**Testing Done**:
It's not easy to reproduce the scenario where different container replicas are on different nodes with different CPU frequency.
Make sure the changes didn't break anything. ContainerSpec commodity used, peak and utilizationData are set properly as before. Commodity capacity values in discovery entity DTOs are consistent in different discovery cycles. Here's discovery entity dto data of a container spec with 2 container replicas after the changes, where capacity value is consistent after multiple discoveries:
```
entityDTO {
  entityType: CONTAINER_SPEC
  id: "dae49cbe-f80f-4ee6-b61a-af1b48810c5c/cpu-mem-load-up"
  displayName: "cpu-mem-load-up"
  commoditiesSold {
    commodityType: VCPU
    used: 2089.6981156081642
    capacity: 2131.0224
    peak: 2117.78390282004
    active: true
    resizable: true
    utilizationData {
      point: 98.502996375
      point: 98.722195625
      point: 98.2995105
      point: 97.139004625
      point: 97.66716675
      point: 99.37877250000001
      point: 98.64722525
      point: 95.604298
      point: 98.651853
      point: 98.32470962500001
      point: 98.345640625
      point: 99.20541625
      point: 98.78926650000001
      point: 97.92553900000001
      point: 98.77287137500001
      point: 97.31368387500001
      point: 97.82096825
      point: 96.17998725
      point: 98.020709
      point: 97.90465675
      lastPointTimestampMs: 1600830177582
      intervalMs: 28396
    }
  }
```

### DIFF
--- a/pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
+++ b/pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
@@ -38,9 +38,16 @@ func (avgUsageDataAggregator *avgUsageDataAggregator) String() string {
 }
 
 func (avgUsageDataAggregator *avgUsageDataAggregator) Aggregate(resourceMetrics *repository.ContainerMetrics) (float64, float64, float64, error) {
-	if len(resourceMetrics.Used) == 0 {
-		err := fmt.Errorf("error to aggregate container usage data using %s: used data points list is empty", avgUsageDataAggregator)
+	if len(resourceMetrics.Used) == 0 || len(resourceMetrics.Capacity) == 0 {
+		err := fmt.Errorf("error to aggregate container usage data using %s: used or capacity data points list is empty", avgUsageDataAggregator)
 		return 0.0, 0.0, 0.0, err
+	}
+	// Use the max of resource capacity values from all container replicas.
+	// CPU capacity has been converted from milli-cores to MHz. This will make the aggregated CPU capacity on container
+	// spec more consistent if container replicas are on different nodes with different CPU frequency.
+	capacity := 0.0
+	for _, capVal := range resourceMetrics.Capacity {
+		capacity = math.Max(capacity, capVal)
 	}
 	usedSum := 0.0
 	peak := 0.0
@@ -49,7 +56,7 @@ func (avgUsageDataAggregator *avgUsageDataAggregator) Aggregate(resourceMetrics 
 		peak = math.Max(peak, usedPoint.Value)
 	}
 	avgUsed := usedSum / float64(len(resourceMetrics.Used))
-	return resourceMetrics.Capacity, avgUsed, peak, nil
+	return capacity, avgUsed, peak, nil
 }
 
 // ---------------- Max usage data aggregation strategy ----------------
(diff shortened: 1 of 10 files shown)


pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
```diff
@@ -38,18 +38,25 @@ func (avgUsageDataAggregator *avgUsageDataAggregator) String() string {
 }
 
 func (avgUsageDataAggregator *avgUsageDataAggregator) Aggregate(resourceMetrics *repository.ContainerMetrics) (float64, float64, float64, error) {
-	if len(resourceMetrics.Used) == 0 {
-		err := fmt.Errorf("error to aggregate container usage data using %s: used data points list is empty", avgUsageDataAggregator)
+	if len(resourceMetrics.Used) == 0 || len(resourceMetrics.Capacity) == 0 {
+		err := fmt.Errorf("error to aggregate container usage data using %s: used or capacity data points list is empty", avgUsageDataAggregator)
```
irfanurrehman, line 42: nit: `error aggregating` or `error while aggregating`

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
```diff
@@ -62,13 +69,20 @@ func (maxUsageDataAggregator *maxUsageDataAggregator) String() string {
 }
 
 func (maxUsageDataAggregator *maxUsageDataAggregator) Aggregate(resourceMetrics *repository.ContainerMetrics) (float64, float64, float64, error) {
-	if len(resourceMetrics.Used) == 0 {
-		err := fmt.Errorf("error to aggregate container usage data using %s: used data points list is empty", maxUsageDataAggregator)
+	if len(resourceMetrics.Used) == 0 || len(resourceMetrics.Capacity) == 0 {
+		err := fmt.Errorf("error to aggregate container usage data using %s: used or capacity data points list is empty", maxUsageDataAggregator)
```
irfanurrehman, line 73: Nit: as previous comment.
Suggestion (can be ignored here if you have more changes in pipeline): the common code within all aggregators can go into a common function.

pkg/discovery/worker/aggregation/container_utilization_data_aggregator.go
```diff
@@ -39,11 +39,17 @@ func (allDataAggregator *allUtilizationDataAggregator) String() string {
...
 
 func (allDataAggregator *allUtilizationDataAggregator) Aggregate(resourceMetrics *repository.ContainerMetrics) ([]float64, int64, int32, error) {
-	if len(resourceMetrics.Used) == 0 {
-		err := fmt.Errorf("error aggregating container utilization data using %s: used data points list is empty", allDataAggregator)
+	if len(resourceMetrics.Used) == 0 || len(resourceMetrics.Capacity) == 0 {
+		err := fmt.Errorf("error aggregating container utilization data using %s: used or capacity data points list is empty", allDataAggregator)
 		return []float64{}, 0, 0, err
 	}
-	capacity := resourceMetrics.Capacity
+	// Use the max of resource capacity values from all container replicas.
```
irfanurrehman, line 46: suggestion: duplicate code suggestion as above. I think there are 4 instances.

irfanurrehman:
LGTM, except the few nits/suggestions.
Thanks @marvelyue.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
marvelyue: Updated.

pkg/discovery/worker/aggregation/container_usage_data_aggregator.go
marvelyue: Good point. Thanks. Created common functions

ading1977 (approved):
LGTM

PR #483, opened 2020-10-07
Make storage amount commodity provider optional for pods

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-62722


enlinxu (approved):
LGTM

PR #484, opened 2020-10-09
Fallback to the API Server node proxy path if kubelet call fails

irfanurrehman:
This implements falling back on to the nodes proxy endpoint via API server if the kubelet call to the node IP fails.
Each of this fallback will log a message at loglevel 2.
Too many of these messages would mean a global problem in the environment, in which case the flag to use the proxy endpoint always can be turned on.


ading1977 (approved):
LGTM.

PR #485, opened 2020-10-12
Discovery tests

irfanurrehman:
This submits a test which compares total number of entities and groups discovered.
There is a bunch of more items that I could identify while writing this and are captured in TODOs. I will continue adding test as per those items.

This is on top of https://github.com/turbonomic/kubeturbo/pull/484 as the discovery needs to be external to the cluster.


irfanurrehman:
@enlinxu @ading1977 @marvelyue I would appreciate if you can please take a look.

enlinxu (approved):
Sorry for the delay @irfanurrehman , really nice and let's extend it!

enlinxu:
@irfanurrehman  Shall we merge ?

irfanurrehman:
Thanks for the review @enlinxu

PR #486, opened 2020-10-15
Add legacy support of cpu and memory as keys in quota spec

irfanurrehman:
This fixes https://vmturbo.atlassian.net/browse/OM-63295

Verified the fix using scenario listed at https://github.com/turbonomic/Turbo-ScenarioAsCode/tree/master/namespaceQuota

Using [scenario4](https://github.com/turbonomic/Turbo-ScenarioAsCode/tree/master/namespaceQuota/namespaceQuota4CPU)
and creating a quota resource with below spec:
```apiVersion: v1
kind: ResourceQuota
metadata:
  name: nsquota-test-cpu-multi-deployments
  namespace: nsquota-test-cpu-multi-deployments
spec:
  hard:
    cpu: "30m"
    memory: 100Mi
    limits.cpu: "30m"
    limits.memory: 150Mi
```
request capacities for namespace (which correspond to request.cpu and request.memory) are properly discovered:
image

### DIFF
--- a/pkg/discovery/metrics/metric.go
+++ b/pkg/discovery/metrics/metric.go
@@ -53,7 +53,13 @@ var (
 	}
 
 	// Mapping of Kubernetes API Server resource names to the allocation resource types
+	// A resource quota definition, although undocumented, honours both "cpu" and "requests.cpu"
+	// as keys for cpu request quota and likewise for memory.
+	// This has more to do with legacy support, where when the resourcequota proposal was introduced
+	// had support only for "cpu" and "memory" corresponding to limiting only requests.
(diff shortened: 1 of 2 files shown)


irfanurrehman:
@marvelyue @ading1977  this looked way too simple to update, however there are many resource maps we create with predefined string keys with the strings coming from both k8s defined ones and/or those that we have defined in metrics.go, which I personally find confusing. This thus needs extra diligence... :-)

enlinxu (approved):
LGTM

ading1977 (approved):
LGTM.

PR #494, opened 2020-11-06
Fix deploymentconfig handling in merged actions

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-63955


pkg/action/executor/workload_controller_resizer.go
```diff
@@ -166,7 +173,10 @@ func (r *WorkloadControllerResizer) getChildPod(parentKind, namespace, name stri
 				// We still try rest of the pods
 				continue
 			}
-			if match {
+
+			// isOpenShiftDeployerPod is needed to exclude pods which openshift uses to manage
+			// deployment configs. We should not select the deployer pods.
+			if match && !r.isOpenShiftDeployerPod(pod.Labels) {
```
enlinxu, line 179: We shouldn't do it at the action execution level. If we don't want to execute actions on the OpenShift deployed pod, we should mark them controllable false (and daemon true) during the discovery.

irfanurrehman: @enlinxu This is not to exclude pods on which we execute the action. The action result spec (updated resources) will be set on the parent, the deployment config in this case. While executing the merged action we only get the parent info and the resource values in the action spec. We but additionally need the node cpufrequency for the value conversions. To get the node cpu frequency we get one of the child pods and then get the node details from that child pod. [This comment](https://github.com/turbonomic/kubeturbo/blob/master/pkg/action/executor/workload_controller_resizer.go#L41) explains the same.

Now with deployment config we additionally need to exclude the deployer pod (which runs like a job and then exits, but the pod resource remains behind) from being selected for above.

enlinxu: I see, thanks for the explanation!

enlinxu (approved):

PR #495, opened 2020-11-10
Fix deploymentconfig handling in merged actions for 7.22.7

irfanurrehman:
This pushes a fix for https://vmturbo.atlassian.net/browse/OM-63955 to `7.22.7`.
This includes few more fixes, which the deployment config fix is based on (and certainly good to have).

cc @enlinxu @evat-pm

### DIFF
--- a/pkg/action/executor/workload_controller_resizer.go
+++ b/pkg/action/executor/workload_controller_resizer.go
@@ -153,16 +153,23 @@ func (r *WorkloadControllerResizer) getChildPod(parentKind, namespace, name stri
 	if err != nil {
 		return nil, err
 	}
+
+	noPodFoundError := fmt.Errorf("could not find any matching pod for %s %s/%s", parentKind, namespace, name)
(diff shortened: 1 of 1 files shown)


irfanurrehman:
I have verified the deployment config merged resize  working against the latest (`8.0.3-SNAPSHOT`) XL.
@enlinxu I don't see a reason, but do you think I should verify this against a `7.22.7` instance also?

enlinxu (approved):

enlinxu:
@irfanurrehman it will be great if you can have a quick sanity check over 7.22.7 server. Not mandatory though since I don't remember anything changed in between that may break this functionality.

irfanurrehman:
> @irfanurrehman it will be great if you can have a quick sanity check over 7.22.7 server. Not mandatory though since I don't remember anything changed in between that may break this functionality.

@enlinxu verified working on `7.22.7`
image


------------


image


------------


image

--------------


image

PR #499, opened 2020-11-18
Flipping the relationship between ContainerCluster and VM to avoid a single pod being tied to all VMs in the cluster

chlam4:
More specifically, this is to flip the relationship between ContainerCluster and VM to fix some unwanted supply chain traversals such as a single pod can go around the whole circle thru the ContainerCluster to get tied with all VMs in the cluster; this is now an "AggregatedBy" relationship.  See this JIRA for more about the issue: https://vmturbo.atlassian.net/browse/OM-64295.

With this change, the supply chains starting from a pod or a workload controller or a namespace will be back to previously before the introduction of the container cluster, i.e. they will only be linked to the relevant VMs, not all VMs in the cluster - please see the screenshots below.
From a ContainerPod:
from-container-pod
From a WorkloadController:
from-workload-controller
From a Namespace:
from-namespace

Conversely, the supply chain starting from a VM will contain all pods running there:
from-vm

### DIFF
--- a/pkg/discovery/dtofactory/cluster_dto_builder.go
+++ b/pkg/discovery/dtofactory/cluster_dto_builder.go
@@ -80,11 +80,6 @@ func (builder *clusterDTOBuilder) BuildEntity(entityDTOs []*proto.EntityDTO) (*p
 
 	entityDTOBuilder.WithPowerState(proto.EntityDTO_POWERED_ON)
 
-	// Link member nodes as connected entities
-	for _, node := range builder.cluster.NodeList {
-		entityDTOBuilder.ConnectedTo(string(node.UID))
-	}
-
 	// build entityDTO.
 	entityDto, err := entityDTOBuilder.Create()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


enlinxu:
The change is bigger than necessary to fix the issue. Let's limit the change as much as possible for 8.0.3 release and include the additional change in 8.0.4 together with the supply chain change.

enlinxu (approved):
LGTM

esara (approved):

chlam4:
> The change is bigger than necessary to fix the issue. Let's limit the change as much as possible for 8.0.3 release and include the additional change in 8.0.4 together with the supply chain change.

Agreed; those additional changes have just been taken out.

ading1977 (approved):

marvelyue:
Hi @chlam4 , quick question, without setting up `aggregatedBy` from namespace to container cluster in your initial commit, when scoping to a VM, namespaces are still shown correctly in the supply chain right (namespaces from the pods running on the scoped VM)?

marvelyue (approved):

evat-pm:
@chlam4 can you please share with me what is in 8.0.3 vs 8.0.4 for what the user will see.  I cannot tell from the comments what will be different in 8.0.4 that the user will see change.  I will ping you on slack.

PR #500, opened 2020-11-18
Adding aggregated-by relationship from container pod to workload controller to namespace to cluster to facilitate desired supply chain traversals

chlam4:
This PR is adding the AggregatedBy relationship to the followings to facilitate desired supply chain traversals without special rules (requiring corresponding turbo server side changes which will be in 8.0.4):
- From ContainerPod to WorkloadController
- From WorkloadController to Namespace
- From Namespace to ContainerCluster

With the server side changes, this will address the following two JIRAs:
- https://vmturbo.atlassian.net/browse/OM-64005
- https://vmturbo.atlassian.net/browse/OM-64087

The ContainerCluster supply chain will now look like this
from-container-cluster
The ContainerCluster will now be reached from a Container or a ContainerSpec or a Container Volume:
From a Container:
from-container
From a ContainerSpec:
from-container-spec
From a container Volume:
from-container-volume


dblinn (approved):
LGTM although if there is a simple way to unit test the new relationship that might be nice too.

marvelyue (approved):
LGTM along with platform side changes

chlam4:
> LGTM although if there is a simple way to unit test the new relationship that might be nice too.

Good point @dblinn.  I've added corresponding unit tests in workload controller and namespace builders, which is pretty straightforward since those test classes are pretty well established.  May need some work to add in the pod entity builder; I will see what I can do.

PR #506, opened 2020-11-30
Use endian reversed ids additionally for VM stiching wherever possible

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-64293.
I have also added reversed strings whereever we can.
I have verified this with k8s clusters on aws, azure and vsphere in our environments.


ading1977 (commented):
Looks fine to me. Just to confirm, this is to append the reversed ID to the original ID as the stitching attribute, right?

Also, please fix the unit test.

ading1977 (approved):

esara (approved):

PR #510, opened 2020-12-10
Use spec from the workload controller for merged actions

irfanurrehman:
As there is no need to have the nodes cpu frequency for merged actions after #496, we don't need to have the complicated logic of finding out a pod from the workload controller and land into issues like https://vmturbo.atlassian.net/browse/OM-64853.

~~Marking it as WIP as there are many different scenarios that need test after this change and I could not test them yet.~~

### DIFF
--- a/pkg/action/executor/workload_controller_resizer.go
+++ b/pkg/action/executor/workload_controller_resizer.go
@@ -7,9 +7,7 @@ import (
 	k8sapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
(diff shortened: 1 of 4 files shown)


irfanurrehman:
Verified this to be working fine for resizes with a deployment and a deployment config.

ading1977 (approved):
Look good to me. 
Maybe wait until @marvelyue has reviewed as he is more familiar with this part of logic.

enlinxu (approved):

pkg/action/executor/workload_controller_resizer.go
```diff
@@ -282,13 +170,14 @@ func resizeWorkloadController(clusterScraper *cluster.ClusterScraper, ormClient
...
+func getContainerIndex(podSpec *k8sapi.PodSpec, containerName string) int {
 	// We assume that the pod spec is valid.
-	for i, cont := range pod.Spec.Containers {
+	for i, cont := range podSpec.Containers {
 		if cont.Name == containerName {
 			return i
 		}
 	}
 
+	glog.V(4).Infof("Match for container %s not found in pod template spec : %v", containerName, podSpec)
```
marvelyue, line 181: Do we want to show such message as Error instead of hiding it in V4 info? Looks normally it's not expected to miss the target container in pod template spec.

PR #512, opened 2020-12-28
Update dependencies after bootstrapping cluster autoscaler code usage

irfanurrehman:
These are the main updates:
    1. Updates k8s core and its subrepositories to 1.19.1.
    2. Bootstraps the cluster autoscaler dependency and pulls in cloudprovider
    packages from https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.19.1. This should ideally map to k/k 1.19.1.
    3. Updates the API calls which have changes as per new client-go and relevant k/k updates.
    4. Changes the openshift cluster api dependencies from github.com/openshift/cluster-api to github.com/openshift/machine-api-operator.
    5. Had to stick to k/k 1.19.1 because opneshift machine-api-operator is not yet on k/client-go 1.20.x and causes problems if we vendor 1.20.x.
    6. Update the min version of the go toolchain to 1.14 in go.mod. This seem to force syncing the vendored files to the current dependencies, if the vendor folder is present, for a successful build.
    7. Fix tests.

This change is quite large and intrusive. It changes almost all dependencies and toches code all around to update API calls. I have checked discovery cycles and it seems to work all fine with all entities generated as expected. I would however still prefer to have it tested more. 
Is there some way we can run this change against Arsents e2e test suite before merginig this? @enlinxu 
I have published an image with this change as `irfdocker/kubeturbo:k8s-1.19.1`


irfanurrehman:
Edit:
   ~~ 6. Update the min version of the go toolchain to 1.14 in go.mod. This seem to force syncing the vendored files to the current dependencies, if the vendor folder is present, for a successful build.~~
Had to revert this change as the way we are using goveralls in travis does not work with it.

enlinxu (approved):
LGTM

go.mod
```diff
@@ -1,60 +1,82 @@
...
-	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.0.0-20190620090116-299a7b270edc
-	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.0.0-20190620085325-f29e2b4a4f84
-	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.0.0-20190620085942-b7f18460b210
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.0.0-20190620085809-589f994ddf7f
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.0.0-20190620085912-4acac5405ec6
-	k8s.io/kubelet => k8s.io/kubelet v0.0.0-20190620085838-f1cb295a73c9
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.0.0-20190620090156-2138f2c9de18
-	k8s.io/metrics => k8s.io/metrics v0.0.0-20190620085625-3b22d835f165
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20190620085408-1aef9010884e
+	k8s.io/api => k8s.io/api v0.19.1
```
ading1977, line 53: I do not quite understand why we need the `replace` directive here? Why can't we put everything in the `require` directive? For example, the `k8s.io/api` is already defined in the `require` directive with version v0.19.4, but is now being replaced with a different version v0.19.1, is it to make it consistent with all other `k8s.io` modules?

ading1977: Talked with @irfanurrehman, this is needed as a workaround to manage dependency for k8s and cluster autoscaler.

ading1977 (approved):
LGTM. Just need to resolve the conflict.

irfanurrehman:
Rebased and completed a complete discovery and verified that entities are discovered all right on a k8s cluster (ae-cluster2) as basic sanity.
@enlinxu @ading1977

PR #517, opened 2021-01-15
Fix volume based pod moves for deployment configs

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-65720
This also includes couple of new integration tests which can test the same.

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -260,9 +260,8 @@ func deleteInvalidPendingPods(parent *unstructured.Unstructured, podClient v1.Po
 	// Fetch the parents pods again
 	newPodList, err := parentsPods(parent, podClient)
 	if err != nil {
-		if err != nil {
-			glog.Errorf("Error getting pod list for %s: %s.", parent.GetKind(), parent.GetName())
-		}
+		glog.Errorf("Error getting pod list for %s: %s: %v.", parent.GetKind(), parent.GetName(), err)
+		return
 	}
 
 	// Find the pods which weren't there and if attributed to invalid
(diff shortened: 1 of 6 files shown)


pkg/action/executor/move_util.go
```diff
@@ -281,6 +280,24 @@ func deleteInvalidPendingPods(parent *unstructured.Unstructured, podClient v1.Po
 
 func newPendingInvalidPods(list1, list2 *api.PodList) []*api.Pod {
 	var diffList []*api.Pod
+	if list2 == nil {
```
ading1977, line 283: @irfanurrehman, sorry I wasn't able to talk to you last night, I would like to catch up with you maybe tonight to understand under what circumstances `list1` or `list2` would become nil?

But just looking at this function overall, I was wondering if it can be simplified to the following?

```
func newPendingInvalidPods(list1, list2 *api.PodList) (diffList []*api.Pod) {
	if list1 == nil {
		return
	}
	for _, pod1 := range list1.Items {
		found := false
		if list2 != nil {
			for _, pod2 := range list2.Items {
				if pod1.Name == pod2.Name {
					// pod1 from list1 also exists in list2
					found = true
				}
			}
		}
		if found == true {
			continue
		}
		if pod1.Status.Phase == api.PodPending &&
			pod1.Spec.SchedulerName == "turbo-scheduler" {
			diffList = append(diffList, &pod1)
		}
	}
	return
}
```

ading1977 (changes requested):

pkg/action/executor/move_util.go
irfanurrehman: @ading1977 thanks for the suggestion. Updated!

ading1977 (approved):

enlinxu (approved):

PR #519, opened 2021-01-22
Enhance pod move to copy annotations from parent

chlam4:
Currently when executing pod move, Kubeturbo copies the spec from the parent while the annotations are still from the pod since they're part of the metadata and not in the spec.  We've observed that pod move would fail under the following scenario:
- The pod has a sidecar injected by some webhook such as the Hashicorp vault one which injection uses an annotation to signal a successful mutation and blocks any further mutation.  See https://www.vaultproject.io/docs/platform/k8s/injector/annotations#vault-hashicorp-com-agent-inject-status
- When creating a new pod, Kubeturbo copies the entire annotations from the pod including that “agent-inject-status” field.  As such, no proper mutation/injection is done.  In the case the pod relies on the sidecar to properly start up (like needing the sidecar to pull db credentials for example), the pod will crash and the pod move action will fail.

To support this use case, we will also copy annotations from the parent instead of from the pod.  The parent's annotations do not have the extra status field, making the sidecar injection successfully and allowing the pod move action to succeed.

Tested in the failure environment and observed the move action now will now be successful with this change.  Also tested in an OpenShift environment where pods have extra annotations not coming from the parent resource (but from some OpenShift admission controller), and verified that, during our execution of pod move with this change, the new pod will still have those extra annotations injected properly.

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -482,22 +482,15 @@ func createClonePod(client *kclient.Clientset, pod *api.Pod, parent *unstructure
 		return nil, fmt.Errorf("movePod: error converting unstructured pod spec to typed pod spec for %s %s: %v", kind, parentName, err)
 	}
 	// annotations
-	annotationsUnstructured, found, err := unstructured.NestedFieldCopy(parent.Object, "metadata", "annotations")
-	if err != nil || !found {
+	annotations := make(map[string]string)
+	annotationsUnstructured, found, err := unstructured.NestedFieldCopy(parent.Object, "spec", "template", "metadata", "annotations")
+	if err != nil {
 		return nil, fmt.Errorf("movePod: error retrieving annotations from %s %s: %v", kind, parentName, err)
 	}
-	annotations := make(map[string]string)
-	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(annotationsUnstructured.(map[string]interface{}), &annotations); err != nil {
-		return nil, fmt.Errorf("movePod: error converting unstructured annotations to typed annotations for %s %s: %v", kind, parentName, err)
-	}
-	// labels
-	labelsUnstructured, found, err := unstructured.NestedFieldCopy(parent.Object, "metadata", "labels")
-	if err != nil || !found {
-		return nil, fmt.Errorf("movePod: error retrieving labels from %s %s: %v", kind, parentName, err)
-	}
-	labels := make(map[string]string)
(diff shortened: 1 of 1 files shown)


chlam4:
Closing this per discussion with @irfanurrehman.  Irfan has observed hat some controllers will add annotations to pods.  If we copy the annotations from the parent, we will miss those added annotations. We need a more robust solution to solve this problem, such as Irfan suggested to whitelist some annotations to remove when doing the pod copy.

irfanurrehman (approved):
Thanks for doing this @chlam4.

irfanurrehman:
After more discussion with @chlam4, summary is that Cheuks proposal is safe for majority cases. Also it is certainly cleaner too (rather than hand curating the list of annotations we remove among the set of annotations we copy over). We can look for any future reports from customers, if ever, they complain that we have missed an annotation they wanted to be carried over to the cloned pod.

chlam4:
Thanks @irfanurrehman for the putting out our discussion comment.  I meant to do the same and put it out here, but it looks like I missed clicking "save".  Here are my comments to reopen this PR:

Reopening the PR after further discussions with @irfanurrehman with the following reasoning.

We were declining the PR/change due to the concern that we would miss extra pod annotations that are not coming from the parent.  These extra annotations can be classified into the following two categories:
1. Annotations injected by some other controllers or programs that monitor or subscribe to pod creation events upon which annotations will be injected.  This kind of extra annotations not from the parent will still be injected and won't be missed.  In fact, it's more correct and safer not to directly copy but let the other controller/program to inject them dynamically according to their logic, so that we could avoid running into the issue like with the Hashicorp vault webhook which will append an annotation AFTER some operations are done.
2. One-time injections by some program or user (e.g. via kubectl).  This kind of annotations however will not be persisted upon pod restarts.  Therefore, I'd argue they are not designed to be kept around.  If they were to be kept around, then the annotations should be injected onto the high-level resource than the pod.

Also note that the net effect of this proposed change is exactly the same as the execution of the resizing actions - the resized pods will also have the same kind of one-time annotations stripped off.  No difference.  This change makes our handling of annotations consistent between pod move and container resize actions, in addition to solving the Hashicorp vault sidecar issue.

On a related note, @irfanurrehman also brought up the new Kubernetes feature called "server side apply".  I see this feature will actually help Kubeturbo know which updates are from controllers and which are directly annotated.  So Kubeturbo can further improve from the above to even retain the one-time annotations added by users manually.

ading1977 (approved):

enlinxu (approved):

marvelyue (approved):

PR #520, opened 2021-01-27
Garbage collection

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-66008

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -54,7 +54,7 @@ const (
 	DefaultDiscoveryTimeoutSec        = 180
 	DefaultDiscoverySamples           = 10
 	DefaultDiscoverySampleIntervalSec = 60
-	DefaultGCIntervalMin              = 20
+	DefaultGCIntervalMin              = 10
 )
 
 var (
@@ -369,7 +369,7 @@ func (s *VMTServer) Run() {
 
 	gCChan := make(chan bool)
 	defer close(gCChan)
-	worker.NewGarbageCollector(kubeClient, gCChan, s.GCIntervalMin, time.Minute*30).StartCleanup()
+	worker.NewGarbageCollector(kubeClient, gCChan, s.GCIntervalMin*60, time.Minute*30).StartCleanup()
 
 	glog.V(1).Infof("********** Start running Kubeturbo Service **********")
 	// Disconnect from Turbo server when Kubeturbo is shutdown
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 	"strconv"
 	"syscall"
+	"time"
 
 	agg "github.com/turbonomic/kubeturbo/pkg/discovery/worker/aggregation"
 	"github.com/turbonomic/kubeturbo/pkg/resourcemapping"
@@ -367,14 +368,14 @@ func (s *VMTServer) Run() {
 	glog.V(2).Infof("No leader election")
 
 	gCChan := make(chan bool)
-	worker.NewGarbageCollector(kubeClient, gCChan, s.GCIntervalMin).StartCleanup()
+	defer close(gCChan)
+	worker.NewGarbageCollector(kubeClient, gCChan, s.GCIntervalMin, time.Minute*30).StartCleanup()
 
 	glog.V(1).Infof("********** Start running Kubeturbo Service **********")
 	// Disconnect from Turbo server when Kubeturbo is shutdown
 	handleExit(func() { k8sTAPService.DisconnectFromTurbo() })
 	k8sTAPService.ConnectToTurbo()
 
-	close(gCChan)
 	glog.V(1).Info("Kubeturbo service is stopped.")
 }
 
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	kubeturbo "github.com/turbonomic/kubeturbo/pkg"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/worker"
 	"github.com/turbonomic/kubeturbo/pkg/util"
 	"github.com/turbonomic/kubeturbo/test/flag"
 
@@ -52,6 +53,7 @@ const (
 	DefaultDiscoveryTimeoutSec        = 180
 	DefaultDiscoverySamples           = 10
 	DefaultDiscoverySampleIntervalSec = 60
+	DefaultGCIntervalMin              = 20
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/discovery/worker/garbage_collector.go
```diff
@@ -0,0 +1,94 @@
...
+	}()
+}
+
+func (g *GarbageCollector) cleanupLeakedPods() {
+	// Get all pods which have the gc annotation. This can include those on which an
+	// action is being executed right now.
+	listOpts := metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", executor.TurboGCLabelKey, executor.TurboGCLabelVal)}
+	podList, err := g.client.CoreV1().Pods("").List(context.TODO(), listOpts)
+	if err != nil {
+		glog.Warningf("Error getting leaked pods: %v", err)
```
chlam4, line 61: Should this be returned just to be safe?  Is the `List` function always return nil if there is an error?

pkg/discovery/worker/garbage_collector.go
```diff
@@ -0,0 +1,94 @@
...
+	if podList == nil {
+		// Nothing to clean
+		return
+	}
+
+	for _, pod := range podList.Items {
+		if g.isLeakedPod(pod) {
+			err := g.client.CoreV1().Pods(pod.Namespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+			if err != nil {
+				glog.Warningf("Encountered error trying to clean up leaked pods: %v", err)
```
chlam4, line 72: Does the error here already contain the pod info such as name and namespace?

pkg/action/executor/move_util.go
```diff
@@ -525,6 +530,8 @@ func createClonePod(client *kclient.Clientset, pod *api.Pod, parent *unstructure
 	npod.Annotations = annotations
 	// this annotation can be used for future garbage collection if action is interrupted
 	util.AddAnnotation(npod, TurboActionAnnotationKey, TurboMoveAnnotationValue)
+	// This label is used for garbage collection if a given action leaks this pod.
+	util.AddLabel(npod, TurboGCLabelKey, TurboGCLabelVal)
```
chlam4, line 534: I'm probably misunderstanding the strategy; shouldn't we also have some code too to remove the label so that properly moved pods will not be cleaned up?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -360,11 +367,15 @@ func (s *VMTServer) Run() {
...
 
+	gCChan := make(chan bool)
+	worker.NewGarbageCollector(kubeClient, gCChan, s.GCIntervalMin, time.Minute*30).StartCleanup()
+
 	glog.V(1).Infof("********** Start running Kubeturbo Service **********")
 	// Disconnect from Turbo server when Kubeturbo is shutdown
 	handleExit(func() { k8sTAPService.DisconnectFromTurbo() })
 	k8sTAPService.ConnectToTurbo()
 
+	close(gCChan)
```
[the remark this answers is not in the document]
irfanurrehman: Updated to `defer`.

pkg/discovery/worker/garbage_collector.go
irfanurrehman: As far as I know list is always nil on error. As you said to be safe on error have added a `return`.

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Probably not. Have added the pod name and ns too.

pkg/action/executor/move_util.go
irfanurrehman: We do update the labels on the cloned pod as the last step in the move function.
Ref (https://github.com/turbonomic/kubeturbo/pull/520/files#diff-db23cf9770312fe51ba93b7df47be72f93f5e1df923e4c2c1e10c68e4b0439f0R244)

irfanurrehman:
Thanks @chlam4 for taking a look.
Updated!

pkg/discovery/worker/garbage_collector.go
```diff
@@ -0,0 +1,95 @@
...
+func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
+	labels := pod.Labels
+	creationTime := pod.CreationTimestamp
+	if labels != nil {
+		gcLabelVal, gcLabelExists := labels[executor.TurboGCLabelKey]
+		// This assumes that the cluster is running a synced os time.
+		// In a remote case of a misconfigured cluster, there is a possibility of
+		// a move actions failing because of this. But this cluster misconfiguration also
+		// means that any other controller can also misbehave.
+		if gcLabelExists && gcLabelVal == executor.TurboGCLabelVal && creationTime.Add(g.podAge).Before(time.Now()) {
```
enlinxu, line 88: Shall we only count it as a leakedpod if the pod is not in running/ready state? What if kubeturbo crashed right before removing the GC label, but everything is ok, will we end up with deleting a proper running pod?

enlinxu: Discussed with Irfan, resolved the comment.

enlinxu (approved):

ading1977:
@irfanurrehman , I have some general questions:

* Looks like we are only trying to clean leaked cloned pod, as I only see cloned pod has the GC label added? What are about those potential leaked pending pods with `turbo-scheduler` scheduler?
* That brings some additional concerns that is beyond the scope of this fix: what if `kubeturbo` crashes right after we pause the rollout of a deployment, or after we disable scheduling? Will we end up with a bad deployment?

chlam4 (approved):

marvelyue:
> Looks like we are only trying to clean leaked cloned pod, as I only see cloned pod has the GC label added? What about those potential leaked pending pods with turbo-scheduler scheduler?
Hi @ading1977, initially I had the similar concern on the pods with volumes. Then I noticed `turbo-scheduler` is set on pod parents only if pods have persistent volumes, and there would always be following `defer` func called when things go wrong even with panic:
```
	defer func() {
		if !flag {
			glog.Errorf("Move pod failed, begin to delete cloned pod: %v/%v", npod.Namespace, npod.Name)
			podClient.Delete(context.TODO(), npod.Name, metav1.DeleteOptions{})
		}
		if podUsingVolume {
			// A failure can leave a pod (or pods) in pending state, delete them.
			// This can happen in case the newly created pod did fail to reach ready
			// state for whatever reason. After timeout we will end up deleting the
			// newly created pod. The invalid pod (or pods) would have been created
			// for when the parent's scheduler was set to invalid scheduler (turbo-scheduler)
			// and at exactly the same time the total replica count dropped below
			// desired count (for  whatever reason).
			deleteInvalidPendingPods(parent, podClient, podList)
		}
	}()
```
If I understand your concern correctly and if I understand `defer` correctly, looks @irfanurrehman's previous changes here are good enough to clean such invalid pending pods.

And agree that the eventual consistent actions mechanism would be an ideal solution.

marvelyue (approved):
lgtm

irfanurrehman:
@ading1977 I have added code to find and delete the pods whose scheduler might remains as `turbo-scheduler` also.
Additionally on second thoughts I have updated the cleanup timer to 10 mins from 20. It makes the worst case scenario of cleaning up the pods more closer to 30 mins.

PR #521, opened 2021-02-07
Removing namespace bought quota commodities

chlam4:
Because cluster doesn't sell them.  Also taking this chance to refactor the cluster entity builder:
- Changing to use a cluster access commodity between Namespace and Cluster instead of a keyed commodity for every compute resource
- Introducing a prefix ("cluster-") for that cluster commodity key to distinguish the one used between Pod and Node
- Skipping to accumulate access commodities from Node to Cluster @irfanurrehman (this is by assumption that the Node resource commodities aren't keyed)
- Adding some unit tests for the Cluster entity builder

### DIFF
--- a/pkg/discovery/dtofactory/cluster_dto_builder.go
+++ b/pkg/discovery/dtofactory/cluster_dto_builder.go
@@ -11,6 +11,8 @@ import (
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
 )
 
+const ClusterKeyPrefix = "cluster-"
+
 type clusterDTOBuilder struct {
 	cluster  *repository.ClusterSummary
 	targetId string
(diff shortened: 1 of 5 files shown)


dblinn (approved):
Are there any other commodities exchanged between NS and Cluster besides the new access commodity after this review? I personally prefer using movable=false over a keyed commodity to lock an entity to its current provider (because of superior performance) but the market doesn't support this in 100% of the use cases (ie https://vmturbo.atlassian.net/browse/OM-65686).

chlam4:
> Are there any other commodities exchanged between NS and Cluster besides the new access commodity after this review? I personally prefer using movable=false over a keyed commodity to lock an entity to its current provider (because of superior performance) but the market doesn't support this in 100% of the use cases (ie https://vmturbo.atlassian.net/browse/OM-65686).
@dblinn Yes, there are all the resource commodities (VCPU, VCPURequest, VMem, VMemRequest).  Prior to this change, they are keyed.  After this change, they are not keyed.

irfanurrehman (approved):
LGTM

PR #523, opened 2021-02-15
Dangling controllers cleanup at kubeturbo start

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-66590

~~This is on top of https://github.com/turbonomic/kubeturbo/pull/522. Only the last 2 commits are reviewable here.~~
#522 is merged.

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -176,9 +176,9 @@ func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName,
 				glog.V(3).Infof("Unpausing pods controller: %s for pod: %s", gPKind, podQualifiedName)
 				// We conservatively try additional 3 times in case of any failure as we don't
 				// want the parent to be left in a bad state.
-				err := commonutil.RetryDuring(defaultRetryLess, defaultRetryShortTimeout,
-					defaultRetrySleepInterval, func() error {
-						return resourceRollout(gPClient, grandParent, unpause)
+				err := commonutil.RetryDuring(DefaultRetryLess, DefaultRetryShortTimeout,
+					DefaultRetrySleepInterval, func() error {
+						return ResourceRollout(gPClient, grandParent, unpause)
 					})
 				if err != nil {
 					glog.Errorf("Move pod warning: %v", err)
@@ -187,7 +187,7 @@ func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName,
 
 			// step B2:
 			glog.V(3).Infof("Pausing pods controller: %s for pod: %s", gPKind, podQualifiedName)
-			err := resourceRollout(gPClient, grandParent, pause)
+			err := ResourceRollout(gPClient, grandParent, pause)
 			if err != nil {
 				return nil, err
 			}
@@ -205,9 +205,9 @@ func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName,
 			glog.V(3).Infof("Updating scheduler of %s for pod %s back to default.", parentKind, podQualifiedName)
 			// We conservatively try additional 3 times in case of any failure as we don't
 			// want the parent to be left in a bad state.
-			err := commonutil.RetryDuring(defaultRetryLess, defaultRetryShortTimeout,
-				defaultRetrySleepInterval, func() error {
-					return changeScheduler(pClient, parent, valid)
+			err := commonutil.RetryDuring(DefaultRetryLess, DefaultRetryShortTimeout,
+				DefaultRetrySleepInterval, func() error {
+					return ChangeScheduler(pClient, parent, valid)
 				})
 			if err != nil {
 				glog.Errorf("Move pod warning: %v", err)
@@ -216,7 +216,7 @@ func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName,
 		}()
 		// step B3:
 		glog.V(3).Infof("Updating scheduler of %s for pod %s to unknown (turbo-scheduler).", parentKind, pod.Name)
-		err = changeScheduler(pClient, parent, invalid)
+		err = ChangeScheduler(pClient, parent, invalid)
 		if err != nil {
 			return nil, err
 		}
@@ -289,8 +289,8 @@ func deleteInvalidPendingPods(parent *unstructured.Unstructured, podClient v1.Po
 	// Find the pods which weren't there and if attributed to invalid
 	// turbo-scheduler, delete them.
 	for _, pod := range newPendingInvalidPods(newPodList, podList) {
-		err := commonutil.RetryDuring(defaultRetryLess, defaultRetryShortTimeout,
-			defaultRetrySleepInterval, func() error {
+		err := commonutil.RetryDuring(DefaultRetryLess, DefaultRetryShortTimeout,
+			DefaultRetrySleepInterval, func() error {
 				return podClient.Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
 			})
 		if err != nil {
@@ -451,8 +451,9 @@ func getPodOwnersInfo(clusterScraper *cluster.ClusterScraper, pod *api.Pod,
 	return parent, nil, nsParentClient, nil, "", nil
 }
 
-// resourceRollout pauses/unpauses the rollout of a deployment or a deploymentconfig
-func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstructured, pause bool) error {
+// ResourceRollout pauses/unpauses the rollout of a deployment or a deploymentconfig
+// This additionally adds or removes the garbage collection label on the resource
+func ResourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstructured, pause bool) error {
 	kind := obj.GetKind()
 	name := obj.GetName()
 	// This takes care of conflicting updates for example by operator
@@ -462,9 +463,16 @@ func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstruc
 		if err != nil {
 			return err
 		}
+
 		if err := unstructured.SetNestedField(objCopy.Object, pause, "spec", "paused"); err != nil {
 			return err
 		}
+		if pause {
+			addGCLabel(objCopy)
+		} else {
+			removeGCLabel(objCopy)
+		}
+
 		_, err = client.Update(context.TODO(), objCopy, metav1.UpdateOptions{})
 		if err != nil {
 			return err
@@ -478,8 +486,9 @@ func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstruc
 	return nil
 }
 
-// changeScheduler sets the scheduler value to default or unsets it to an invalid one
-func changeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
+// ChangeScheduler sets the scheduler value to default or unsets it to an invalid one
+// This additionally adds or removes the garbage collection label on the resource
+func ChangeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
 	kind := obj.GetKind()
 	name := obj.GetName()
 	// This takes care of conflicting updates for example by operator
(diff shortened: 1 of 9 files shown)


irfanurrehman:
Rebased!

pkg/discovery/worker/garbage_collector.go
```diff
@@ -3,62 +3,108 @@ package worker
...
-	// Also cleanup immediately at startup
-	g.cleanupLeakedClonePods()
-	g.cleanupLeakedWrongSchedulerPods()
+func (g *GarbageCollector) StartCleanup() error {
+	glog.V(4).Info("Start garbage cleanup.")
+
+	// We cleanup the controllers before cleaning up the pods to ensure
+	// that the pending (invalid controller) pods are deleted at the right time.
+	if err := g.CleanupControllers(); err != nil {
+		return nil
```
ading1977, line 72: Why return `nil` when there is an error?

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
 
 	return false
 }
+
+// The controllers cleanup is supposed to happen only at the startup
+// We fail in case of errors to force kubeturbo to restart and try the cleanup again
+func (g *GarbageCollector) CleanupControllers() error {
```
ading1977, line 162: Suggest renaming to something like `revertControllers` or `resetControllers`, or some other name to indicate that we are reverting back to the original state.

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+	if err := g.cleanupParents(); err != nil {
```
ading1977, line 163: Suggest renaming to something like `revertScheduler`

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+		return err
+	}
+
+	if err := g.cleanupGrandParents(); err != nil {
```
ading1977, line 167: Suggest renaming to something like `unpauseRollout`

pkg/action/executor/move_util.go
```diff
@@ -478,20 +484,23 @@ func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstruc
 	return nil
 }
 
-// changeScheduler sets the scheduler value to default or unsets it to an invalid one
-func changeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
+// ChangeScheduler sets the scheduler value to default or unsets it to an invalid one
+// This additionally adds or removes the garbage collection label on the resource
+func ChangeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
```
enlinxu, line 489: Can we also split this method to two methods, i.e., ChangeScheduler and RevertScheduler ? It's clearer to me compared to using valid / invalid (nothing is invalid here)

pkg/action/executor/move_util.go
```diff
@@ -451,8 +451,9 @@ func getPodOwnersInfo(clusterScraper *cluster.ClusterScraper, pod *api.Pod,
 	return parent, nil, nsParentClient, nil, "", nil
 }
 
-// resourceRollout pauses/unpauses the rollout of a deployment or a deploymentconfig
-func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstructured, pause bool) error {
+// ResourceRollout pauses/unpauses the rollout of a deployment or a deploymentconfig
+// This additionally adds or removes the garbage collection label on the resource
+func ResourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstructured, pause bool) error {
```
enlinxu, line 456: Similarly, let's split it into pauseRollout and resumeRollout. Using a boolean here maybe more error prone.

irfanurrehman: I have a different opinion:
Its a lot of duplication of code and what we get in terms of the clarity is the `Pause` in the name of the function call in place of the name of argument. For example this is the implementation of `PauseRollout` (Which will be exact same implementation for `UnpauseRollout` except the flag value `true` in `PauseRollout`).
```
func PauseRollout(client dynamic.ResourceInterface, obj *unstructured.Unstructured) error {
	kind := obj.GetKind()
	name := obj.GetName()
	// This takes care of conflicting updates for example by operator
	// Ref: https://github.com/kubernetes/client-go/blob/master/examples/dynamic-create-update-delete-deployment/main.go
	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
		objCopy, err := client.Get(context.TODO(), name, metav1.GetOptions{})
		if err != nil {
			return err
		}

		if err := unstructured.SetNestedField(objCopy.Object, true, "spec", "paused"); err != nil {
			return err
		}
		addGCLabel(objCopy)

		_, err = client.Update(context.TODO(), objCopy, metav1.UpdateOptions{})
		if err != nil {
			return err
		}
		return nil
	})

	if retryErr != nil {
		return fmt.Errorf("error setting 'spec.paused' to true for %s %s: %v", kind, name, retryErr)
	}
	return nil
}
```

To ensure clarity in such cases, I always try to use self documenting code  as below:
```
			pause := true
			unpause := false
```
and then use these names while passing the arguments.

pkg/action/executor/move_util.go
irfanurrehman: Please see the comment below (for which I did create an example). The same applies here.
Would a better naming (invalid -> incorrect/bad/somethingelse) explain it more? 
Meanwhile the intention is indeed set a name of a scheduler to something `invalid` or `bad`. That means if the cluster for some advanced reasons also has a scheduler named `turbo-scheduler`, we will need to use some other `invalid` name in place. `ChangeScheduler` and `RevertScheduler` does not make this explicit, rather might make it sound that we are using another scheduler to the reader of the code without context.

pkg/action/executor/move_util.go
```diff
@@ -511,6 +520,29 @@ func changeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstruc
...
+	labels := obj.GetLabels()
+	if labels == nil {
+		// nothing to do
+		return
+	}
+	if _, exists := labels[TurboGCLabelKey]; exists {
+		delete(labels, TurboGCLabelKey)
+	}
+
+	obj.SetLabels(labels)
```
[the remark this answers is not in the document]
irfanurrehman: Done!

pkg/action/executor/move_util.go
```diff
@@ -478,20 +484,23 @@ func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstruc
...
+func ChangeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
 	kind := obj.GetKind()
 	name := obj.GetName()
 	// This takes care of conflicting updates for example by operator
 	// Ref: https://github.com/kubernetes/client-go/blob/master/examples/dynamic-create-update-delete-deployment/main.go
 	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		objCopy, err := client.Get(context.TODO(), name, metav1.GetOptions{})
 		if valid {
-			if err := unstructured.SetNestedField(objCopy.Object, "default-scheduler",
+			removeGCLabel(objCopy)
```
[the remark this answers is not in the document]
irfanurrehman: Done!

pkg/action/executor/move_util.go
```diff
@@ -462,6 +463,11 @@ func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstruc
 		if err != nil {
 			return err
 		}
+		if pause {
+			addGCLabel(objCopy)
+		} else {
+			removeGCLabel(objCopy)
```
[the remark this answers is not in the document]
irfanurrehman: Done!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Thanks for catching this. Updated!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Done!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Done!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Done!

irfanurrehman:
@ading1977 @enlinxu Updated. PTAL!

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+
+	return nil
+}
+
+func (g *GarbageCollector) revertSchedulers() error {
+	for _, parentRes := range supportedParents {
+		objList, err := g.dynClient.Resource(parentRes).Namespace("").List(context.TODO(), gcListOpts)
+		if err != nil {
+			// we fail forcing kubeturbo to restart and try this again
+			glog.Errorf("Error garbage cleaning parent controller for %s: %v", parentRes.String(), err)
```
ading1977, line 179: nit: "garbage cleaning parent controller" -> "reverting scheduler"

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+			return nil
+		}
+		for _, item := range objList.Items {
+			valid := true
+			err := commonutil.RetryDuring(executor.DefaultRetryLess, 0,
+				executor.DefaultRetrySleepInterval, func() error {
+					return executor.ChangeScheduler(g.dynClient.Resource(parentRes).Namespace(item.GetNamespace()), &item, valid)
+				})
+			if err != nil {
+				glog.Errorf("Error garbage cleaning controller for %s %s/%s: %v", item.GetKind(), item.GetNamespace(), item.GetName(), err)
```
ading1977, line 193: nit: "garbage cleaning parent controller" -> "reverting scheduler"

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+func (g *GarbageCollector) unpauseRollouts() error {
+	for _, grandParentRes := range supportedGrandParents {
+		objList, err := g.dynClient.Resource(grandParentRes).Namespace("").List(context.TODO(), gcListOpts)
+		if err != nil {
+			// This error could be because of mising deploymentConfig type from the cluster, so we don't fail here
+			if apierrors.IsNotFound(err) && strings.Contains(err.Error(), "the server could not find the requested resource") {
+				glog.V(3).Infof("Resource not found enabled on cluster while garbage cleaning controllers %s: %v", grandParentRes.String(), err)
+				continue
+			}
+			glog.Errorf("Error garbage cleaning grand parent controller for %s: %v", grandParentRes.String(), err)
```
ading1977, line 211: nit: "garbage cleaning grand parent controller" -> "unpausing rollout"

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +156,78 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+			return nil
+		}
+		for _, item := range objList.Items {
+			unpause := false
+			err := commonutil.RetryDuring(executor.DefaultRetryLess, 0,
+				executor.DefaultRetrySleepInterval, func() error {
+					return executor.ResourceRollout(g.dynClient.Resource(grandParentRes).Namespace(item.GetNamespace()), &item, unpause)
+				})
+			if err != nil {
+				glog.Errorf("Error garbage cleaning controller for %s %s/%s: %v", item.GetKind(), item.GetNamespace(), item.GetName(), err)
```
ading1977, line 225: nit: "garbage cleaning grand parent controller" -> "unpausing rollout"

ading1977 (approved):

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Updated!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Updated!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Updated!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Updated!

irfanurrehman:
@ading1977 Updated the nits. PTAL!

PR #524, opened 2021-02-17
Cleanup dangling quotas at kubeturbo startup

irfanurrehman:
Implements https://vmturbo.atlassian.net/browse/OM-66784

This is on top of #523 and only the last 2 commits are reviewable here.


irfanurrehman:
Rebased!

pkg/action/executor/quota_util.go
```diff
@@ -307,3 +310,26 @@ func getCodecForGV(gv schema.GroupVersion) (runtime.Codec, error) {
...
+	labels := quota.GetLabels()
+	if labels == nil {
+		// nothing to do
+		return
+	}
+	if _, exists := labels[TurboGCLabelKey]; exists {
+		delete(labels, TurboGCLabelKey)
+	}
+
+	quota.SetLabels(labels)
```
ading1977, line 334: nit: Move this into the `exists` block after `delete(labels, TurboGCLabelKey)`

pkg/discovery/worker/garbage_collector.go
```diff
@@ -231,3 +235,70 @@ func (g *GarbageCollector) unpauseRollouts() error {
...
+	var err error
+	if quota.Annotations != nil {
+		origSpec, exists := quota.Annotations[executor.QuotaAnnotationKey]
+		if exists {
+			revertedQuota, err = executor.DecodeQuota([]byte(origSpec))
+			if err != nil {
+				// This is very unlikely but not an error which we can recover from, for example in the next run.
+				// We need to ignore this. The annotation if there still is any would then carry the opportunity
+				// for an user to use this as a reference in case a manual correction is required ever.
+				glog.Warningf("Error reverting quota while garbage collecting resources: %s/%s: %v", quota.Namespace, quota.Name, err)
```
ading1977, line 272: I don't quite understand this. If decoding fails, we ignore it, and then we remove the quota annotation later. Is this desired?

pkg/discovery/worker/garbage_collector.go
```diff
@@ -231,3 +235,70 @@ func (g *GarbageCollector) unpauseRollouts() error {
...
+	annotations := quota.GetAnnotations()
+	if annotations == nil {
+		// nothing to do
+		return
+	}
+	if _, exists := annotations[executor.QuotaAnnotationKey]; exists {
+		delete(annotations, executor.QuotaAnnotationKey)
+	}
+
+	quota.SetAnnotations(annotations)
```
ading1977, line 303: nit: same as above, move it into the `exists` block.

pkg/action/executor/quota_util.go
irfanurrehman: Done!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Thanks for catching this. Have updated (which was my original intention anyways) to not remove the annotation if decode could not get a valid quota.

pkg/discovery/worker/garbage_collector.go
irfanurrehman: Done!

irfanurrehman:
@ading1977 Updated. PTAL!

ading1977 (approved):

pkg/discovery/worker/garbage_collector.go
```diff
@@ -110,3 +160,144 @@ func (g *GarbageCollector) isLeakedPod(pod api.Pod) bool {
...
+			}
+		}
+	}
+	if revertedQuota != nil {
+		// Although unlikely, revertedQuota being nil possibly the result of the decode error above
+		// so we go ahead and try to remove the GC label alone and leave the annotation behind.
+		quota.Spec.Hard = revertedQuota.Spec.Hard
+		RemoveTurboAnnotionFromQuota(quota)
+	}
+	executor.RemoveGCLabelFromQuota(quota)
```
marvelyue, line 282: Hi @irfanurrehman, I may be missing here. I understand it would be a rare case to fail the decoding. But I'm wondering if we want to remove BOTH turbo annotation and GC label from quota only if decoding succeeds (revertedQuota != nil).

I noticed in your changes above in `Revert()` func in `quota_util`, when decoding fails, we'll continue in the for loop without calling `RemoveGCLabelFromQuota`:
```
func (q *QuotaAccessorImpl) Revert() {
	for _, quota := range q.quotas {
		var revertedQuota *corev1.ResourceQuota
		var err error
		if quota.Annotations != nil {
			origSpec, exists := quota.Annotations[QuotaAnnotationKey]
			if exists {
				revertedQuota, err = decodeQuota([]byte(origSpec))
				revertedQuota, err = DecodeQuota([]byte(origSpec))
				if err != nil {
					glog.Warningf("Error reverting the updated quota: %s/%s: %v", quota.Namespace, quota.Name, err)
					continue
				}
			}
		}
		if revertedQuota == nil {
			continue
		}

		RemoveGCLabelFromQuota(revertedQuota)
		...
	}
}
```

Or maybe to make it simple, we could just return the err when decoding fails above? Something like:
```
revertedQuota, err = executor.DecodeQuota([]byte(origSpec))
			if err != nil {
				// This is very unlikely but not an error which we can recover from, for example in the next run.
				// We need to ignore this. The annotation if there still is any would then carry the opportunity
				// for an user to use this as a reference in case a manual correction is required ever.
				glog.Warningf("Error reverting quota while garbage collecting resources: %s/%s: %v", quota.Namespace, quota.Name, err)
                                return err
                         }
```

pkg/action/executor/move_util.go
```diff
@@ -478,24 +486,27 @@ func resourceRollout(client dynamic.ResourceInterface, obj *unstructured.Unstruc
...
-func changeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
+// ChangeScheduler sets the scheduler value to default or unsets it to an invalid one
+// This additionally adds or removes the garbage collection label on the resource
+func ChangeScheduler(client dynamic.ResourceInterface, obj *unstructured.Unstructured, valid bool) error {
 	kind := obj.GetKind()
 	name := obj.GetName()
 	// This takes care of conflicting updates for example by operator
 	// Ref: https://github.com/kubernetes/client-go/blob/master/examples/dynamic-create-update-delete-deployment/main.go
 	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		objCopy, err := client.Get(context.TODO(), name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
irfanurrehman: Thanks @marvelyue, good catch. Updated!

pkg/discovery/worker/garbage_collector.go
irfanurrehman: There are 2 cases I could think of the decode would fail.
1. Dependency encode/decode functionality is updated (in k8s) and there is some mismatch in the encode/decode that we are using such that the data encoded by the encode method cannot be decoded by our decode method.
2. After we place the annotation, another entity (for example an user) updates it before the decode is hit in the flow, such that the update breaks the expected json structure and is no longer valid for decode.

In both the above cases (or for that matter any other possible decode failure case), the failure won't be transient. That is it will always fail and there would be no recovery. 
Now I chose not to fail in the normal flow (`quota_util.go revert()`), because failing at that time would mean that we will skip updating rest of the quotas (if multiple quota resources were updated to allow the pod move) for a failure which we cannot anyways recover from. The label is left behind for it to be tried once more in the garbage collection flow (when ever kubturbo restarts again). In the garbage collection flow, we fail kubeturbo start on failure, so if we return failure on a non recoverable failure, kubeturbo will remain in an infinite restart loop and hence the comment in the snippet you posted:
```
// This is very unlikely but not an error which we can recover from, for example in the next run.
// We need to ignore this. The annotation if there still is any would then carry the opportunity
// for an user to use this as a reference in case a manual correction is required ever.
```

Does this make sense? Plz let me know if you still have concerns?

PR #525, opened 2021-02-23
Fix quota update race

irfanurrehman:
Implements https://vmturbo.atlassian.net/browse/OM-66786

This is on top of #524 and only the last 2 commits are reviewable here

### DIFF
--- a/pkg/action/action_handler.go
+++ b/pkg/action/action_handler.go
@@ -86,6 +86,18 @@ type ActionHandler struct {
 	// concurrency control
 	lockStore IActionLockStore
 
+	// TODO(irfanurrehman): Improve this.
+	// Currently the action handler uses a key based lock
+	// implementing an expiration map and then abstracts the implementation
+	// via the interface IActionLockStore. This does not seem to be needed.
+	// The abstraction also renders other uses of the lock map impossible.
+	// It seems that there is no need to have all the added complexity and a simple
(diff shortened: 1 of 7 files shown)


irfanurrehman:
Rebased!

ading1977:
The change looks fine to me. I just have one question, do we want to synchronize the `cleanupQuotas` during restart using the same lock?

irfanurrehman:
> The change looks fine to me. I just have one question, do we want to synchronize the `cleanupQuotas` during restart using the same lock?

The `cleanupQuotas` during restart would happen before anything else runs in kubeturbo. Also any failure (well except 1 specific failure type) during this process will restart kubeturbo. A lock is therefor not needed.

ading1977:
> > The change looks fine to me. I just have one question, do we want to synchronize the `cleanupQuotas` during restart using the same lock?
> 
> The `cleanupQuotas` during restart would happen before anything else runs in kubeturbo. Also any failure (well except 1 specific failure type) during this process will restart kubeturbo. A lock is therefor not needed. Is there any specific scenario that comes to your mind to use a lock at this stage.. ?

Thanks, I thought `cleanupQuotas` is done in a goroutine, but after another look it is not. So it should be fine.

ading1977 (approved):
LGTM

marvelyue (approved):
LGTM

PR #527, opened 2021-02-24
Introduce cpuOvercommitment and memOvercommitment attributes to ContainerPlatformCluster entity.

marvelyue:
**Intent**:

We'd like to present ContainerPlatformCluster cpu/memory overcommitment in the result of new "Optimize Container Cluster" plan so that our users will understand better the risks of their K8s clusters.

turbo-go-sdk changes in https://github.com/turbonomic/turbo-go-sdk/pull/116

**Background**:

We introduced a new "Optimize Container Cluster" plan under a feature flag in https://git.turbonomic.com/turbonomic/ux-app/-/merge_requests/2185.

In plan result, for container platform cluster entities, we'd like to present resource changes including VCPU, VMem, VCPURequest, VMemRequest, VCPULimitQuota, VMemLimitQuota, VCPURequestQuota and VMemRequestQuota, as well as CPU overcommitment and memory overcommitment values.

CPU overcommitment and memory overcommitment values are defined as:

```
CPU overcommitment = sum of all containers CPU limits / sum of all VMs CPU capacity
Memory overcommitment = sum of all containers memory limits / sum of all VMs CPU capacity
```

Currently we have all resources modeled as commodities. We don't have a clear use case to model the overcommitment as commodities to drive actions. So our solution is to set `cpuOvercommitment` and `memOvercommitment` as attributes to ContainerPlatformCluster entity so that they can be returned along with other commodity values by stats API call.

**Implementation**:
When calculating CPU/memory overcommitment, sum of all containers CPU limits is equivalent to sum of all namespace CPU limit quota usage. So in `cluster_dto_builder.go`, calculate the CPU/memory overcommitment values by:
```
sum of all namespace resource limit quota usage / sum of all VMs resource capacity
```

**Unit Test**:
Added unit test to test `ClusterData`

**Testing Done**:
In discovery response, `container_platform_cluster_data` is populated with cpuOvercommitment and memOvercommitment values:
```
entityDTO {
  entityType: CONTAINER_PLATFORM_CLUSTER
  id: "863bcf35-34c2-11eb-a0d1-005056b81083"
  displayName: "Kubernetes-Turbonomic"
  ...
  container_platform_cluster_data {
    cpuOvercommitment: 0.0625
    memOvercommitment: 7.142769859988617
  }
}
```

And verified the values match the calculation of
```
sum of all namespace resource limit quota usage / sum of all VMs resource capacity
```

**Note**:
Travis CI is failed because of:
```
pkg/discovery/dtofactory/cluster_dto_builder.go:159: undefined: "github.com/turbonomic/turbo-go-sdk/pkg/proto".EntityDTO_ContainerPlatformClusterData
pkg/discovery/dtofactory/cluster_dto_builder.go:189: undefined: "github.com/turbonomic/turbo-go-sdk/pkg/proto".EntityDTO_ContainerPlatformClusterData
```

Will update go.mod to include turbo-go-sdk changes to pass the CI build after merging https://github.com/turbonomic/turbo-go-sdk/pull/116


pkg/discovery/dtofactory/cluster_dto_builder.go
```diff
@@ -140,5 +150,59 @@ func (builder *clusterDTOBuilder) getCommoditiesSold(entityDTOs []*proto.EntityD
...
+			}
+		case proto.CommodityDTO_VMEM:
+			resourceCapacity := builder.getNodeResourceCapacity(clusterName, commodityType, nodeResourceCapacityMap)
+			if resourceCapacity != 0 {
+				vmemOvercommitment = usage / resourceCapacity
+			}
+		default:
+			glog.Errorf("Unsupported commodity type %s for cluster overcommitment", commodityType)
+		}
+	}
```
dblinn, line 187: Hmm... I'm wondering if it's important to be able to distinguish the case when there's no overcommitment (ie if the cluster is empty) vs if there's no usage in the quota. I'm thinking it might not be because actually getting 0 usage in practice would be extremely unusual. Just wanted to see if you had thought about this distinction. Like setting the overcommitment in the data to null if capacity == 0.

Maybe at least we could log a message if capacity is 0.

dblinn (approved):

pkg/discovery/dtofactory/cluster_dto_builder.go
marvelyue: Thanks @dblinn. This is a good point. Same 0 value could represent different meanings (empty cluster vs. limits not specified). As we discussed, we'll present overcommitment values together with other resource utilization data. We could identify if the cluster is empty based on resource utilization.

And I have added log messages if capacity is 0 in `getNodeResourceCapacity` func.

PR #532, opened 2021-03-04
Create NumberReplicas commodity for Service entity

ading1977:
This PR creates `NumberReplicas` commodity (sold) for Service entity.

For every discovered **Service** in the cluster:

* Scan all pods that provide to this service
* Collect controllers of these pods with valid desired Replicas. Normally there should only be 1 controller for all pods that provide to a service, but we also need to capture the case where pods may belong to multiple controllers for whatever reasons
* Create `NumberReplicas` commodity sold with `used` as the number of ready pods, and `capacity` as the max replicas from all controllers

Tested using the twitter application:
image


pkg/discovery/dtofactory/service_entity_dto_builder.go
```diff
@@ -95,14 +98,81 @@ func (builder *ServiceEntityDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error)
...
+	// The list of pods that provide to the service may belong to multiple controllers
+	// Determine capacity by getting the max of all replicas
+	var replicas int64
+	for _, controller := range controllers {
+		if *controller.Replicas > replicas {
+			replicas = *controller.Replicas
+		}
+	}
+	used := float64(readyPods)
+	capacity := float64(replicas)
```
enlinxu, line 159: what will happen if the replica is 0?

enlinxu: never mind, doesn't look like it will happen here.

ading1977: Right, controllers without replicas, or with replicas < 1 will not be considered.

enlinxu (approved):

marvelyue (approved):
LGTM

PR #535, opened 2021-03-09
Skip pending pods when building containerspecs owned by controller

ading1977:
This fixes https://vmturbo.atlassian.net/browse/OM-64209. 

The root cause of this problem is that a container pod is scheduled on a node, but still in Pending phase, i.e., image pull failed, so there is no container started in the pod, and there is no corresponding containerSpec entity either created either. However, in `kubeturbo`, we still create the `consistsOf` relationship of the workload controller. 

image

The fix is to skip the pending pods when building containerSpecs owned by workloadController. 

Verified that the error messages do not show up in `topology-processor` log any more.

### DIFF
--- a/pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
@@ -137,6 +137,14 @@ func (builder *workloadControllerDTOBuilder) getCommoditiesBought(kubeController
 func (builder *workloadControllerDTOBuilder) getContainerSpecIds(kubeController *repository.KubeController) []string {
 	containerNameSet := make(map[string]struct{})
 	for _, pod := range kubeController.Pods {
+		// Pods scheduled on a node, but still in Pending phase do not have containers started, exclude them
+		// when getting container specs
+		if discoveryUtil.PodIsPending(pod) {
+			glog.V(3).Infof("Skipping pod %v when building containerSpecs owned by controller %v."+
+				" There is no container started in the pod.", discoveryUtil.GetPodClusterID(pod),
+				kubeController.GetFullName())
+			continue
+		}
 		for _, container := range pod.Spec.Containers {
 			containerNameSet[container.Name] = struct{}{}
 		}


enlinxu (approved):

pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
```diff
@@ -137,6 +137,14 @@ func (builder *workloadControllerDTOBuilder) getCommoditiesBought(kubeController
 func (builder *workloadControllerDTOBuilder) getContainerSpecIds(kubeController *repository.KubeController) []string {
 	containerNameSet := make(map[string]struct{})
 	for _, pod := range kubeController.Pods {
+		// Pods scheduled on a node, but still in Pending phase do not have containers started, exclude them
+		// when getting container specs
+		if discoveryUtil.PodIsPending(pod) {
+			glog.V(3).Infof("Skipping pod %v/%v when building containerSpecs owned by controller %v/%v."+
+				" There is no container started in the pod.",
+				pod.Namespace, pod.Name, kubeController.Namespace, kubeController.Name)
```
marvelyue, line 145: Nit: we have some existing functions to show the proper pod/controller name with namespace something like:
```
glog.V(3).Infof("Skipping pod %v when building containerSpecs owned by controller %v."+
				" There is no container started in the pod.",
				discoveryUtil.GetPodClusterID(pod), kubeController.GetFullName())

```

marvelyue (approved):

pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
ading1977: Changed accordingly. Now we see the following logs:
```
I0309 18:41:58.056181   43802 workload_controller_entity_dto_builder.go:143] Skipping pod ccp/cert-manager-7c4fdf69b7-bc67h when building containerSpecs owned by controller ccp/Deployment/cert-manager. There is no container started in the pod.
I0309 18:41:58.056294   43802 workload_controller_entity_dto_builder.go:143] Skipping pod kube-system/coredns-7c4d89dd66-jr5zj when building containerSpecs owned by controller kube-system/Deployment/coredns. There is no container started in the pod.
```

PR #537, opened 2021-03-11
Discover imagefs commodity

irfanurrehman:
Implements https://vmturbo.atlassian.net/browse/OM-67515

This is on top of #533. Only the last commit is reviewable here

### DIFF
--- a/pkg/discovery/dtofactory/general_builder.go
+++ b/pkg/discovery/dtofactory/general_builder.go
@@ -179,18 +179,19 @@ func (builder generalBuilder) getSoldResourceCommodityWithKey(entityType metrics
 		return nil, err
 	}
 
-	if (resourceType == metrics.Memory || resourceType == metrics.VStorage) && entityType == metrics.NodeType {
-		thresholdAvailable, err := builder.metricValue(entityType, entityID,
+	if resourceType == metrics.Memory && entityType == metrics.NodeType {
+		threshold, err := builder.metricValue(entityType, entityID,
 			resourceType, metrics.Threshold, nil)
 		if err != nil {
 			glog.Warningf("Missing threshold value for %v for node %s.", resourceType, entityID)
 		}
 		// TODO: The settable method for UtilizationThresholdPct can be added to the sdk instead.
-		if thresholdAvailable.Avg > 0 && thresholdAvailable.Avg <= 100 {
-			thresholdUtilization := 100 - thresholdAvailable.Avg
+		if threshold.Avg > 0 && threshold.Avg <= 100 {
+			// Threshold values set in kubelet are in terms of metrics (eg rootfs size) value available.
+			thresholdUtilization := 100 - threshold.Avg
 			resourceCommoditySold.UtilizationThresholdPct = &thresholdUtilization
 		} else {
-			glog.Warningf("Threshold value [%.2f] outside range and will not be set for %v for node %s.", thresholdAvailable.Avg, resourceType, entityID)
+			glog.Warningf("Threshold value [%.2f] outside range and will not be set for %v for node %s.", threshold.Avg, resourceType, entityID)
 		}
 	}
 	return resourceCommoditySold, nil
@@ -264,6 +265,12 @@ func (builder generalBuilder) getResourceCommoditiesBought(entityType metrics.Di
 
 		// set peak value as the used value
 		commBoughtBuilder.Peak(metricValue.Peak)
+
+		if rType == metrics.VStorage {
+			// set commodity key only for vstorahge.
+			// currently pods only report and buy rootfs usage.
+			commBoughtBuilder.Key("k8s-node-rootfs")
+		}
 
 		// set additional attribute
 		if commodityAttrSetter != nil && commodityAttrSetter.Settable(rType) {
--- a/pkg/discovery/dtofactory/node_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/node_entity_dto_builder.go
@@ -27,7 +27,6 @@ var (
 		metrics.CPURequest,
 		metrics.MemoryRequest,
 		metrics.NumPods,
-		metrics.VStorage,
 		// TODO, add back provisioned commodity later
 	}
 
@@ -82,7 +81,7 @@ func (builder *nodeEntityDTOBuilder) BuildEntityDTOs(nodes []*api.Node) []*proto
 		}
 
 		// compute and constraint commodities sold.
-		commoditiesSold, err := builder.getNodeCommoditiesSold(node, clusterId)
+		commoditiesSold, isAvailableForPlacement, err := builder.getNodeCommoditiesSold(node, clusterId)
 		if err != nil {
 			glog.Errorf("Error when create commoditiesSold for %s: %s", node.Name, err)
 			nodeActive = false
@@ -163,6 +162,10 @@ func (builder *nodeEntityDTOBuilder) BuildEntityDTOs(nodes []*api.Node) []*proto
 			continue
 		}
 
+		entityDto.ProviderPolicy = &proto.EntityDTO_ProviderPolicy{AvailableForPlacement: &isAvailableForPlacement}
+		if !isAvailableForPlacement {
+			glog.Warningf("Node %s has been marked unavailable for placement.", node.GetName())
+		}
 		result = append(result, entityDto)
 
 		glog.V(4).Infof("Node DTO : %+v", entityDto)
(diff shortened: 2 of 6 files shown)


pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -221,11 +224,82 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
...
-		return nil, err
+		return nil, isAvailableForPlacement, err
 	}
 	commoditiesSold = append(commoditiesSold, clusterComm)
 
-	return commoditiesSold, nil
+	return commoditiesSold, isAvailableForPlacement, nil
+}
+
+func (builder *nodeEntityDTOBuilder) getNodeStorageCommoditiesSold(nodeName string) ([]*proto.CommodityDTO, bool) {
```
enlinxu, line 234: Please add a comment for the method to explain the behavior of isAvailableForPlacement

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -163,6 +163,7 @@ func (builder *nodeEntityDTOBuilder) BuildEntityDTOs(nodes []*api.Node) []*proto
 			continue
 		}
 
+		entityDto.ProviderPolicy = &proto.EntityDTO_ProviderPolicy{AvailableForPlacement: &isAvailableForPlacement}
```
enlinxu, line 166: Please print a warning log if the isAvailableForPlacement is false to help troubleshooting

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -221,11 +224,82 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
...
+	vstorageResources := []string{"rootfs", "imagefs"}
+	entityType := metrics.NodeType
+	resourceType := metrics.VStorage
+	protoCommodityType := proto.CommodityDTO_VSTORAGE
+	isAvailableForPlacement := true
+
+	for _, resource := range vstorageResources {
+		entityID := nodeName
+		if resource == "imagefs" {
+			entityID = fmt.Sprintf("%s-%s", nodeName, resource)
```
enlinxu, line 245: Will this be unique globally?

irfanurrehman: This needs to be unique among the vstorage commodities on the give node. Is that not so?

pkg/discovery/dtofactory/node_entity_dto_builder.go
irfanurrehman: I generally try to use self documenting names and sometimes an additional variable to explain the purpose, especially for evaluating conditionals like this. For example as done [here](https://github.com/turbonomic/kubeturbo/pull/537/files#diff-c8b9a819f9d9f29527ec258b2617a74fe62e1645dab6805fb4ab05991f2d1e3eR300) for this variable.
I have anyways added a comment at the top of the function also now.

pkg/discovery/dtofactory/node_entity_dto_builder.go
irfanurrehman: Done!

irfanurrehman:
@enlinxu Updated as per your comments.
Also find below screenshot of how the commodity key looks in the UI:
image
I have further dropped the node name from the  key. The pods report the rootfs used values and buys the vstorage commodity from the node, so they also need to include the same key in the bought commodity that the node is selling with. If we keep the node name in the key, then the pod would be stuck to the node its currently on and could not be moved to any other node.

We can further discuss the need of uniqueness of this key, if need be.

enlinxu:
@irfanurrehman  Please rebase and we will continue the review

irfanurrehman:
> I also noticed that the imagefs used is reported more then the capacity on some nodes (from ae2 cluster). I need to find out the reason as to why is this reported this way on this cluster.

So I spent time on this and figured out that the used value reported for imagefs is actually the sum total of all images, which means that the common image layers can figure in the calculation multiple times. I have left comments in the code to pinpoint this. I resorted to using the available bytes instead. With the usage of available bytes, however, we would end up with exactly the same values (capacity and used) for `rootfs` and `imagefs`, if both are configured to use the same partition. I therefor had to add more logic to  ensure we don't add the imagefs commodity, which might appear as a duplicate, if both are configured to use the same partition. With this logic in place, I have kept consideration of threshold for imagefs still in place even if we don't add the commodity.
I have kept the additional changes into a separate commit to help review.

@ading1977 @marvelyue, it will be great if you can also have a look at this once, please.

ading1977:
> @ading1977 @marvelyue, it will be great if you can also have a look at this once, please.
I will take a look.

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -221,11 +226,108 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
...
+		// comparing the reported capacity and available bytes, to the last byte, for both
+		// rootfs and imagefs to determine if we are getting the reported values for the same partition.
+		isPartitionSame := resource == "imagefs" && rootfsCapacityBytes == capacityBytes.Avg && rootfsAvailableBytes == availableBytes.Avg
+		if isPartitionSame {
+			// We skip adding imagefs commodity, however we still honor the thresholds set for imagefs
+			// which can be different compared to rootfs, even when the partitions are same.
+			// isAvailableForPlacement is still calculated for both above and would be set to false
+			// if either of the values cross threshold.
+			continue
+		}
```
ading1977, line 325: nit: Looks like `isPartitionSame` is not an unusual case, maybe move this section above to around line 287 so we can continue early if needed?

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -163,6 +162,10 @@ func (builder *nodeEntityDTOBuilder) BuildEntityDTOs(nodes []*api.Node) []*proto
 			continue
 		}
 
+		entityDto.ProviderPolicy = &proto.EntityDTO_ProviderPolicy{AvailableForPlacement: &isAvailableForPlacement}
+		if !isAvailableForPlacement {
+			glog.Warningf("Node %s has been marked unavailable for placement.", node.GetName())
```
ading1977, line 167: Maybe we can be more clear, saying something like `Node %s has been marked unavailable for placement due to disk pressure`?

ading1977 (approved):
Looks fine to me. Just a few small things.

pkg/discovery/dtofactory/node_entity_dto_builder.go
irfanurrehman: Thanks @ading1977 for the suggestion. This is however intentionally here, because we want the thresholds to be evaluated even when its the same partition.

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -221,11 +226,108 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
...
+
+			if threshold.Avg > 0 && threshold.Avg <= 100 {
+				isAvailableAboveThreshold := availableBytes.Avg > threshold.Avg*capacityBytes.Avg/100
+				isAvailableForPlacement = isAvailableAboveThreshold
+				utilizationThreshold := 100 - threshold.Avg
+				// TODO: The settable method for UtilizationThresholdPct can be added to the sdk instead.
+				resourceCommoditySold.UtilizationThresholdPct = &utilizationThreshold
+			} else {
+				glog.Warningf("Threshold value [%.2f] outside range and will not be set for %v : %s for node %s.",
+					threshold.Avg, resourceType, resource, entityID)
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/discovery/dtofactory/node_entity_dto_builder.go
irfanurrehman: Updated!

enlinxu:
@irfanurrehman please prepare a demo to explain the changes. We need to make sure PM (Eva) and Documentation team is aware of this change.

pkg/discovery/dtofactory/general_builder.go
```diff
@@ -265,6 +266,12 @@ func (builder generalBuilder) getResourceCommoditiesBought(entityType metrics.Di
 		// set peak value as the used value
 		commBoughtBuilder.Peak(metricValue.Peak)
 
+		if rType == metrics.VStorage {
+			// set commodity key only for vstorahge.
+			// currently pods only report and buy rootfs usage.
+			commBoughtBuilder.Key("k8s-node-rootfs")
```
marvelyue, line 272: Nit: can you create a constant for "k8s-node-rootfs"?

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -221,11 +226,108 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
...
+}
+
+// getNodeStorageCommoditiesSold builds sold storage commodities for the node
+// Returns the built commodities and if this node is available for placement or not.
+// The availability for placement is evaluated based on the current usage crossing the
+// configured threshold. If the usage has crossed the threshold, we mark the node
+// NOT available for placement.
+func (builder *nodeEntityDTOBuilder) getNodeStorageCommoditiesSold(nodeName string) ([]*proto.CommodityDTO, bool) {
+	var resourceCommoditiesSold []*proto.CommodityDTO
+	vstorageResources := []string{"rootfs", "imagefs"}
```
marvelyue, line 243: Nit: can you create constants for "rootfs" and "imagefs" to avoid unexpected typo when using in multiple places?

marvelyue (approved):

PR #538, opened 2021-03-11
Set availableForPlacement false to cluster and set up hasCpuLimitsSet and hasMemLimitsSet to ContainerData.

marvelyue:
JIRA: https://vmturbo.atlassian.net/browse/OM-67203
# Intent
There are 2 goals:
1. Set up `hasCpuLimitsSet` and `hasMemLimitsSet` to `ContainerData` so that in Turbo platform, we'll be able to identify the containers with CPU/memory limits set.
2. Set "availableAsProvider" to false to cluster to avoid unexpected entities incorrectly placed on cluster during Turbo analysis.

# Background
1. CPU overcommitment and memory overcommitment of ContainerPlatformCluster
```
VCPUOvercommitment = sum of all containers CPU limits / sum of all VMs CPU capacity
VMemOvercommitment = sum of all containers memory limits / sum of all VMs memory capacity
```
If a container doesn't have resource limits specified, we set commodity capacity to corresponding VM resource capacity. In this case, container capacity won't contribute to overcommitment calculation. So we need to identify the containers with CPU/memory limits specified and count them into overcommitment calculation for container cluster.

2. When testing "Add Pod" in plan, the added pods are incorrectly placed on cluster instead of node. This is because constraints (commodities with keys) are ignored by default. So in this case, the commodities bought by the pods can be satisfied by both VMs (nodes) and container platform clusters.
The solution is to use `availableForPlacement` flag in `EntityDTO_ProviderPolicy`. When setting `availableForPlacement` to false to cluster, no new entities will be allowed to be placed on clusters. Note that this flag won't affect resize actions, so this won't affect the use case in the future if we want to make cluster as provider to resize namespaces.

# Implementation
1. Set `hasCpuLimitsSet` and `hasMemLimitsSet` to `ContainerData` in `container_dto_builder`
2. Set `availableAsProvider` to false in `cluster_dto_builder`
3. Added additional unit tests

# Testing Done
1. `hasCpuLimitsSet` and `hasMemLimitsSet` in `ContainerData`
Tested kubeturbo with the changes. After discovery, the discovery DTO response of container entity has containerData set up:
```
entityDTO {
  entityType: CONTAINER
  id: "92b0610d-2438-11eb-af5a-005056b88929-0"
  displayName: "action-merge-test/action-merge-test-7569fd8d8b-thx55/cpu-mem-load"
  ...
  container_data {
    hasCpuLimit: true
    hasMemLimit: true
  }
}
```

2. Tested "Add Pod" in plan.

Before the changes, the added pod was incorrectly placed on cluster:
Screen Shot 2021-03-11 at 2 00 37 AM

After the changes, the added pod can be correctly placed on corresponding node (VM):
Screen Shot 2021-03-11 at 2 01 44 AM

# Note:
CI build will be failed because of missing go sdk changes. Will update go.mod and corresponding vendors after merging go sdk PR https://github.com/turbonomic/turbo-go-sdk/pull/118


ading1977 (approved):
LGTM

dblinn (approved):

enlinxu (approved):
LGTM

PR #539, opened 2021-03-11
Add Tag properties based on labels and annotations for workload contr…

enlinxu:
This is the third review for the function, for the kubeturbo change.
## Intent
When introducing the new container model, we also decided to organize the settings associate to the right Entity. For example, the percentile related policy is associate to the Container Spec, and the action mode policy is associate to the Workload Controller due to the action merge. Because of that, we need to be able to create the groups for these entities. Customers like PWC has special labels and annotations on K8s, which they want to use for group creation.

## Solution
Add group criteria (by tags) for both Workload Controller and Container Spec. Please noted that the Container Spec is using the same tags as the Workload Controller, since container spec itself doesn't have labels and annotations.

## Tests done: (Together with the change in api and kubeturbo)
Screen Shot 2021-03-11 at 9 45 45 AM
Screen Shot 2021-03-11 at 9 45 22 AM


ading1977 (approved):
LGTM

irfanurrehman (approved):
Thanks @enlinxu for doing this.
Fo some reason this PR has 2 commits with the same comment. Please squash them while merging.

marvelyue (approved):
LGTM

PR #398, opened 2020-04-06
Create new Namespace entity for new container model

marvelyue:
**Intent**:
The intent is to
1. represent k8s namespace using Namespace entity instead of VDC (VirtualDataCenter)
2. have the Namespace entity sell `VCPULimitQuota`/`VMemLimitQuota`and `VCPURequestQuota`/`VMemRequestQuota` commodities to take K8s namespace resource quota into consideration for container resizing

**Background**:
The supply chain of the Namespace is shown in the screenshot below:
Screen Shot 2020-04-07 at 11 31 51
where 
- Namespace sells `VCPULimitQuota`/`VMemLimitQuota`and `VCPURequestQuota`/`VMemRequestQuota` resource commodities as well as  to WorkloadController (representing K8s controller like Deployment, StatefulSet, etc. which will be created in a different story)
- if there's bare pod deployed without k8s controller, Namespace directly sells the commodities to Pod

**Implementation**:
1. Renamed `quota_entity_dto_builder.go` to `namespace_entity_dto_builder.go` and replaced `VirtualDatacenter` entity type with `Namespace` entity type.
2. For quota commodities sold by Namespace:
- Default quota commodity capacity is set to a large number 1e12 (represents "infinity") to make sure when quota is not configured on namespace, quota commodities have least impact on market analysis result.
- Quota commodity capacity is updated to the configured limits or requests quota in `kube_cluster::ReconcileQuotas`
- Quota commodity used is sum of compute resources limits or requests of all containers of the corresponding pods in the Namespace. Pod quota used is collected in `metrics_collector::createPodMetrics` and stored in `PodMetrics::QuotaUsed`. Quota usage sum of a namespace quota resource is calculated in `metrics_collector::SumQuotaUsage`
3. Removed the logic of node selling allocation commodities and VDC (namespace) buying allocation commodities from node. Removed `NodeMetrics` from `entity_metrics.go`, which used to be used to provide allocation commodities sold by Node.
4. Refactored the code to use `KubeNamespace` entity instead of `KubeQuota` entity in `kube_cluster.go`.
5. Renamed `AllocationResource` to `QuotaResource` in multiple places. Renamed `QuotaMetrics` to `NamespaceMetrics` in `entity_metrics.go`. 
6. In `pod_entity_dto_builder.go`, if it is a bare pod with no k8s controller, pod buys quota commodities from Namespace. Additional changes will be made in a different story to have pod buy quota commodities from WorkloadController If pod is deployed by k8s controller.

**Unit Tests**:
[...]

PR #540, opened 2021-03-18
Only take active nodes into account for cluster resource commodities.

marvelyue:
JIRA: https://vmturbo.atlassian.net/browse/OM-68477
# Intent
1. Update `cluster_dto_builder` to take only active nodes when calculating commodity capacity. Unschedule nodes should be counted as cluster capacity if they are in Ready status. 
2. We need to be consistent everywhere in kubeturbo (node builder, cluster builder, cluster summary) to count active nodes.

# Background
In kubeturbo discovery, we identify following nodes:
1. Nodes in ready status - `NodeCondition` type is ready and `NodeCondition` status is true
2. Nodes which are unschedulable — `node.Spec.Unschedulable` (controls node schedulability of new pods)

Unschedulable nodes can be ready or not ready.

# Implementation
1. In `cluster_dto_builder.go`, only take active nodes into account for cluster resource commodities.
2. In `namespace_discovery_worker.go` and `kubelet_client.go`, set `nodeActive` only if node is in Ready status.
3. In `kube_cluster.go`, remove the redundant `ClusterResources` field and `computeClusterResources()` function. The field and function were introduced in PR #173 but looks not used anywhere since then. A background of a previous use case of clusterResource during discovery is that we used to set namespace (which was modeled as VDC entity) capacity as cluster capacity (sum of node capacity) if quota is not defined. After the refactoring to model namespace as Namespace entity in our new container model in #398 , Namespace capacity has been set to an super large number (representing infinity) so that clusterResource is not needed anymore during discovery before creating cluster entity dto. So it should be safe to delete `computeClusterResources()` function here.
4. Updated unit test

### DIFF
--- a/pkg/discovery/dtofactory/cluster_dto_builder.go
+++ b/pkg/discovery/dtofactory/cluster_dto_builder.go
@@ -131,7 +131,8 @@ func (builder *clusterDTOBuilder) getCommoditiesSold(entityDTOs []*proto.EntityD
 	used := make(map[proto.CommodityDTO_CommodityType]float64)
 	capacity := make(map[proto.CommodityDTO_CommodityType]float64)
 	for _, entityDTO := range entityDTOs {
-		if entityDTO.GetEntityType() == proto.EntityDTO_VIRTUAL_MACHINE {
+		// Only take active nodes into account for cluster resource commodities.
+		if entityDTO.GetEntityType() == proto.EntityDTO_VIRTUAL_MACHINE && entityDTO.GetPowerState() == proto.EntityDTO_POWERED_ON {
 			for _, commodity := range entityDTO.GetCommoditiesSold() {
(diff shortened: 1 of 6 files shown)


enlinxu (approved):
LGTM, please make sure
1. Arsen is aware the change so that he can change his test cases if it's there (if not, we should add)
2. Make sure doc is updated.

chlam4 (approved):

ading1977 (approved):

marvelyue:
> LGTM, please make sure
> 
> 1. Arsen is aware the change so that he can change his test cases if it's there (if not, we should add)
> 2. Make sure doc is updated.

Hi @enlinxu, thanks for bringing this up. I updated the acceptance criteria to include the kubeturbo changes here in the story https://vmturbo.atlassian.net/browse/OM-65311

PR #544, opened 2021-04-07
Addon tasks for throttling metrics collection

irfanurrehman:
This change places the throttling metrics collection and discovery code behind a feature gate, but feature enabled by default.

To disable the feature following needs to be added to the kubeturbo config:

```
        "featureGates": {
          "disabledFeatures": ["ThrottlingMetrics"]
        }
```

@evat-pm do you think we need to add the option to configure this in the kubeturbo operator also? As far as I understand the gate gives us the ability to disable this feature in case we see any problems in some customer environment.

--------------------------------------------

EDIT:
I have added 2 more commits on this one (instead of a separate PR), to enable a faster review cycle.
1. Add support for throttling metrics counter resets and tests for that.
2. Add throttling commodity on containers and container specs even when we don't get these metrics from kubelet. Please note that if the metrics are not collected because of intermittent issues with kubelet connection, we would see a separate error in the logs, but the metrics would still be added with current value as zero.
3. Fix [this](https://github.com/turbonomic/kubeturbo/pull/542#discussion_r608281509) review comment from the previous PR.

### DIFF
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -8,8 +8,10 @@ import (
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/worker/aggregation"
+	"github.com/turbonomic/kubeturbo/pkg/features"
 	sdkbuilder "github.com/turbonomic/turbo-go-sdk/pkg/builder"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
 var (
@@ -81,6 +83,9 @@ func (builder *containerSpecDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, error) 
 func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics *repository.ContainerSpecMetrics) ([]*proto.CommodityDTO, error) {
 	var commoditiesSold []*proto.CommodityDTO
 	for _, resourceType := range ContainerSpecResourceTypes {
+		if resourceType == metrics.VCPUThrottling && !utilfeature.DefaultFeatureGate.Enabled(features.ThrottlingMetrics) {
+			continue
+		}
 		commodityType, exist := rTypeMapping[resourceType]
 		if !exist {
 			glog.Errorf("Unsupported resource type %s when building commoditiesSold for ContainerSpec %s",
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -108,7 +108,7 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 		if resourceType == metrics.VCPUThrottling {
 			typedUsed, ok := resourceMetrics.Used.([][]metrics.ThrottlingCumulative)
 			if ok {
-				aggregatedCap, aggregatedUsed, aggregatedPeak := aggregateThrottlingSamples(typedUsed)
+				aggregatedCap, aggregatedUsed, aggregatedPeak := aggregateThrottlingSamples(containerSpecMetrics.ContainerSpecId, typedUsed)
 				commSoldBuilder.Capacity(aggregatedCap)
 				commSoldBuilder.Peak(aggregatedPeak)
 				commSoldBuilder.Used(aggregatedUsed)
@@ -160,7 +160,7 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 // Throttled values is the percentage of overall throttled counter sum wrt the overall total counter sum
 // across all containers. Peak is the peak of peaks, ie. the peak of individual container peaks calculated
 // from diff of subsequent samples per container.
-func aggregateThrottlingSamples(samples [][]metrics.ThrottlingCumulative) (float64, float64, float64) {
+func aggregateThrottlingSamples(containerSpecId string, samples [][]metrics.ThrottlingCumulative) (float64, float64, float64) {
 	var throttledOverall, totalOverall, peakOverall float64
 	for _, singleContainerSamples := range samples {
 		containerThrottled, containerTotal, containerPeak, ok := aggregateContainerThrottlingSamples("", singleContainerSamples)
@@ -176,6 +176,8 @@ func aggregateThrottlingSamples(samples [][]metrics.ThrottlingCumulative) (float
 	if totalOverall > 0 {
 		if throttledOverall > totalOverall {
 			avgThrottledOverall = 100
+			glog.Warningf("Aggregated throttled cpu samples overshoots total cpu samples for containerSpec %s."+
+				" Throttling set to 100 percent.", containerSpecId)
 		} else {
 			avgThrottledOverall = throttledOverall * 100 / totalOverall
 		}
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -174,7 +174,11 @@ func aggregateThrottlingSamples(samples [][]metrics.ThrottlingCumulative) (float
 	}
 	avgThrottledOverall := float64(0)
 	if totalOverall > 0 {
-		avgThrottledOverall = throttledOverall * 100 / totalOverall
+		if throttledOverall > totalOverall {
+			avgThrottledOverall = 100
+		} else {
+			avgThrottledOverall = throttledOverall * 100 / totalOverall
+		}
 	}
 
 	return 100, avgThrottledOverall, peakOverall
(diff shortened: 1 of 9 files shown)


pkg/discovery/dtofactory/general_builder.go
```diff
@@ -381,16 +385,63 @@ func aggregateContainerThrottlingSamples(entityID string, samples []metrics.Thro
 	sort.SliceStable(samples, func(i, j int) bool {
 		return samples[i].Timestamp < samples[j].Timestamp
 	})
-	throttled = (samples[numberOfSamples-1].Throttled - samples[0].Throttled)
-	total = samples[numberOfSamples-1].Total - samples[0].Total
 
+	resetIndexes := []int{}
```
ading1977, line 394: Just to confirm. If I understand correctly, a map is not needed here because counter reset is so rare, and even if there is a reset, it would be extremely unlikely to have multiple resets between two discoveries, right?

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -169,7 +174,11 @@ func aggregateThrottlingSamples(samples [][]metrics.ThrottlingCumulative) (float
 	}
 	avgThrottledOverall := float64(0)
 	if totalOverall > 0 {
-		avgThrottledOverall = throttledOverall * 100 / totalOverall
+		if throttledOverall > totalOverall {
+			avgThrottledOverall = 100
```
ading1977, line 178: Thanks for addressing this, can we log a warning message here?

pkg/discovery/dtofactory/general_builder.go
```diff
@@ -235,7 +240,11 @@ func (builder generalBuilder) metricValue(entityType metrics.DiscoveredEntityTyp
 		}
 
 		if total > 0 {
-			metricValue.Avg = throttled * 100 / total
+			if throttled > total {
+				metricValue.Avg = 100
```
ading1977, line 244: Can we log a warning message here?

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -152,17 +152,19 @@ func (m *KubeletMonitor) scrapeKubelet(node *api.Node) {
...
-		glog.Warningf("Failed to read kubelet cadvisor metrics for %s, %v.", node.Name, err)
-	}
-	if _, found := metricFamilies[kubeclient.ContainerCPUThrottledTotal]; !found {
-		glog.V(3).Infof("No throttling metrics found for node %s.", node.Name)
-	}
-
 	// TODO Use time stamp attached to the discovered CPUStats/MemoryStats of node and pod from kubelet to be more precise
 	currentMilliSec := time.Now().UnixNano() / int64(time.Millisecond)
-	m.generateThrottlingMetrics(metricFamilies, currentMilliSec)
+	if utilfeature.DefaultFeatureGate.Enabled(features.ThrottlingMetrics) {
```
ading1977, line 157: Should we disable by default for private preview?

pkg/discovery/dtofactory/general_builder.go
irfanurrehman: I didn't give it much thought wrt performance while implementing this, but your arguments are all valid. Additionally this slice would atmost be size 10 max (unless the sample collection per discovery is configured to something outrageously low), where the additional memory and  the lookup overhead of a map might result in similar (if not worse) performance.

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
irfanurrehman: Done!

pkg/discovery/dtofactory/general_builder.go
irfanurrehman: Done!

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
irfanurrehman: The feature flag that we implemented, we did decide to keep only disabling as an option (as far as I remember for brevity) for kubeturbo, ie. be able to disable a given metrics collection only if need be. I have left this as is in this PR. If we want to enhance the feature flag, and have features disabled by default I can extend the feature flag and update this one to false in a separate PR. Does that make sense?
@enlinxu you can also comment on this please.

pkg/discovery/dtofactory/general_builder.go
```diff
@@ -381,16 +390,64 @@ func aggregateContainerThrottlingSamples(entityID string, samples []metrics.Thro
...
-	throttled = (samples[numberOfSamples-1].Throttled - samples[0].Throttled)
-	total = samples[numberOfSamples-1].Total - samples[0].Total
 
+	resetIndexes := []int{}
+	// calculate peak
 	for i := 0; i < numberOfSamples-1; i++ {
-		total := samples[i+1].Total - samples[i].Total
+		if samples[i+1].Total < samples[i].Total || samples[i+1].Throttled < samples[i].Throttled {
+			// This probably means the counter was reset for some reason
+			// we ignore this sample diff for our calculations
```
chlam4, line 399: @irfanurrehman Another really creative idea that takes advantage of the higher sampling rate to handle counter resets without needing to know the counter max!

In terms of calculating the average, the current optimized implementation consists of two loops and maintains the resetIndexes slice, kind of a little complex.  I wonder if it'd make sense to trade that off with a simplified implementation to use the first loop and accumulate the average within that as well, eliminating the second loop as well as the resetIndexes.  I understand that in most cases the 2nd loop will actually not be executed, but the introduced code complexity seems to me overweight the performance benefit - IMHO one extra add operation in each iteration is very minimum performance-wise so I'd vote for the simplicity.

Btw, about the frequency of counter resets, if I calculate it correctly, one will only happen when a container is up for 13 years - I wonder if one would ever live that long!

pkg/discovery/dtofactory/general_builder_test.go
```diff
@@ -265,12 +265,70 @@ func TestMetricValueWithThrottlingCumulativePoints(t *testing.T) {
...
+	// time                  t1   t2    t3     t4      t5                  t6     t7    t8               t9     t10
+	// throttled samples =   1    3     5      6       11                  0      1     3                1       3
+	// total samples =       5    8     10     15      25                  0      5     8                5       8
+	// diff                   d1    d2     d3      d4          d5              d6    d7       d8              d9
+	// throttled =           (3-1) (5-3)  (6-5)   (11-6)  (reset/ignore 0-11) (1-0) (3-1) (reset/ignore 1-3) (3-1)
+	// total =               (8-5) (10-8) (15-10) (25-15) (reset/ignore 0-25) (5-0) (8-5) (reset/ignore 5-8) (8-5)
+	// Window                |_____________w1___________|                     |___w2____|                    |_w3_|
+
+	// Avg = ((11-1)+(3-0)+(3-1))*100/((25-5)+(8-0)+(8-5)) = 48.387
+	// Peak = (5-3)*100/(10-8) = 100
```
chlam4, line 331: Very nice test example!

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
chlam4: @evat-pm can chime in here, but I think she wants throttling metrics by default which is in the acceptance criteria too.

enlinxu: @irfanurrehman We may want to negate the logic for feature gate to better serve the private preview purpose, we can talk more later.
@chlam4 In terms of the default here, I want to make sure the market part is private preview only. So I need to confirm with Shravan if a separate feature flag will be used for the analysis side.

irfanurrehman: @enlinxu completely agree interms of the feature gate feature. I will create a separate task (outside the current story) to update the feature gates this way, while also disabling the throttling metrics collection by default.

pkg/features/features.go
```diff
@@ -19,6 +19,12 @@ const (
 	//
 	// Persistent volumes support.
 	PersistentVolumes featuregate.Feature = "PersistentVolumes"
+
+	// owner: @irfanurrehman
+	// beta:
+	//
+	// Persistent volumes support.
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/discovery/dtofactory/general_builder.go
```diff
@@ -381,16 +390,64 @@ func aggregateContainerThrottlingSamples(entityID string, samples []metrics.Thro
...
+			}
 		}
 		peak = math.Max(peak, throttledPercent)
 	}
+
+	lastReset := 0
+	// calculate avg percentage
+	//
+	// we actually don't need this len(resetIndexes) <= 0 check and we will have
+	// same results keeping the code only in the else block. however, the conter
```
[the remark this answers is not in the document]
irfanurrehman: Updated!

pkg/discovery/dtofactory/general_builder.go
irfanurrehman: @chlam4 you are awesome.. :-) Updated!

pkg/discovery/dtofactory/general_builder_test.go
irfanurrehman: Thanks on later thoughts added sample at time t11 to ensure the code works as expected if the last sample itself is reset (ignores the last sample).

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
enlinxu: Thanks @irfanurrehman . Agreed that let's handle it separately. I have talked to Shravan and he's looking into adding the feature flag on server side and will get back on Monday. If he's able to add feature flag in market, then we can keep this enabled by default!

pkg/discovery/dtofactory/general_builder.go
chlam4: Wow, very clever implementation even in terms of calculating the average.  Kudos!

chlam4 (approved):

enlinxu (approved):

PR #549, opened 2021-04-23
Added support to specify communication binding channel when registering probe and adding target

chlam4:
Enhanced Kubeturbo to inject the k8s service id as the communication binding channel during both probe registration and target creation so to leverage the capability to bind the target with the probe.  This also means that we could get rid of the target type suffix which was the old way of sticking the target with the probe.

**Overall solution**
- If the target type is configured, then that will be used as the probe type suffix; this is the current behavior.
- If the target type is not configured, then a normalized probe type will be used (e.g. "Kubernetes" for kubeturbo probe); this is a new behavior.
Please see a more detailed summary below:
kubeturbo-binding-channel

This is also translated to a simple adjustment in Kubeturbo to adopt the default target type ("Kubernetes") if no target type is specified in the configuration, rather than using the target name as the target type.

**Tests done** (with corresponding changes in [turbo-go-sdk](https://github.com/turbonomic/turbo-go-sdk/pull/120) and [turbo-api](https://github.com/turbonomic/turbo-api/pull/11)):
- Tested adding a new target using the binding channel method
- Tested upgrade scenarios with different combinations of target type, target name and whether credentials are specified; please see the attached screenshot
kubeturbo-binding-channel-test-scenarios
- Tested triggering discovery requests on one of the targets with the normalized probe type, and ensured that the request only went to the specific probe/cluster to prove that the binding channel method is working.
- Triggered discoveries multiple times on the same target and confirmed they all went to the same probe.

========
Note: go.mod dependencies will be adjusted once the [turbo-go-sdk PR](https://github.com/turbonomic/turbo-go-sdk/pull/120) is merged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/prometheus/client_golang v1.7.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
-	github.com/turbonomic/turbo-go-sdk v0.0.0-20210326125330-5d184a461727
+	github.com/turbonomic/turbo-go-sdk v0.0.0-20210513033620-9b641b508d10
 )
 
 // k8s and relevant dependencies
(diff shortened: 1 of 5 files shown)


ading1977 (approved):
LGTM

PR #550, opened 2021-04-27
Update step3_turbo_serviceAccountRoleBinding_admin_sample.yaml

evat-pm:
took out namespace because it is irrelevant to CRB


evat-pm:
@esara or @enlinxu could either of you please approve?  This is just clean up because a namespace is not needed for a cluster role binding resource, and looks naive :)

esara (approved):

PR #560, opened 2021-05-21
Change the containerSpec to containers relation as controlled by

irfanurrehman:
Implements https://vmturbo.atlassian.net/browse/OM-71018.
This seems to be the only change that is required on kubeturbo.

Server side the supply chain traversal still works, but the merging of actions on to the parent doesn't. An executable action is generated only on the container. Checking/updating the same will be done as part of `OM-71023`

image


irfanurrehman:
Updated the PR and the action merging and execution seems to be working fine after the update:

Generated merged actions:
image

Rest of the testing details are updated against https://vmturbo.atlassian.net/browse/OM-71023

irfanurrehman:
Have updated to remove portions of code responsible for pods by parent and controllers by parent groups, effectively implementing https://vmturbo.atlassian.net/browse/OM-71174 also.
The test results are updated against https://vmturbo.atlassian.net/browse/OM-71174

enlinxu (approved):

PR #564, opened 2021-05-30
Setting VCPU_THROTTLING commodity's util threshold to 5%

chlam4:
This is equivalent to setting the effective capacity to 5% so to maintain a low level of throttling.  This requires the turbo-go-sdk changes: https://github.com/turbonomic/turbo-go-sdk/pull/121, and we will update the go.mod dependency once that PR is merged.

Testing done - with this change, the VCPU_THROTTLING commodity DTO now contains the util threshold of 5%:
```
      {
        "commodityType": "VCPU_THROTTLING",
        "used": 45.48651817116061,
        "capacity": 100,
        "peak": 100,
        "active": true,
        "resizable": true,
        "utilizationThresholdPct": 5
      }
```

Also confirmed that the topology broadcast now has the effective capacity set to 5% for VCPU_THROTTLING:
```
                {
                  "commodityType": {
                    "type": 119
                  },
                  "used": 45.07545484462413,
                  "peak": 99.81884057971014,
                  "capacity": 100.0,
                  "effectiveCapacityPercentage": 5.0,
                  "isResizeable": true,
                  "isThin": true,
                  "active": true,
                  "historicalUsed": {
                    "maxQuantity": 46.02606187207799,
                    "histUtilization": 43.10684585571289,
                    "timeSlot": []
                  },
                  "historicalPeak": {
                    "histUtilization": 93.86170959472656,
                    "timeSlot": []
                  },
                  "displayName": "",
                  "aggregates": []
                }
```


pkg/discovery/dtofactory/general_builder.go
```diff
@@ -33,6 +33,7 @@ var (
 		metrics.StorageAmount:      proto.CommodityDTO_STORAGE_AMOUNT,
 		metrics.VCPUThrottling:     proto.CommodityDTO_VCPU_THROTTLING,
 	}
+	vcpuUtilThreshold = 5.0
```
marvelyue, line 36: I may miss previous discussions around this threshold. Have we considered making it configurable or just leave it as a fixed value?

pkg/discovery/dtofactory/general_builder.go
```diff
@@ -164,6 +165,7 @@ func (builder generalBuilder) getSoldResourceCommodityWithKey(entityType metrics
 		// This is better then separately posting the capacity into metrics sync
 		// and then reading it here.
 		commSoldBuilder.Capacity(100)
+		commSoldBuilder.UtilizationThresholdPct(vcpuUtilThreshold) // keeping throttling low
```
marvelyue, line 168: The changes look good. Can you update the corresponding unit test for this additional field?

pkg/discovery/dtofactory/general_builder.go
chlam4: Good question.  For the time being as this is a global probe-wise config, it does make sense to make it configurable.  However, we're considering to adjust the threshold individually based on say the limit of the container.  With that in mind, my thought process is not to externalize the parameter at the moment to minimize handling of migration etc.  What do you think?

pkg/discovery/dtofactory/general_builder.go
chlam4: Yes, updated!

pkg/discovery/dtofactory/general_builder.go
marvelyue: I see. Yeah, it would better and smarter to have the threshold adjusted based on the limit of the container. Sounds good to start with such small utilThreshold value and collect more customers' feedback to see how it works.

Nit: would it be better to name this variable as `vcpuThrottlingUtilThreshold` to clearly indicate this is for VCPUThrottling not VCPU utilization?

marvelyue (approved):

pkg/discovery/dtofactory/general_builder.go
chlam4: Yes totally!  I thought that's what I had; thanks for pointing it out.  Changed.

marvelyue: Thanks Cheuk!

PR #569, opened 2021-06-15
Option to specify image pull secret for cpu freq job busybox image

irfanurrehman:
Implements https://vmturbo.atlassian.net/browse/OM-71032

@esara for your reference I updated the helm charts too, reusing the `imagePullSecret` already there. You will need to update the operator image with the next release.

@evat-pm The usage for this option is as below:
1. With yamls alone:
a cmd line argument needs to be added to kubetubro deployment yaml, ie. `--busybox-image-pull-secret=<secret-name>`.
I have updated the yamls, with this argument in a comment.
2. With operator:
once the new operator is built and published (by @esara), the already existing `imagePullSecret` field value is used for both the kubeturbo imagePullSecrets and  busybox job imagePullSecrets.

### DIFF
--- a/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
+++ b/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
@@ -10,6 +10,7 @@ image:
   tag: 8.1.6
   pullPolicy: IfNotPresent
 #  busyboxRepository: busybox
+#  imagePullSecret: ""
 
 annotations:
   kubeturbo.io/controllable: "false"
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -137,6 +137,9 @@ type VMTServer struct {
 	// Busybox image uri used for cpufreq getter job
 	BusyboxImage string
 
+	// Name of the secret that stores the image pull credentials of cpu freq getter job image
+	BusyboxImagePullSecret string
+
 	// CpufreqJobExcludeNodeLabels is used to specify node labels for nodes to be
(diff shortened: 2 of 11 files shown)


deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
```diff
@@ -10,6 +10,7 @@ image:
   tag: 8.1.6
   pullPolicy: IfNotPresent
 #  busyboxRepository: busybox
+#  imagePullSecret: ""
```
esara, line 13: this may not be necessary, but if you add this, do you want to make deploy/kubeturbo/values.yaml to match this?

esara (approved):

deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
irfanurrehman: @esara updated. Meanwhile, I was struggling to understand why the helm-charts are replicated at 2 places, ie. `deploy/kubeturbo` and `deploy/kubeturbo-operator/helm-charts/kubeturbo`. If that is to have the charts at a more reasonable path for users to access (`deploy/kubturbo`), a link instead might be more useful to ensure the 2 don't go out of synk for whatever reason.
Another question I had was ther version of operator-sdk you are using now and the set of steps you use to publish the same. I think it will be a good reference for us to be able to handle this if need be, and ensure we don't do stuff any different than you do it.. :-)

esara: you are right, if there is a way to link, we should do that (the reason it exist in both places, is because we support both helm based and operator based deployments)

PR #570, opened 2021-06-18
kubeturbo certified operator 8.2.2

esara:


evat-pm:
Thanks @esara for pointing this out.
Looks good, just for kubeturbo CRD, we need to document **busyboxRepository** to support an alternate image repo for busybox (- --busybox-image=registry.access.redhat.com/ubi8/ubi-minimal). which I see in the values.yaml: #  busyboxRepository: busybox

Regarding commit 572 and 571 - these are still needed. I do not see the updates required to support straight yaml deployments - hit an issue with the field, so we know SEs are using these: https://github.com/turbonomic/kubeturbo/blob/master/deploy/kubeturbo_yamls/step5_turbo_kubeturboDeploy.yaml and https://github.com/turbonomic/kubeturbo/blob/master/deploy/kubeturbo_yamls/turbo_kubeturbo_full.yaml which is what is captured in commit 572 and 571 respectively.  Should we incorporate these into 570 or handle separately?

esara:
let me make the crd changes part this pr, too - I am still testing this anyway

evat-pm (approved):
Looks good!

PR #571, opened 2021-06-21
Update turbo_kubeturbo_full.yaml

evat-pm:
v8.2.2 and added all arg options


esara:
consider that change that are already part of https://github.com/turbonomic/kubeturbo/pull/570

PR #572, opened 2021-06-21
Update step5_turbo_kubeturboDeploy.yaml

evat-pm:
v8.2.2 and options for args


esara:
consider that change that are already part of https://github.com/turbonomic/kubeturbo/pull/570

PR #575, opened 2021-07-15
Fix handling of stop channnel addressing a possible race condition

irfanurrehman:
- Fixes https://vmturbo.atlassian.net/browse/OM-71964
- Improves the overall handling of stopping all the monitoring (kubelet and cluster) goroutines when forced to stop (eg at a timeout).
- Enhances the dummy monitoring worker used for testing to mimic the actual workers.

***Tests***
The updated unit tests pass
Long running kubeturbo continues with valid discovery and mornitoring cycles, with all entities discovered fine.

### DIFF
--- a/pkg/discovery/monitoring/dummy_monitor.go
+++ b/pkg/discovery/monitoring/dummy_monitor.go
@@ -1,10 +1,14 @@
 package monitoring
 
 import (
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/monitoring/types"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/task"
-	"time"
 )
 
 // This implementation is for Test only
@@ -23,25 +27,52 @@ func (config DummyMonitorConfig) GetMonitoringSource() types.MonitoringSource {
 type DummyMonitor struct {
 	config     *DummyMonitorConfig
 	metricSink *metrics.EntityMetricSink
+	wg         sync.WaitGroup
 }
 
 func NewDummyMonitor(config *DummyMonitorConfig) (*DummyMonitor, error) {
 	return &DummyMonitor{
 		config:     config,
-		metricSink: metrics.NewEntityMetricSink()}, nil
+		metricSink: metrics.NewEntityMetricSink(),
+	}, nil
+}
+
+func (m *DummyMonitor) reset() {
+	m.metricSink = metrics.NewEntityMetricSink()
 }
 
 func (m *DummyMonitor) GetMonitoringSource() types.MonitoringSource {
-	return types.ClusterSource
+	return types.DummySource
 }
 
 func (m *DummyMonitor) ReceiveTask(task *task.Task) {
+	m.reset()
 }
 
-func (m *DummyMonitor) Stop() {
-}
-
-func (m *DummyMonitor) Do() *metrics.EntityMetricSink {
-	time.Sleep(time.Second * time.Duration(m.config.TaskRunTime))
+func (m *DummyMonitor) Do(stopChan <-chan struct{}) *metrics.EntityMetricSink {
+	glog.V(4).Infof("%s has started task.", m.GetMonitoringSource())
+	err := m.ActualDummyWork(stopChan)
+	if err != nil {
+		glog.Errorf("Failed to execute task: %s", err)
+	}
+	glog.V(4).Infof("%s monitor has finished task.", m.GetMonitoringSource())
 	return m.metricSink
+}
+
+func (m *DummyMonitor) ActualDummyWork(stopChan <-chan struct{}) error {
+	numWorkers := 4 // some arbitrary number of parallel routines
+	for i := 0; i < numWorkers; i++ {
+		m.wg.Add(1)
+		go func() {
+			defer m.wg.Done()
+			select {
+			case <-stopChan:
+				return
+			default:
+				time.Sleep(time.Second * time.Duration(m.config.TaskRunTime))
+			}
+		}()
+	}
+	m.wg.Wait()
+	return nil
 }
--- a/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
+++ b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
@@ -35,8 +35,6 @@ type KubeletMonitor struct {
 
 	metricSink *metrics.EntityMetricSink
 
-	stopCh chan struct{}
-
 	wg sync.WaitGroup
 
 	// Whether this kubelet monitor runs during full discovery
@@ -48,14 +46,12 @@ func NewKubeletMonitor(config *KubeletMonitorConfig, isFullDiscovery bool) (*Kub
 		kubeletClient:   config.kubeletClient,
 		kubeClient:      config.kubeClient,
 		metricSink:      metrics.NewEntityMetricSink(),
-		stopCh:          make(chan struct{}, 1),
 		isFullDiscovery: isFullDiscovery,
 	}, nil
 }
 
 func (m *KubeletMonitor) reset() {
 	m.metricSink = metrics.NewEntityMetricSink()
-	m.stopCh = make(chan struct{}, 1)
 }
 
 func (m *KubeletMonitor) GetMonitoringSource() types.MonitoringSource {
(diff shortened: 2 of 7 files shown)


ading1977:
Hi @irfanurrehman , I can understand the new logic, but I don't understand why there could be a `close of closed channel` error in the first place? Do you have an analysis of the sequence of events that lead to the closing of `m.stopCh` twice in the current logic?

irfanurrehman:
@ading1977 yes. I should have posted that in the first place.

So the [monitoring workers](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L87) persist across the lifecycle of kubeturbo and the same workers (one set for full discovery and another set for sampling discovery) are used at each discovery cycle.

Now the actual monitors, for example kubelet monitor [stored the stop channel in the objects (interfaces) mentioned above](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L38), kind of making it global (the same monitors are used for each discovery).

When a new discovery was trigerred, this stopChan was reinitialized to a new variable via [receiveTask](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L205) (which would call reset and [reinitiale the stopChan](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L58)). This channel was used to stop the monitors running goroutines via [stop](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L241) which would [cancel the runnings tasks](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L241).

The issue is that this can be accessed from multiple discoveries, for example the monitoring discovery which took more then 60 seconds and so it times out, while at the same time a new monitoring discovery is initialized (each minute) which causes the race situation with stopChan. 

Taking a possible situation as example:
The current monitoring discovery (MD1) times out but hasn't got a chance of calling [w.Stop()](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L241) yet, while at the same time a new monitoring discovery is trigerred (MD2) and calls [w.ReceiveTask()](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L205), reinitialising the stopChan. Immediately after this MD1 gets a chance to continue with w.Stop actually closing the just initialised channel (not the one it should have, which is [lost to garbage collection](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L58)). MD2 continues with the wait on [this channel finishing immediately because its already closed](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L58) and goes on to [close it again](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L88).

My update ensures that this channels is written only from one goroutine. I did also skim the discovery part of the code to find similar situations with channels, I find the code unecessarily complicated but I think we are ok with channels atleast in this part of the code.

irfanurrehman:
Rebased!

ading1977:
Thanks @irfanurrehman for the detailed analysis. I am still not completely sure about the race condition. 

>The current monitoring discovery (MD1) times out but hasn't got a chance of calling w.Stop() yet, while at the same time a new monitoring discovery is trigerred (MD2) and calls w.ReceiveTask(), reinitialising the stopChan.

I wonder how is this possible? We have logic in our code, where before a worker can finish its current task and return itself to the worker pool, it cannot receive another task:

```go
func (worker *k8sDiscoveryWorker) RegisterAndRun(dispatcher *Dispatcher, collector *ResultCollector) {
	dispatcher.RegisterWorker(worker)
	for {
		// wait for a Task to be submitted
		select {
		case currTask := <-worker.taskChan:
			logLevel := glog.Level(2)
			if !worker.isFullDiscoveryWorker {
				logLevel = glog.Level(3)
			}

			glog.V(logLevel).Infof("Worker %s has received a task %v.", worker.id, currTask)
			result := worker.executeTask(currTask)
			if collector != nil {
				collector.ResultPool() <- result
			}
			glog.V(logLevel).Infof("Worker %s has finished the task %v.", worker.id, currTask)
			dispatcher.RegisterWorker(worker)
		}
	}
}
```
In each worker, we have [waitgroup](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/worker/k8s_discovery_worker.go#L185) implemented such that the `worker.executeTask` won't finish until all the sub-monitoring workers are done, this includes waiting until timeout is [handled](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/worker/k8s_discovery_worker.go#L249) , by which time, the [w.Stop()](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/worker/k8s_discovery_worker.go#L241) should have been called already.

Maybe what really is happening is that the [w.Stop()](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/worker/k8s_discovery_worker.go#L241) for the current task is indeed called before the [w.ReceiveTask](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L205) is called for the next task, but because we are not guarding the access of [m.StopCh](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L88), so before it is closed, it gets reset by the new worker?

ading1977:
Based on the above analysis, I believe the new change should work, because it makes sure that `worker.executeTaks` can close the `stopCh` before it returns with no opportunity for race condition.

ading1977 (approved):

irfanurrehman:
> Maybe what really is happening is that the [w.Stop()](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/worker/k8s_discovery_worker.go#L241) for the current task is indeed called before the [w.ReceiveTask](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/k8s_discovery_worker.go#L205) is called for the next task, but because [m.StopCh](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L88) is being notified in a [separate](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L99) go routine, and there may be some delay before it is finally closed in the [deferred](https://github.com/turbonomic/kubeturbo/blob/3184b7f8555131e0ad929ac2d9884affb8cc911b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go#L88) function, making it possible to be closed after it is reset.

Thanks @ading1977 for taking a look.
You are almost right. The channel in question is a buffered channel 
```
func (m *KubeletMonitor) reset() {
	m.metricSink = metrics.NewEntityMetricSink()
	m.stopCh = make(chan struct{}, 1)
}
```
and write (not close) to channel is used to notify the individual monitoring workers
```
func (m *KubeletMonitor) Stop() {
	m.stopCh <- struct{}{}
}
```
Which means the individual monitoring goroutine (created from MD0) will still persist as 1 write (buffer value) will free up only one sub goroutine and the rest would still go on doing their jobs. During this time the upper level finishes and restarts the monitoring worker, rekicking the discovery and thus reinitialising  the channel (`m.stopCh = make(chan struct{}, 1)`) from MD1. The below `defer func()` is eventually called when remaining sub goroutines finish (from MD0) actually closing the new reinitialised (from MD1) channel, while MD1 is still at work. 
```
func (m *KubeletMonitor) RetrieveResourceStat() error {
	defer func() {
		close(m.stopCh)
	}()
	if m.nodeList == nil || len(m.nodeList) == 0 {
		return errors.New("Invalid nodeList or empty nodeList. Finish Immediately...")
	}
	m.wg.Add(len(m.nodeList))

	for _, node := range m.nodeList {
		go func(n *api.Node) {
			defer m.wg.Done()
			select {
			case <-m.stopCh:
				return
			default:
				m.scrapeKubelet(n)
			}
		}(node)
	}

	m.wg.Wait()

	return nil
}
```
All the MD1's sub goroutines all get notified (because its a close, not a write) immediately and call MD1's defer, closing this chan again.

Anywhichways, yes the changes ensure that none of such situation can be used. The pattern used for the channels to notify workers to finish up was wrong. I did correct it.

enlinxu (approved):

PR #576, opened 2021-07-15
Change go version to 1.16

ading1977:
Change Go version to 1.16 to pass security scan.

* Change the go version from `1.14` to `1.16` in `.travis.yml`
* I had to update the `go.mod`, `go.sum` and `vendor` directory to include dependency for `github.com/mattn/goveralls@HEAD`, otherwise, the build will fail


marvelyue:
Hi @ading1977 , shall we update Go version also for turbo-go-sdk?

ading1977:
> Hi @ading1977 , shall we update Go version also for turbo-go-sdk?

You are right, I should change `turbo-go-sdk` and `turbo-api` too.

enlinxu (approved):

marvelyue (approved):

PR #557, opened 2021-05-18
Refactor handling of unschedulable nodes

irfanurrehman:
Implements https://vmturbo.atlassian.net/browse/OM-70839

The generated topology does not have any access commodity with key`schedulable` against a node or a pod.
No placement policies are created additionally to handle the same.
image

No move/placement actions generated for the pods on unschedulable nodes:
```
NAME              STATUS                     ROLES                  AGE    VERSION
pt-k8s-master-1   Ready,SchedulingDisabled   control-plane,master   571d   v1.20.4
pt-k8s-master-2   Ready,SchedulingDisabled   control-plane,master   571d   v1.20.4
pt-k8s-master-3   Ready,SchedulingDisabled   control-plane,master   571d   v1.20.4
pt-k8s-node-1     Ready                      worker                 571d   v1.20.4
pt-k8s-node-2     Ready                      worker                 571d   v1.20.4
pt-k8s-node-3     Ready                      worker                 571d   v1.20.4
```
image
 
unschedulable nodes are marked `availableForPlacement: false`
```
      "providerPolicy": {
        "availableForPlacement": false
      },
```

PR #590, opened 2021-08-13
Exclude the unschedulable node taint from taints processor handling

irfanurrehman:
Fixes https://vmturbo.atlassian.net/browse/OM-74036

As part of the unschedulable node handling refactor in https://github.com/turbonomic/kubeturbo/pull/557 we removed the logic which earlier would use VMPM_ACCESS commodities named `schedulable` to ensure the pods on unschedulable nodes don't get moved out. We used flag `AvailableForPlacement` in place.
Post that change, if a node is marked unschedulable via a taint using key `node.kubernetes.io/unschedulable`, the taint processor treats it as any regular taint and still handles it the way it handles other taints, leading to the pods on the unschedulable node being moved out. This PR excludes this specific taint from be handled by taints processor.

Please note that there are couple of other taints k8s uses to mark similar conditions especially [use these to evict pods](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions):
node.kubernetes.io/not-ready
node.kubernetes.io/unreachable
node.kubernetes.io/memory-pressure
node.kubernetes.io/disk-pressure
node.kubernetes.io/pid-pressure
node.kubernetes.io/network-unavailable
node.kubernetes.io/unschedulable
node.cloudprovider.kubernetes.io/uninitialized

We don't skip others the same way we skip `node.kubernetes.io/unschedulable` because:
1. All others except `node.kubernetes.io/unschedulable` are applied by k8s in situations which are problematic to the node and will mostly be used with effect `NoExecute` forcing eviction anyways.
2. If the evictions are not configured or a user or a user managed subsystem applies these with the effect  `NoSchedule`, its still beneficial for turbo to recommend the pod move out like any other normal taint applied on a node not tolerated by a pod on it.

*Tests performed*:
No pod move out generated with a node marked unschedulable
```
Irfans-MacBook-Pro:kubeturbo irfanurrehman$ ae2 get nodes
NAME                                STATUS                     ROLES    AGE    VERSION
ae-cluster2-master-gro-e0bce0d0b3   Ready                      master   437d   v1.16.3
ae-cluster2-node-group-6b6dc365c4   Ready                      <none>   437d   v1.16.3
ae-cluster2-node-group-8f961899c6   Ready,SchedulingDisabled   <none>   437d   v1.16.3
ae-cluster2-node-group-d37ee354b9   Ready                      <none>   437d   v1.16.3
ae-cluster2-node-group-d618c2418b   Ready                      <none>   437d   v1.16.3
```

image
image

### DIFF
--- a/pkg/discovery/worker/compliance/taint_toleration_processor.go
+++ b/pkg/discovery/worker/compliance/taint_toleration_processor.go
@@ -10,7 +10,11 @@ import (
 	"github.com/golang/glog"
 )
 
-const schedAccessCommodityKey string = "schedulable"
+// k8s marks a node as unschedulable when it sees a taint with below key by setting
+// the field spec.unschedulable: true.
+// we handle the unschedulable nodes by marking them with AvailableForPlacement = false
+// so we don't need taints processor to handle this taint via VMPM_ACCESS saperately.
+const unschedulableNodeTaintKey string = "node.kubernetes.io/unschedulable"
 
 type NodeAndPodGetter interface {
 	GetAllNodes() ([]*api.Node, error)
@@ -161,7 +165,8 @@ func getTaintCollection(nodes map[string]*api.Node) map[api.Taint]string {
 		taints := node.Spec.Taints
 
 		for _, taint := range taints {
-			if taint.Effect == api.TaintEffectNoExecute || taint.Effect == api.TaintEffectNoSchedule {
+			if !isUnschedulableNodeTaint(taint) && (taint.Effect == api.TaintEffectNoExecute ||
+				taint.Effect == api.TaintEffectNoSchedule) {
 				taintCollection[taint] = taint.Key + "=" + taint.Value + ":" + string(taint.Effect)
 				glog.V(2).Infof("Found taint (comm key = %s): %+v)", taintCollection[taint], taint)
 			}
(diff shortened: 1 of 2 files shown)


enlinxu:
Thanks @irfanurrehman  Would you please add a unit test to cover?

pkg/discovery/worker/compliance/taint_toleration_processor.go
```diff
@@ -173,6 +178,13 @@ func getTaintCollection(nodes map[string]*api.Node) map[api.Taint]string {
 	return taintCollection
 }
 
+func isUnschedulableNodeTaint(taint api.Taint) bool {
+	if taint.Key == unschedulableNodeTaintKey {
+		return true
+	}
+	return false
```
marvelyue, line 185: Nit: this can be simplified to:
```
return taint.Key == unschedulableNodeTaintKey
```

marvelyue (commented):
Thanks @irfanurrehman, just a nit comment. And agree with Enlin it would be more helpful to add an additional unit test

pkg/discovery/worker/compliance/taint_toleration_processor.go
irfanurrehman: Updated!

irfanurrehman:
@enlinxu @marvelyue thanks for taking a look. Updated!

enlinxu (approved):

PR #593, opened 2021-08-24
Filter out container CPU throttling samples when corresponding CPU limits do not match container spec VCPU capacity

marvelyue:
JIRA: https://vmturbo.atlassian.net/browse/OM-74062

# Intent
The intent is to properly collect container CPU throttling data samples on container spec by including only samples with same CPU limits as container spec capacity. This is to make sure the data samples sent to Turbo platform are correctly attached to corresponding CPU capacity, which will help drive proper VCPU resizing action results based on CPU throttling

# Problem
There was a ping-pong issue of VCPU resizing based on CPU throttling moving statistics. Corresponding containers kept resizing up and down between 2 values, which is not good.

A container in low CPU limits (500m) had very high CPU throttling. It was first resized up to 2200m and then resized up to 3000m. Afterwards, it was resized between 2200m and 3000m:
Screen Shot 2021-08-12 at 1 10 08 PM

Here're moving stats records:
```
statistic_records {
  entity_oid: 74083536599111
  throttling_record {
    ...
    capacity_records {
      vcpu_capacity: 2200
      last_sample_timestamp: 1628771698072
      sample_count: 42
      throttling_max_sample: 18.376645291933851
      throttling_fast_moving_average: 3.3498132213017553
      throttling_fast_moving_variance: 55.522075582729961
      throttling_slow_moving_average: 15.936646250332029
      throttling_slow_moving_variance: 39.1511103873128
    }
    capacity_records {
      vcpu_capacity: 3000
      last_sample_timestamp: 1628775297803
      sample_count: 54
      throttling_max_sample: 0.90144230769230771
      throttling_fast_moving_average: 0.021976738282271948
      throttling_fast_moving_variance: 0.016923339719339438
      throttling_slow_moving_average: 0.0036044069183694943
      throttling_slow_moving_variance: 0.0029061986192307908
    }
    active_vcpu_capacity: 3000
```

# Diagnosis
Noticed the moving statistics record when capacity is 2200 is the problem. From grafana dashboard above, there's no very high CPU throttling when capacity became 2200m and the high pike happened when capacity was low (500m). However moving averages are very high when `vcpu_capacity` is 2200.

Each main discover only discovers latest running containers so any changes on the containers within 2 discovery cycle should be fine. This issue could happen when rolling update happens during the main discovery on one replica or multiple replicas. For example container `foo2` (with CPU limits 2200 and low CPU throttling) starts but container `foo1` hasn't terminated yet (still running with CPU limits 500 and high throttling). ContainerSpec VCPU capacity would be 2200m in this case (largest CPU capacity of container replicas). In this case, we'll aggregate high throttling samples from container `foo1` to final CPU throttling usage of this container spec., which leads to high CPU throttling usage of this container spec although CPU throttling is pretty low with CPU limits 2200m.

# Solution
We should filter out the CPU throttling data samples when corresponding CPU limit is different from final ContainerSpec VCPU capacity.

1. In `kubelet_client.go`, collect "container_spec_cpu_quota" and "container_spec_cpu_period" from cadvisor.
2. In `kubelet_monitor.go`, parse, calculate and store CPU limits for each CPU throttling data sample by: `cpu_limits = 1000 * container_spec_cpu_quota / container_spec_cpu_period`
3. In `container_spec_entity_dto_builder.go`, when aggregating CPU throttling data, include only CPU throttling samples if corresponding CPU limit matches ContainerSpec VCPU capacity.
4. Added additional unit tests

# Testing Done
The issue can be easier to reproduce using multiple replicas.

## Environment
1. Kept 2 instances running with same version of Turbo XL server and added same kubeturbo target. One ran kubeturbo with the fix and the other didn't.
2. Set replicas of a ContainerSpec to 8 with low CPU limits (10m) which triggered high CPU throttling.
3. Update CPU limits in deployment to 300m, which resulted in low CPU throttling.
4. During the rolling update of this deployment, some pods were restarted with new CPU limits (300m) and others were still running old CPU limits (10m)
5. Manually triggered rediscovery on this target on both instances during the rolling update to make sure 2 instances collect same set of throttling data in kubeturbo
6. After the discovery, triggered broadcast on both instances and dumped throttling moving statistics data.

## Evidence
1. Moving statistics data without the fix:
```
statistic_records {
  entity_oid: 74101254870497
  throttling_record {
    capacity_records {
      vcpu_capacity: 10
      last_sample_timestamp: 1629775225876
      sample_count: 2
      throttling_max_sample: 85.580182529335076
      throttling_fast_moving_average: 84.265749276427641
      throttling_fast_moving_variance: 0.0060041834107298266
      throttling_slow_moving_average: 84.261577098565411
      throttling_slow_moving_variance: 0.000501141825574166
    }
    capacity_records {
      vcpu_capacity: 300
      last_sample_timestamp: 1629775529419
      sample_count: 1
      throttling_max_sample: 80.918097754293257
      throttling_fast_moving_average: 80.918097754293257
      throttling_fast_moving_variance: 0
      throttling_slow_moving_average: 80.918097754293257
      throttling_slow_moving_variance: 0
    }
    active_vcpu_capacity: 300
  }
}
```
where `moving_averages` are pretty high when `vcpu_capacity` is 300 (where throttling is supposed to be very low)

2. Moving statistics data with the fix:
```
statistic_records {
  entity_oid: 74103952674104
  throttling_record {
    capacity_records {
      vcpu_capacity: 10
      last_sample_timestamp: 1629775285661
      sample_count: 2
      throttling_max_sample: 84.314765694076044
      throttling_fast_moving_average: 84.308177416849389
      throttling_fast_moving_variance: 0.0070013798542989817
      throttling_slow_moving_average: 84.314215104739631
      throttling_slow_moving_variance: 0.00058511276246782091
    }
    capacity_records {
      vcpu_capacity: 300
      last_sample_timestamp: 1629775514734
      sample_count: 1
      throttling_max_sample: 0
      throttling_fast_moving_average: 0
      throttling_fast_moving_variance: 0
      throttling_slow_moving_average: 0
      throttling_slow_moving_variance: 0
    }
    active_vcpu_capacity: 300
  }
}
```
where `moving_averages` are very low when `vcpu_capacity` is 300, which is correct.

Furthermore, such log messages got printed with the fix:
```
I0823 23:25:13.707675   48517 container_spec_entity_dto_builder.go:214] Container data sample with CPU limits 10 collected at timestamp 1629775344665 doesn't match VCPU capacity 300 for ContainerSpec 82344272-0e46-41f5-a1bd-72406e6781fb/resize-up-low-cpu. Skip this data sample.
I0823 23:25:13.707830   48517 container_spec_entity_dto_builder.go:214] Container data sample with CPU limits 10 collected at timestamp 1629775404902 doesn't match VCPU capacity 300 for ContainerSpec 82344272-0e46-41f5-a1bd-72406e6781fb/resize-up-low-cpu. Skip this data sample.
I0823 23:25:13.707858   48517 container_spec_entity_dto_builder.go:214] Container data sample with CPU limits 10 collected at timestamp 1629775464646 doesn't match VCPU capacity 300 for ContainerSpec 82344272-0e46-41f5-a1bd-72406e6781fb/resize-up-low-cpu. Skip this data sample.
```

### DIFF
--- a/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
@@ -100,15 +100,23 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 		}
 		commSoldBuilder := sdkbuilder.NewCommodityDTOBuilder(commodityType)
 
-		// Aggregate container replicas utilization data.
-		// Note that the returned dataIntervalMs is not the real sampling interval because there could be multiple data
-		// points from container replicas discovered at the same time in one set of data samples. This is just a calculated
-		// equivalent interval between 2 data points to be fed into percentile based algorithm in Turbo server side.
-		// We don't need utilization data for VCPU Throttling and average aggregation strategies are not applicable.
 		if resourceType == metrics.VCPUThrottling {
+			// We don't need utilization data for VCPU Throttling and average aggregation strategies are not applicable.
 			typedUsed, ok := resourceMetrics.Used.([][]metrics.ThrottlingCumulative)
 			if ok {
-				aggregatedCap, aggregatedUsed, aggregatedPeak := aggregateThrottlingSamples(containerSpecMetrics.ContainerSpecId, typedUsed)
+				containerCPUMetrics, exists := containerSpecMetrics.ContainerMetrics[metrics.CPU]
+				if !exists {
+					glog.Errorf("CPU metrics do not exist in resource metrics for ContainerSpec %s", containerSpecMetrics.ContainerSpecId)
+					continue
+				}
+				// Get max container CPU value as containerSpec VCPU capacity which matches the logic in
+				// container_usage_data_aggregator#Aggregate
+				containerSpecVCPUCapacity := float64(0)
+				for _, containerVCPUCapacity := range containerCPUMetrics.Capacity {
+					containerSpecVCPUCapacity = math.Max(containerSpecVCPUCapacity, containerVCPUCapacity)
+				}
+				aggregatedCap, aggregatedUsed, aggregatedPeak :=
+					aggregateThrottlingSamples(containerSpecMetrics.ContainerSpecId, containerSpecVCPUCapacity, typedUsed)
 				commSoldBuilder.Capacity(aggregatedCap)
 				commSoldBuilder.Peak(aggregatedPeak)
 				commSoldBuilder.Used(aggregatedUsed)
@@ -117,6 +125,10 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 				glog.Warningf("Invalid throttling metrics type: expected: [][]metrics.ThrottlingCumulative, got: %T.", resourceMetrics.Used)
 			}
 		} else {
+			// Aggregate container replicas utilization data.
+			// Note that the returned dataIntervalMs is not the real sampling interval because there could be multiple data
+			// points from container replicas discovered at the same time in one set of data samples. This is just a calculated
+			// equivalent interval between 2 data points to be fed into percentile based algorithm in Turbo server side.
 			utilizationDataPoints, lastPointTimestamp, dataIntervalMs, err := builder.containerUtilizationDataAggregator.Aggregate(resourceMetrics)
 			if err != nil {
 				glog.Errorf("Error aggregating %s utilization data for ContainerSpec %s, %v",
@@ -161,10 +173,12 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics 
 // Throttled values is the percentage of overall throttled counter sum wrt the overall total counter sum
 // across all containers. Peak is the peak of peaks, ie. the peak of individual container peaks calculated
 // from diff of subsequent samples per container.
-func aggregateThrottlingSamples(containerSpecId string, samples [][]metrics.ThrottlingCumulative) (float64, float64, float64) {
+func aggregateThrottlingSamples(containerSpecId string, containerSpecVCPUCapacity float64, samples [][]metrics.ThrottlingCumulative) (float64, float64, float64) {
 	var throttledOverall, totalOverall, peakOverall float64
 	for _, singleContainerSamples := range samples {
-		containerThrottled, containerTotal, containerPeak, ok := aggregateContainerThrottlingSamples("", singleContainerSamples)
+		// Include container samples only if corresponding CPU limit is same as containerSpec VCPU capacity.
+		filteredContainerSamples := filterContainerThrottlingSamples(containerSpecId, singleContainerSamples, containerSpecVCPUCapacity)
+		containerThrottled, containerTotal, containerPeak, ok := aggregateContainerThrottlingSamples("", filteredContainerSamples)
 		if !ok {
 			// We don't have enough samples to calculate this value.
 			continue
@@ -185,4 +199,21 @@ func aggregateThrottlingSamples(containerSpecId string, samples [][]metrics.Thro
 	}
 
 	return 100, avgThrottledOverall, peakOverall
+}
+
+func filterContainerThrottlingSamples(containerSpecId string, singleContainerSamples []metrics.ThrottlingCumulative, containerSpecVCPUCapacity float64) []metrics.ThrottlingCumulative {
+	var filteredSamples []metrics.ThrottlingCumulative
+	for _, sample := range singleContainerSamples {
+		if int(math.Round(sample.CPULimits)) == int(math.Round(containerSpecVCPUCapacity)) {
+			filteredSamples = append(filteredSamples, sample)
+		} else if sample.CPULimits != 0 {
+			// Log a message only when CPU limit of container sample is not 0, which means CPU limit is defined for corresponding
+			// container.
+			// When CPU limit is not defined on a container, VCPU capacity is node VCPU capacity. We don't need to log
+			// such valid case.
+			glog.V(3).Infof("Container data sample with CPU limits %v collected at timestamp %v doesn't match VCPU capacity %v for ContainerSpec %s. Skip this data sample.",
+				sample.CPULimits, sample.Timestamp, containerSpecVCPUCapacity, containerSpecId)
+		}
+	}
+	return filteredSamples
 }
--- a/pkg/discovery/metrics/metric.go
+++ b/pkg/discovery/metrics/metric.go
@@ -212,6 +212,8 @@ type ThrottlingCumulative struct {
 	Throttled float64
 	// Cumulative total number of runnable periods for the resource
 	Total float64
+	// Container CPU limits when collecting this set of throttling metrics
+	CPULimits float64
 	// Time at which the metric value is collected
 	Timestamp int64
 }
--- a/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
+++ b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
@@ -162,33 +162,51 @@ func (m *KubeletMonitor) generateThrottlingMetrics(metricFamilies map[string]*dt
 	parsedMetrics := parseMetricFamilies(metricFamilies)
 	for metricID, tm := range parsedMetrics {
 		if tm != nil {
-			glog.V(4).Infof("Throttling Metrics for container: %s, cpuThrottled: %.3f, cpuTotal: %.3f.", metricID, tm.cpuThrottled, tm.cpuTotal)
-			m.genThrottlingMetrics(metrics.ContainerType, metricID, tm.cpuThrottled, tm.cpuTotal, timestamp)
+			glog.V(4).Infof("Throttling Metrics for container: %s, cpuThrottled: %.3f, cpuTotal: %.3f, cpuQuota: %.3f, cpuPeriod: %.3f",
+				metricID, tm.cpuThrottled, tm.cpuTotal, tm.cpuQuota, tm.cpuPeriod)
+			m.genThrottlingMetrics(metrics.ContainerType, metricID, tm, timestamp)
 		}
 	}
 }
 
 type throttlingMetric struct {
+	// container_cpu_cfs_throttled_periods_total
 	cpuThrottled float64
-	cpuTotal     float64
+	// container_cpu_cfs_periods_total
+	cpuTotal float64
+	// container_spec_cpu_quota
+	cpuQuota float64
+	// container_spec_cpu_period
+	cpuPeriod float64
 }
 
-// parseMetricFamilies parses the incoming prometheus format metric from two metric families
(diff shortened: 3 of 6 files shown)


dblinn:
Awesome work and summary @marvelyue, the behavior you describe in your reproduction sounds exactly like the issue we saw that caused the ping-pong problem which gives me high confidence you have the issue nailed. The moving statistics records you included from your reproduction case are very convincing.

dblinn (commented):
A small nit otherwise looks good to me. Maybe wait for @chlam4's review too before merging because he's more familiar with the kubeturbo codebase than I am.

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -100,15 +100,23 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics
...
+				if !exists {
+					glog.Errorf("CPU metrics do not exist in resource metrics for ContainerSpec %s", containerSpecMetrics.ContainerSpecId)
+					continue
+				}
+				// Get max container CPU value as containerSpec VCPU capacity which matches the logic in
+				// container_usage_data_aggregator#Aggregate
+				containerSpecVCPUCapacity := float64(0)
+				for _, containerVCPUCapacity := range containerCPUMetrics.Capacity {
+					containerSpecVCPUCapacity = math.Max(containerSpecVCPUCapacity, containerVCPUCapacity)
+				}
```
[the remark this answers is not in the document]
marvelyue: Thanks @dblinn! Good point. I just noticed we actually have an existing utilization function to get the max capacity. Will use it here.

dblinn (approved):
LGTM, but would be good to get approval from Cheuk or someone else more familiar with the codebase too.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -218,19 +236,28 @@ func parseMetricFamilies(metricFamilies map[string]*dto.MetricFamily) map[string
...
+				if metricName == kubeclient.ContainerCPUTotal {
+					tm.cpuTotal = metric.Counter.GetValue()
+				} else {
+					tm.cpuThrottled = metric.Counter.GetValue()
+				}
+			} else if metricFamily.GetType() == dto.MetricType_GAUGE {
+				if metricName == kubeclient.ContainerCPUQuota {
+					tm.cpuQuota = metric.Gauge.GetValue()
+				} else {
+					tm.cpuPeriod = metric.Gauge.GetValue()
```
chlam4, line 259: Perhaps slightly more robust to have another if statement to ensure it is indeed cpu period?  I.e.,
```
				if metricName == kubeclient.ContainerCPUQuota {
					tm.cpuQuota = metric.Gauge.GetValue()
				} else if metricName == kubeclient.ContainerCPUPeriod {
					tm.cpuPeriod = metric.Gauge.GetValue()
                                 } else {
                                          // log error
                                 }
```

Or maybe even better to use switch:
```
switch(metricName) {
  case kubeclient.ContainerCPUQuota:
    tm.cpuQuota = metric.Gauge.GetValue()
  case kubeclient.ContainerCPUPeriod:
    tm.cpuPeriod = metric.Gauge.GetValue()
  default:
    // log error
}
```

Same on line 253.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -162,44 +162,62 @@ func (m *KubeletMonitor) generateThrottlingMetrics(metricFamilies map[string]*dt
...
+// # TYPE container_spec_cpu_period gauge
+// container_spec_cpu_period{container="metallb-speaker",id="/kubepods/pod8266a379-dd56-42f8-8af0-19fc0d8ea3af/8e1a2ff0f116c9d086af53cbd7430dced0e70fed104aea79e3891870564aed38",image="sha256:8c49f7de2c13b87026d7afb04f35494e5d9ce6b5eeeb7f8983d38e601d0ac910",name="k8s_metallb-speaker_metallb-speaker-m29mf_ccp_8266a379-dd56-42f8-8af0-19fc0d8ea3af_16",namespace="ccp",pod="metallb-speaker-m29mf"} 100000 1629775404902
 //
 //out:
 // map[string]*throttlingMetric{
 //		"ccp/metallb-speaker-m29mf/metallb-speaker": {
 //			cpuThrottled: 5,
 //			cpuTotal:     10,
+//          cpuQuota:     10000,
+//          cpuPeriod:    100000,
```
chlam4, line 209: Nit: indent not aligned with lines 206-207?

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -466,11 +493,16 @@ func (m *KubeletMonitor) parseContainerStats(pod *stats.PodStats, timestamp int6
 	return totalUsedCPU, totalUsedMem, allMetricsMissing
 }
 
-func (m *KubeletMonitor) genThrottlingMetrics(etype metrics.DiscoveredEntityType, key string, throttled, total float64, timestamp int64) {
+func (m *KubeletMonitor) genThrottlingMetrics(etype metrics.DiscoveredEntityType, key string, tm *throttlingMetric, timestamp int64) {
+	cpuLimits := float64(0)
+	if tm.cpuQuota != 0 && tm.cpuPeriod != 0 {
+		cpuLimits = tm.cpuQuota * 1000 / tm.cpuPeriod
+	}
```
chlam4, line 500: To test this logic, should we add some test cases to cover the zero limit scenarios?

chlam4 (approved):
Second @dblinn's comments - great work @marvelyue!!

@irfanurrehman is more familiar with the code than me.  Perhaps he can also take a look.

enlinxu:
Second @dblinn and @chlam4 's comments. Amazing finding and details to reproduce the issues. Both finding and fixing the issue is not easy. Thank you @marvelyue !

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
enlinxu: Agree with Cheuk, let's check the exact match instead of else here.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -218,19 +236,28 @@ func parseMetricFamilies(metricFamilies map[string]*dto.MetricFamily) map[string
...
+					tm.cpuTotal = metric.Counter.GetValue()
+				} else {
+					tm.cpuThrottled = metric.Counter.GetValue()
+				}
+			} else if metricFamily.GetType() == dto.MetricType_GAUGE {
+				if metricName == kubeclient.ContainerCPUQuota {
+					tm.cpuQuota = metric.Gauge.GetValue()
+				} else {
+					tm.cpuPeriod = metric.Gauge.GetValue()
+				}
```
irfanurrehman, line 260: I would support the use of case here. Also I think `if metricFamily.GetType() == dto.MetricType_COUNTER` and `if metricFamily.GetType() == dto.MetricType_GAUGE` are unnecessary, as we have discarded the possibility of any other type at line 217 above.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -466,11 +493,16 @@ func (m *KubeletMonitor) parseContainerStats(pod *stats.PodStats, timestamp int6
 	return totalUsedCPU, totalUsedMem, allMetricsMissing
 }
 
-func (m *KubeletMonitor) genThrottlingMetrics(etype metrics.DiscoveredEntityType, key string, throttled, total float64, timestamp int64) {
+func (m *KubeletMonitor) genThrottlingMetrics(etype metrics.DiscoveredEntityType, key string, tm *throttlingMetric, timestamp int64) {
+	cpuLimits := float64(0)
```
irfanurrehman, line 497: s/cpuLimits/cpuLimit

pkg/discovery/metrics/metric.go
```diff
@@ -212,6 +212,8 @@ type ThrottlingCumulative struct {
 	Throttled float64
 	// Cumulative total number of runnable periods for the resource
 	Total float64
+	// Container CPU limits when collecting this set of throttling metrics
+	CPULimits float64
```
irfanurrehman, line 216: s/CPULimits/CPULimit, its a single value right

pkg/discovery/dtofactory/container_spec_entity_dto_builder.go
```diff
@@ -100,15 +100,20 @@ func (builder *containerSpecDTOBuilder) getCommoditiesSold(containerSpecMetrics
...
 			if ok {
-				aggregatedCap, aggregatedUsed, aggregatedPeak := aggregateThrottlingSamples(containerSpecMetrics.ContainerSpecId, typedUsed)
+				containerCPUMetrics, exists := containerSpecMetrics.ContainerMetrics[metrics.CPU]
+				if !exists {
+					glog.Errorf("CPU metrics do not exist in resource metrics for ContainerSpec %s", containerSpecMetrics.ContainerSpecId)
+					continue
+				}
+				// Get max container CPU value as containerSpec VCPU capacity which matches the logic in
+				// container_usage_data_aggregator#Aggregate
+				containerSpecVCPUCapacity := aggregation.GetResourceCapacity(containerCPUMetrics)
```
irfanurrehman, line 114: Is seems to be always expecting the case where the update (probably by the executed action) has increased the container spec limit. Wouldn't we be actually getting this value wrong in the case where the update has reduced the container spec capacity (say only for example from 2200m to 500m) and we are collecting this data during rolling update.

Would getting the limit (that is the capacity) from the parent spec and then use that as the authoritative capacity (rather then the max) make more sense. Post this any samples which do not match this capacity would be filtered in `filterContainerThrottlingSamples`. 

The change I am suggesting would be in `cluster_monitor.go#genNodePodsMetrics()`, where we would need to get the parent spec to derive capacity(s) from it if they are defined. All specs across all pod replicas will then get same capacity, irresespective of they being older replicas or newer ones. The throttling data from those samples where the capacity does not match this could be discarded.

On the other hand if we don't foresee any implication of using the max always (as is the case in this PR) or possibility of breaking something else by introducing this, we can discard this comment.

irfanurrehman (commented):
Great work @marvelyue. This is lgtm except the comment about the capacity aggregation.
Please let me know if I haven't understood it correctly.

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
marvelyue: Good point. Will add additional unit tests for zero limit

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
marvelyue: The indentations are correct in the IDE:
Screen Shot 2021-08-25 at 10 53 58 AM

Looks it's not properly displayed in github PR here

PR #607, opened 2021-09-24
[OM-73584]: Discovery and creation of node pools based on labels

pallavidn:
Container Platform cluster environments such as K8s  contain multiple logical groups of nodes that have the same compute resources. All the VMs or nodes in the group should be scaled at the same time.  These groups should be represented in the UI as part of the container platform cluster entity. 
This review handles the discovery and creation of node pools as part of DiscoveryResponse from the kubeturbo probe.
This change sits on top of the server side changes to recognize a new node pool group type.

Testing: 
Locally build XL with changes to the search and actions stats APIs. 
Modified API and UX components to recognize the node pool group type.
Azure and k8s target. The k8s cluster in azure contains node pools and discovered by the k8s probe.
See attached screen shot of the TopN widget in the UI

Screen Shot 2021-09-22 at 10 40 24 PM


pkg/discovery/dtofactory/application_entity_dto_builder.go
```diff
@@ -98,9 +98,9 @@ func (builder *applicationEntityDTOBuilder) BuildEntityDTO(pod *api.Pod) ([]*pro
 		appType := util.GetAppType(pod)
 		entityDTO, err := ebuilder.
 			ApplicationData(&proto.EntityDTO_ApplicationData{
-				Type:                    &appType,
-				IpAddress:               &(pod.Status.PodIP),
-				HostingNodeCpuFrequency: &nodeCPUFrequency,
+				Type:      &appType,
+				IpAddress: &(pod.Status.PodIP),
+				//HostingNodeCpuFrequency: &nodeCPUFrequency,
```
marvelyue, line 103: Is there any concern of setting `HostingNodeCpuFrequency` on Applicationdata so that you commented out? We need this field for some logic of CPU millicore support on the server side.

pkg/discovery/util/node_util.go
```diff
@@ -29,6 +29,7 @@ const (
 	// NodeLabelArch and NodeLabelArchBeta specify the arch of a node
 	NodeLabelArch     = "kubernetes.io/arch"
 	NodeLabelArchBeta = "beta.kubernetes.io/arch"
+	NodePoolAKS       = "agentpool"
```
irfanurrehman, line 32: I am assuming there is a reason we only have AKS `agentpool` in the labels list as of now?

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
```diff
@@ -0,0 +1,73 @@
...
+	for poolName, members := range nodeGroups {
+		var dtos []*proto.GroupDTO
+		var err error
+
+		groupID := fmt.Sprintf("NodePool::%s [%s]", poolName, builder.targetId)
+		displayName := fmt.Sprintf("NodePool-%s-%s", poolName, builder.targetId)
+		// static group
+		glog.V(3).Infof("Node Pool group: %s belongs to cluster [%s]",
+			displayName, builder.cluster.Name)
+		//clusterId := "d5788b67-9212-4ca2-a5bf-5fd316fb2eb5"
```
irfanurrehman, line 43: Please remove.

irfanurrehman (commented):
@pallavidn mostly looks good, few minor comments.

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
chlam4, line 43: Nit: this should be removed?

pkg/discovery/dtofactory/application_entity_dto_builder.go
chlam4: I have the same question.

chlam4 (commented):
Looks good to me except for a couple minor comments.

pkg/discovery/util/node_util.go
pallavidn: added node pool labels for GKE and EKS

irfanurrehman: @pallavidn `kubernetes.io/arch` and `beta.kubernetes.io/arch` are generic labels meant for architecture of nodes. 

There would be different label keys used for EKS and GKE to indicate their corresponding node pools. For example in EKS the labels should start with `eks.amazonaws.com` as listed [here](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html). GKE uses `cloud.google.com/gke-nodepool` as listed [here](https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools).

However, I think we can add the EKS labels probably after we have clusters which enable the pools, which I don't know if we have any (as the documentation does not list the exact key used for the same).
cc @evat-pm

pallavidn: @irfanurrehman, the arch related labels are part of existing code.
The labels I added are below - see the updated file.
@evat-pm and I created new node pool groups in GKE and EKS and these are the labels based from those node pools.
GKE uses label - "cloud.google.com/gke-nodepool" and EKS has eks.amazonaws.com but older versions of  EKS clusters also have node pools with a different key.
     ```
       // AKS Node pool label
	NodePoolAKS = "agentpool"
	// GKE Node pool label
	NodePoolGKE = "cloud.google.com/gke-nodepool"
	// EKS Node pool label
	NodePoolEKSIdentifier = "/nodegroup" //alpha.eksctl.io/nodegroup-name or eks.amazonaws.com/nodegroup
```

File: pkg/discovery/util/node_util.go [hunk 4]
```diff
@@ -185,6 +192,21 @@ func DetectNodeRoles(node *api.Node) sets.String {
 	return allRoles
 }
 
+func DetectNodePools(node *api.Node) sets.String {
+	allPools := sets.NewString()
+	for k, v := range node.Labels {
+		switch {
+		case k == NodePoolAKS && v != "":
```
enlinxu, line 199: nit: Shall we merge the conditions here with || ?

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
```diff
@@ -0,0 +1,71 @@
...
+			WithEntities(members).
+			WithDisplayName(displayName).
+			WithOwner(builder.cluster.Name).
+			Build()
+		if err != nil {
+			glog.Errorf("Failed to build Node Pool DTO node group %s: %v", groupID, err)
+			continue
+		}
+		dtos = append(dtos, dto)
+		groupDTOs = append(groupDTOs, dtos...)
```
marvelyue, line 55: This line looks redundant:
```
dtos = append(dtos, dto)
```

We could simplify to one line:
```
groupDTOs = append(groupDTOs, dto)
```

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
```diff
@@ -0,0 +1,71 @@
...
+	nodePools := builder.getNodePools()
+	if len(nodePools) == 0 {
+		glog.V(3).Infof("No node pools detected.")
+		return nil
+	}
+
+	var groupDTOs []*proto.GroupDTO
+	for poolName, members := range nodePools {
+		var dtos []*proto.GroupDTO
+		var err error
```
marvelyue, line 36: Nit: We could simply remove this line as the err can be set below by:
```
dto, err := xxx
```

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return groupDTOs
+}
+
+func (builder *NodePoolsGroupDTOBuilder) getNodePools() map[string][]string {
+	nodePools := make(map[string][]string)
+	for _, node := range builder.cluster.Nodes {
+		allPools := util.DetectNodePools(node)
+		for _, pool := range allPools.List() {
+			glog.Infof("%s node in node pool %s", node.Name, pool)
```
marvelyue, line 66: If there're lots of nodes with node pools in this cluster, this may pollute the log. Will it be better to set this to V(3) log level?

pkg/discovery/util/node_util.go
marvelyue: Nit: would it be cleaner to have a private function to check if it is valid node pool label and then simply use a if condition here without switch-case? Like:
```
func DetectNodePools(node *api.Node) sets.String {
	allPools := sets.NewString()
	for k, v := range node.Labels {
		if isValidNodePoolLabel(k) && v != "" {
			allPools.Insert(v)
		}
	}
	return allPools
}

func isValidNodePoolLabel(label string) bool {
	return label == NodePoolAKS || label == NodePoolGKE || strings.Contains(label, NodePoolEKSIdentifier)
}
```

marvelyue (commented):
Overall looks good. Just some nit comments

pkg/discovery/util/node_util.go
pallavidn: Refactored as above

pkg/discovery/util/node_util.go
```diff
@@ -185,6 +192,21 @@ func DetectNodeRoles(node *api.Node) sets.String {
...
+		if isValidNodePoolLabel(k) && v != "" {
+			allPools.Insert(v)
+		}
+	}
+	return allPools
+}
+
+func isValidNodePoolLabel(label string) bool {
+	return label == NodePoolAKS || label == NodePoolGKE ||
+		strings.Contains(label, NodePoolEKSIdentifier)
```
irfanurrehman, line 207: Query: Why are we not using the exact key string that EKS labels nodepools as.
You seem to have listed that in comments above `alpha.eksctl.io/nodegroup-name or eks.amazonaws.com/nodegroup`

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
irfanurrehman: Please update this.

irfanurrehman (approved):
@pallavidn another query and a[ missed comment fix](https://github.com/turbonomic/kubeturbo/pull/607#discussion_r719735117).
Approved anyways.

pkg/discovery/util/node_util.go
```diff
@@ -185,6 +192,21 @@ func DetectNodeRoles(node *api.Node) sets.String {
 	return allRoles
 }
 
+func DetectNodePools(node *api.Node) sets.String {
+	allPools := sets.NewString()
+	for k, v := range node.Labels {
```
irfanurrehman, line 197: @pallavidn Another query. Have you come across, or you know of any case where same node would be part of more then 1 node pool.. ? It looks like  a waste to go through all labels for each node..

pkg/discovery/util/node_util.go
pallavidn: We did not clear specification about the node labels for node groups. We came across these node pools while working different clusters. For generality, it is better to look for the nodegroup key word in the label name.

enlinxu (approved):

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
```diff
@@ -0,0 +1,71 @@
...
+		for _, pool := range allPools.List() {
+			nodePools[pool] = append(nodePools[pool], string(node.UID))
+		}
+	}
+
+	if glog.V(3) {
+		for poolName, nodes := range nodePools {
+			glog.Infof("%s: %s", poolName, nodes)
+		}
+	}
```
marvelyue, line 69: Nit: We can simply use `glog.V(3).Infof` without the if condition like:
```
glog.V(3).Infof("%s: %s", poolName, nodes)
```

marvelyue (approved):

pkg/discovery/dtofactory/node_pools_group_dto_builder.go
pallavidn: Want to avoid the iteration if the log level is not V(3)

marvelyue: Ohh I see. Make sense!

PR #621, opened 2021-11-18
updates to restricted role yamls

evat-pm:
changes required to turbo-admin and turbo-reader role straight yaml deployment.  Meng is handling operator changes in separate pull request

### DIFF
--- a/deploy/kubeturbo_yamls/turbo-admin.yaml
+++ b/deploy/kubeturbo_yamls/turbo-admin.yaml
@@ -5,37 +5,55 @@ metadata:
 rules:
   - apiGroups:
       - ""
-      - apps
-      - extensions
-      - app.k8s.io
+      - batch
     resources:
-      - nodes
       - pods
-      - deployments
-      - replicasets
-      - replicationcontrollers
-      - daemonsets
-      - statefulsets
-      - services
-      - endpoints
-      - namespaces
-      - limitranges
-      - resourcequotas
+      - jobs
     verbs:
       - '*'
   - apiGroups:
       - ""
       - apps
+      - apps.openshift.io
+      - extensions
+      - machine.openshift.io
+    resources:
+      - deployments
+      - replicasets
+      - replicationcontrollers
+      - statefulsets
+      - daemonsets
+      - deploymentconfigs
+      - machinesets
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+  - apiGroups:
+      - ""
+      - apps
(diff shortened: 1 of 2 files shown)


deploy/kubeturbo_yamls/turbo-admin.yaml
```diff
@@ -24,12 +32,10 @@ rules:
       - services
       - endpoints
       - namespaces
-      - limitranges
-      - resourcequotas
-      - daemonsets
       - persistentvolumes
       - persistentvolumeclaims
       - poddisruptionbudget
+      - applications
```
ading1977, line 38: This should be moved up to the `apiGroups` with `app.k8s.io`

ading1977 (changes requested):

ading1977:
@evat-pm , I think we should also need to add `deploymentconfigs` under `apps.openshift.io`?

evat-pm:
@ading1977 changes made

deploy/kubeturbo_yamls/turbo-admin.yaml
```diff
@@ -3,33 +3,57 @@ kind: ClusterRole
...
+      - daemonsets
+      - deploymentconfigs
+      - machinesets
     verbs:
-      - '*'
+      - get
+      - list
+      - watch
+      - update
+      - patch
```
irfanurrehman, line 33: Do use patch api call for any resource? There is no harm in keeping it, but if we are listing exact capabilities, I think this can be removed.

deploy/kubeturbo_yamls/turbo-reader.yaml
```diff
@@ -20,6 +21,15 @@ rules:
       - resourcequotas
```
irfanurrehman, line 21: As of now kubeturbo does not need `limitranges` in its capabilities.

irfanurrehman (commented):
@evat-pm few nits, else lgtm.

ading1977 (approved):

deploy/kubeturbo_yamls/turbo-reader.yaml
ading1977: `limitrange` is in the backlog to be supported

deploy/kubeturbo_yamls/turbo-admin.yaml
ading1977: As explained in #622, it doesn't hurt to keep the `patch` verb.

PR #622, opened 2021-11-19
Fix permissions in turbo-cluster-reader and turbo-cluster-admin role

ading1977:
### Motivation
The current `turbo-cluster-reader` and `turbo-cluster-admin` roles are outdated. Customers deploying `kubeturbo` using these roles will hit all kinds of permission issues. This PR attempts to update these role definitions with the minimum required privileges. 

This change affects the operator only. There is a separate PR #621 that updates the yamls. We can update that one and also the helm charts after this PR is approved. 

#### For `turbo-cluster-reader` role
* Add the following missing `resources` along with their corresponding `apiGroups`:
  * applications
  * jobs
  * cronjobs
  * statefulsets
  * daemonsets
  * deploymentconfigs
  * machinesets
  * machines
  * operatorresourcemappings

I put `operatorresourcemappings` resource in the `turbo-cluster-reader` just to get rid of the read permission error in the log.

#### For `turbo-cluster-admin` role
* Put `pods` and `jobs` into a separate section with **full** permission. Both `create` and `delete` permissions are needed for pod move and cpu frequency job
* Put the following controller resources into a separate section with `get`, `list`, `watch`, `update` and `patch` permissions
  * deployments
  * replicasets
  * replicationcontrollers
  * statefulsets
  * daemonsets
  * deploymentconfigs
  * machinesets
* Put the rest of the resources into a separate section with read only `get`, `list`, `watch` permissions

For both roles, add `nodes/metrics` and `nodes/proxy` resources to allow querying of throttling and threshold metrics.
 
For `turbo-cluster-admin` role, add `pods/log` resource to allow getting pods output for cpu frequency job.

Questions for reviewers:

Should I add the following resource to the `turbo-cluster-admin` role? I personally don't think this is needed, as customer who execute turbo-on-turbo actions through ORM most likely won't change the `kubeturbo` role, and will just use the default `cluster-admin` role that comes with the xl installer.
* `apiGroups`: `charts.helm.k8s.io`  
* `resources`: `xls`

Also, with `turbo-cluster-reader` and `turbo-cluster-admin` roles, any custom controller resources not listed in the  resources list cannot be discovered.

### Tests
* Deploy `kubeturbo` with `turbo-cluster-reader` role, observe there is no permission errors accessing read-only resources. The job based CPU frequency getter will fail as expected:
```
E1119 18:55:31.061723       1 kubelet_client.go:393] Failed to get CPU frequency from job on node ip-10-0-254-170.us-east-2.compute.internal: error creating cpufreq job for node: ip-10-0-254-170.us-east-2.compute.internal: jobs.batch is forbidden: User "system:serviceaccount:turbo:kubeturbo-meng-lab2" cannot create resource "jobs" in API group "batch" in the namespace "turbo".
```
* Deploy `kubeturbo` with `turbo-cluster-admin` role, observe there is no permission errors accessing any resources, and the job based CPU frequency getter works fine:
```
W1119 19:06:08.952211       1 kubelet_client.go:329] Failed to get CPU frequency for node ip-10-0-146-14.us-east-2.compute.internal from kubelet: "https://10.0.146.14:10250/spec" was not found. Will try job based getter.
...
...
I1119 19:06:15.361575       1 kubelet_client.go:397] CPU frequency of node ip-10-0-146-14.us-east-2.compute.internal: 3099.747 MHz.
```
* Execute a resize action with `turbo-cluster-admin` role:
```
I1119 19:47:31.385906       1 action_handler.go:356] Received an action RIGHT_SIZE for entity WORKLOAD_CONTROLLER [catalogue]
I1119 19:47:31.395784       1 workload_controller_resizer.go:165] Begin to resize workload controller robotshop/catalogue.
I1119 19:47:31.405129       1 resize_container_util.go:69] Try to update container catalogue resource limit from map[cpu:{i:{value:200 scale:-3} d:{Dec:<nil>} s:200m Format:DecimalSI} memory:{i:{value:104857600 scale:0} d:{Dec:<nil>} s:100Mi Format:BinarySI}] to map[cpu:{{300 -3} {<nil>} 300m DecimalSI} memory:{{104857600 0} {<nil>} 100Mi BinarySI}]
I1119 19:47:31.405162       1 resize_container_util.go:69] Try to update container catalogue resource limit from map[cpu:{i:{value:300 scale:-3} d:{Dec:<nil>} s:300m Format:DecimalSI} memory:{i:{value:104857600 scale:0} d:{Dec:<nil>} s:100Mi Format:BinarySI}] to map[cpu:{{300 -3} {<nil>} 300m DecimalSI} memory:{{239075328 0} {<nil>}  BinarySI}]
I1119 19:47:31.427183       1 k8s_controller_updater.go:147] Successfully updated Deployment robotshop/catalogue

```

### DIFF
--- a/deploy/kubeturbo-operator/helm-charts/kubeturbo/templates/serviceaccount.yaml
+++ b/deploy/kubeturbo-operator/helm-charts/kubeturbo/templates/serviceaccount.yaml
@@ -12,7 +12,12 @@ rules:
   - apiGroups:
       - ""
       - apps
+      - app.k8s.io
+      - apps.openshift.io
+      - batch
       - extensions
+      - machine.openshift.io
+      - turbonomic.com
     resources:
       - nodes
       - pods
@@ -26,6 +31,15 @@ rules:
       - resourcequotas
       - persistentvolumes
       - persistentvolumeclaims
+      - applications
+      - jobs
+      - cronjobs
+      - statefulsets
+      - daemonsets
+      - deploymentconfigs
+      - machinesets
+      - machines
+      - operatorresourcemappings
     verbs:
       - get
       - watch
@@ -35,6 +49,8 @@ rules:
     resources:
       - nodes/spec
       - nodes/stats
+      - nodes/metrics
+      - nodes/proxy
     verbs:
       - get
 {{- end }}
@@ -47,31 +63,55 @@ metadata:
 rules:
   - apiGroups:
       - ""
-      - apps
-      - extensions
+      - batch
     resources:
-      - nodes
       - pods
-      - deployments
-      - replicasets
-      - replicationcontrollers
+      - jobs
     verbs:
       - '*'
   - apiGroups:
       - ""
       - apps
+      - apps.openshift.io
+      - extensions
+      - machine.openshift.io
+    resources:
+      - deployments
+      - replicasets
+      - replicationcontrollers
+      - statefulsets
+      - daemonsets
+      - deploymentconfigs
+      - machinesets
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+  - apiGroups:
+      - ""
+      - apps
+      - batch
       - extensions
       - policy
+      - app.k8s.io
+      - turbonomic.com
+      - machine.openshift.io
     resources:
+      - nodes
+      - machines
       - services
       - endpoints
       - namespaces
       - limitranges
       - resourcequotas
-      - daemonsets
       - persistentvolumes
       - persistentvolumeclaims
       - poddisruptionbudget
+      - cronjobs
+      - applications
+      - operatorresourcemappings
     verbs:
       - get
       - list
@@ -81,6 +121,9 @@ rules:
     resources:
       - nodes/spec
       - nodes/stats
+      - nodes/metrics
+      - nodes/proxy
+      - pods/log
     verbs:
       - get
 {{- end }}
--- a/deploy/kubeturbo/templates/serviceaccount.yaml
+++ b/deploy/kubeturbo/templates/serviceaccount.yaml
@@ -12,7 +12,12 @@ rules:
   - apiGroups:
       - ""
       - apps
+      - app.k8s.io
+      - apps.openshift.io
+      - batch
       - extensions
+      - machine.openshift.io
+      - turbonomic.com
     resources:
       - nodes
       - pods
@@ -26,6 +31,15 @@ rules:
       - resourcequotas
       - persistentvolumes
       - persistentvolumeclaims
+      - applications
+      - jobs
+      - cronjobs
+      - statefulsets
+      - daemonsets
+      - deploymentconfigs
+      - machinesets
+      - machines
+      - operatorresourcemappings
     verbs:
       - get
       - watch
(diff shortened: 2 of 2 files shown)


chlam4 (approved):
Good question on the xls resource.  Perhaps we should resolve it into documentation.  When a user intends to use ORM to execute actions, the target set of CRDs should be added.

esara (approved):

esara:
I agree on leaving xls out, we do allow for custom roles and with orm we cannot be expected to prepare for all of them.

chlam4:
We should also test other types of actions, perhaps including the cluster scaling action via cluster api, but maybe as a follow-up task next sprint?

esara:
while we still have it, can you please also change
https://github.com/turbonomic/kubeturbo/blob/master/deploy/kubeturbo/templates/serviceaccount.yaml

ading1977:
> while we still have it, can you please also change https://github.com/turbonomic/kubeturbo/blob/master/deploy/kubeturbo/templates/serviceaccount.yaml

Yes, I will submit the change before committing. I would like @irfanurrehman and @evat-pm take a look too.

ading1977:
> We should also test other types of actions, perhaps including the cluster scaling action via cluster api, but maybe as a follow-up task next sprint?

@chlam4 , yes, this whole thing needs QA testing. @enlinxu mentioned that this will be released for `8.4.1`, so we should have time.

deploy/kubeturbo-operator/helm-charts/kubeturbo/templates/serviceaccount.yaml
```diff
@@ -45,33 +61,57 @@ kind: ClusterRole
...
+      - daemonsets
+      - deploymentconfigs
+      - machinesets
     verbs:
-      - '*'
+      - get
+      - list
+      - watch
+      - update
+      - patch
```
irfanurrehman, line 91: Do use patch api call for any resource? There is no harm in keeping it, but if we are listing exact capabilities, I think this can be removed.

deploy/kubeturbo-operator/helm-charts/kubeturbo/templates/serviceaccount.yaml
```diff
@@ -26,6 +31,15 @@ rules:
       - resourcequotas
```
irfanurrehman, line 31: As of now kubeturbo does not need `limitranges` in its capabilities.

irfanurrehman (approved):
@ading1977 few nits, else lgtm.

deploy/kubeturbo-operator/helm-charts/kubeturbo/templates/serviceaccount.yaml
ading1977: Thank @irfanurrehman for the suggestion, I think it doesn't hurt to keep the `patch` verb, it does not increase the permission in general.

deploy/kubeturbo-operator/helm-charts/kubeturbo/templates/serviceaccount.yaml
ading1977: Confirmed with Enlin, we have a backlog item to support `limitrange` during resize. So I think we should keep this.

PR #623, opened 2021-11-19
Improve logging

ading1977:
While working and testing #622, I observed a lot of noise in the log. 

I propose the following:

#### Raise log level for the following error or warning messages:
* Service does not have pod endpoints
This could be a valid case. For example, certain services have node as the endpoint. Raise the log level from V(2) to V(3).
* Cannot get controller UID for pod
This could also be valid. Raise the log level from Error to V(3).

#### Remove certain duplicated loggings

### DIFF
--- a/pkg/discovery/worker/container_spec_metrics_collector.go
+++ b/pkg/discovery/worker/container_spec_metrics_collector.go
@@ -48,7 +48,7 @@ func (collector *ContainerSpecMetricsCollector) CollectContainerSpecMetrics() []
 		if util.HasController(pod) {
 			controllerUID, err := util.GetControllerUID(pod, collector.metricsSink)
 			if err != nil {
-				glog.Errorf("Error getting controller UID from pod %s, %v", pod.Name, err)
+				glog.V(3).Infof("Cannot find controller UID for pod %s/%s, %v", pod.Namespace, pod.Name, err)
 				continue
 			}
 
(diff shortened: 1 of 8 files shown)


evat-pm:
thank you Meng for being proactive on this.

pkg/discovery/worker/container_spec_metrics_collector.go
```diff
@@ -48,7 +48,7 @@ func (collector *ContainerSpecMetricsCollector) CollectContainerSpecMetrics() []
 		if util.HasController(pod) {
 			controllerUID, err := util.GetControllerUID(pod, collector.metricsSink)
 			if err != nil {
-				glog.Errorf("Error getting controller UID from pod %s, %v", pod.Name, err)
+				glog.V(3).Infof("Cannot find controller UID for pod %s/%s, %v", pod.Namespace, pod.Name, err)
```
marvelyue, line 51: I noticed `util.GetControllerUID` func is used in `container_dto_builder` and `pod_entity_dto_builder` as well. Maybe we can only keep one log as Error (maybe the one in pod_entity_dto_builder) and set others as V(3)

marvelyue (approved):

irfanurrehman (approved):
Thanks for doing this @ading1977.

pkg/discovery/worker/container_spec_metrics_collector.go
ading1977: Make perfect sense. Raised to `V(3)` in `container_dto_builder`, and leave the one `pod_entity_dto_builder` as error.

PR #624, opened 2021-11-23
Update get pod by uid

irfanurrehman:
**Motivation**
While executing pod move actions we have sometimes observed the following error:
```
Error executing action Failed to get Pod by padre-stage/prepay-inventory-detection-11-9gmcg: Error to get pod with uid c37341ba-7518-4e60-b197-7b76f25bdd8d: field label not supported: metadata.uid 
```
We use `metadata.uid` as the filter in the pod list api, which does not seem to be supported as of now.  It probably worked when this piece of code was implemented but at some point (in some subsequent release of k8s) has been removed from default field selectors.

**Implementation**
This PR updates the code to use a list search by comparing the UID of each retrieved pod individually instead. It fixes https://vmturbo.atlassian.net/browse/OM-74949.

**Test**
Implemented a simple unit test to verify the updated function.

### DIFF
--- a/pkg/discovery/util/pod_util.go
+++ b/pkg/discovery/util/pod_util.go
@@ -204,21 +204,27 @@ func GetPodInPhase(podClient v1.PodInterface, name string, phase api.PodPhase) (
 // If pod not found, nil pod pointer will be returned without error.
 func GetPodInPhaseByUid(podClient v1.PodInterface, uid string, phase api.PodPhase) (*api.Pod, error) {
 	podList, err := podClient.List(context.TODO(), metav1.ListOptions{
-		FieldSelector: "status.phase=" + string(phase) + ",metadata.uid=" + uid,
+		FieldSelector: "status.phase=" + string(phase),
 	})
-
 	if err != nil {
-		return nil, fmt.Errorf("Error to get pod with uid %s: %s", uid, err)
+		return nil, fmt.Errorf("error getting podlist %s", err)
 	}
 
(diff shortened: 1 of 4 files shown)


ading1977 (approved):

pkg/discovery/util/pod_util.go
```diff
@@ -424,3 +441,97 @@ func GetPodEvents(kubeClient *client.Clientset, namespace, name string) (podEven
...
+		},
+	}
+
+	return &podList, nil
+}
+
+func (p *MockPodInterface) Create(ctx context.Context, pod *api.Pod, opts metav1.CreateOptions) (*api.Pod, error) {
+	return nil, nil
+}
+
```
chlam4, line 480: Just curious - why are all these test functions in this non-test file?

irfanurrehman: Good question @chlam4. It seems I cannot share code between test files from different packages in golang. [Ref](https://stackoverflow.com/questions/31794141/can-i-create-shared-test-utilities).
I can create another file named something like pod_test_util.go if that sounds better in this case. Please let me know.

chlam4: ic yes then it makes sense to create a different file for the test utils.  Thanks @irfanurrehman!

irfanurrehman: Updated!

chlam4 (approved):

PR #626, opened 2021-11-26
Support custom controllers as pod owners

irfanurrehman:
**Background**
There are instances where a pod is created, owned and/or controlled by a controller which is not one of the built in k8s workload controllers. Further there could be pods managed and/or updated by multiple controllers but owned by 1. In such cases the controllerRef (`ownerRef.controller: true`) was not set. For such cases we used to skip discovering the pod and creating a workload controller.

**Implementation**
- Update the logic to not exclude the ownership data if the `ownerRef.Controller` is `false`, if it is the only `ownerRef` in the list.
- If there are multiple owners in the list, select the one with `ownerRef.Controller` `true`.
- If there are multiple owners in the list, but none with `ownerRef.Controller` set as `true`. Select the first owner in the list.

**Testing**

The pods with customer controllers and `controllerRef` unset are discovered and their controllers are discovered as customer controllers:

All pods discovered in the namespace
image
image

Workload controller discovered as custom controller
image

### DIFF
--- a/pkg/discovery/util/owner_util.go
+++ b/pkg/discovery/util/owner_util.go
@@ -20,25 +20,38 @@ func GetOwnerInfo(owners []metav1.OwnerReference) (OwnerInfo, bool) {
 	var ownerInfo OwnerInfo
 	var ownerSet bool
 	for _, owner := range owners {
-		if owner.Controller == nil || !(*owner.Controller) {
-			glog.V(3).Infof("Owner %+v is not a managing controller.", owner)
-			continue
-		}
 		if len(owner.Kind) > 0 && len(owner.Name) > 0 && len(owner.UID) > 0 {
-			// Found a valid controller owner
+			if IsController(owner) {
+				glog.V(3).Infof("Found managing controller %+v.", owner)
+				// This owner is also the controller, so we use this.
+				return OwnerInfo{
+					Kind: owner.Kind,
+					Name: owner.Name,
+					Uid:  string(owner.UID),
+				}, true
+			}
+
 			if ownerSet {
-				glog.V(3).Infof("Multiple controller owners found: %+v", owner)
 				continue
 			}
+			ownerSet = true
 			ownerInfo = OwnerInfo{
 				Kind: owner.Kind,
 				Name: owner.Name,
 				Uid:  string(owner.UID),
 			}
-			ownerSet = true
 		}
 	}
+
+	glog.V(3).Infof("No managing controller was found, picked the first owner in list: %+v.", ownerInfo)
 	return ownerInfo, ownerSet
+}
+
+func IsController(owner metav1.OwnerReference) bool {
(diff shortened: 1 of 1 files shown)


esara (approved):

pkg/discovery/util/owner_util.go
```diff
@@ -19,13 +19,16 @@ type OwnerInfo struct {
 func GetOwnerInfo(owners []metav1.OwnerReference) (OwnerInfo, bool) {
```
enlinxu, line 19: We may want to refactor the logic here a bit. Now we removed the continue statement when the IsController is false. In this case, the ownerSet var will always return true unless the owners is empty. 
Also, this logging below will be confusing too,
if ownerSet {
				glog.V(3).Infof("Multiple controller owners found: %+v", owner)
				continue
			}
In the case when you have 3 owners , the 2nd one is the controller one, and when you iterate to the 2nd one, ownerSet is set to true already and the log will be printed, which will be confused.

irfanurrehman: Thanks @enlinxu for the diligence. Updated and improved the logging too.

pkg/discovery/util/owner_util.go
```diff
@@ -19,28 +19,46 @@ type OwnerInfo struct {
...
-			continue
+	var controllerIndex int = -1
+	for i, owner := range owners {
+		if IsController(owner) {
+			controllerIndex = i
 		}
+
 		if len(owner.Kind) > 0 && len(owner.Name) > 0 && len(owner.UID) > 0 {
-			// Found a valid controller owner
+			if controllerIndex == i {
```
enlinxu, line 29: Why can't we move the IsController(owner) check within the if len(owner.Kind) > 0 && len(owner.Name) > 0 && len(owner.UID) > 0 , so that we don't need the controllerIndex at all.

irfanurrehman: Updated!

irfanurrehman:
I did go ahead and update this PR to disable actions on the pods and containers of custom controllers.
@enlinxu @evat-pm please let me know if this needs to be done in a different way. For example, I could think of handling this by still generating actions but make them not executable, with enough information for the user to know why this action is not executable. 
 
I updated this because if we leave it as is, we will generate actions on such pods and containers, but when executed these will fail. We cannot have a generic way to execute such actions. Some controllers might have template specs or fields which reflect into the pods, and some don't. Infact an operator managing pods directly is also an example of a custom controller.

I have updated the fix in the second commit on this PR.
**Testing**
Before the fix we see actions (if I have a pod with enough cpu or mem usage, or no usage for a long time):
image

image


After the change no actions generated even with high resource usage:
image

irfanurrehman:
@enlinxu Updated as what we discussed. The actions will still be generated, but fail on execution. Will initiate a separate effort for a way to handle the action failures.

enlinxu (approved):

PR #628, opened 2021-12-01
Add csi based volumes support for stitching

irfanurrehman:
**Background**
We have a volume stitching support in place which deciphers the needed mechanism to stitch based on the provider specific fields in the volume resource. This does not include the CSI based volume provisioners and lacked the code specific to determine the stitching property if the volume resource has the csi based fields instead.

**Implementation**
Introduce a csi based uuid getter, which deciphers and uses the provider specific stitching info based on the strings in the provisioner driver fields.

This PR fixes https://vmturbo.atlassian.net/browse/OM-74800

**Testing**
Introduced unit test to verify the stitching uuid generation.

After the fix the volumes on the k8s cluster provisioned by csi driver are stitched
```
Irfans-MacBook-Pro:kubeturbo irfanurrehman$ kakspt describe pv pvc-c3db2637-965a-49fe-9fb1-945ef37507bb
Name:              pvc-c3db2637-965a-49fe-9fb1-945ef37507bb
Labels:            <none>
Annotations:       pv.kubernetes.io/provisioned-by: disk.csi.azure.com
Finalizers:        [kubernetes.io/pv-protection external-attacher/disk-csi-azure-com]
StorageClass:      default
Status:            Bound
Claim:             default/volume-auto-testing-claim
Reclaim Policy:    Delete
Access Modes:      RWO
VolumeMode:        Filesystem
Capacity:          10Gi
Node Affinity:
  Required Terms:
    Term 0:        topology.disk.csi.azure.com/zone in []
Message:
Source:
    Type:              CSI (a Container Storage Interface (CSI) volume source)
    Driver:            disk.csi.azure.com
    FSType:
    VolumeHandle:      /subscriptions/758ad253-cbf5-4b18-8863-3eed0825bf07/resourceGroups/mc_pt_group_pt_westus3/providers/Microsoft.Compute/disks/pvc-c3db2637-965a-49fe-9fb1-945ef37507bb
    ReadOnly:          false
    VolumeAttributes:      csi.storage.k8s.io/pv/name=pvc-c3db2637-965a-49fe-9fb1-945ef37507bb
                           csi.storage.k8s.io/pvc/name=volume-auto-testing-claim
                           csi.storage.k8s.io/pvc/namespace=default
                           requestedsizegib=10
                           skuname=StandardSSD_LRS
                           storage.kubernetes.io/csiProvisionerIdentity=1636960314516-8081-disk.csi.azure.com
Events:                <none>
```
image

### DIFF
--- a/pkg/discovery/stitching/volume_stitching_manager.go
+++ b/pkg/discovery/stitching/volume_stitching_manager.go
@@ -13,11 +13,32 @@ import (
 	"github.com/golang/glog"
 )
 
+/** Note on volume stitching
+ ** The uuids are picked from the volume resource fields as listed below:
+ ** These are mapped to the cloud specific entity id discovered from specific infra probe.
+ ** Azure:
+ ** k8s -> vol.Spec.AzureDisk.DataDiskURI
+ ** format -> replace `/` with `::`
+ **
+ ** AWS:
+ ** k8s -> vol.Spec.AWSElasticBlockStore.VolumeID
+ ** format -> "aws::%v::VL::%v"
+ **
+ ** Vsphere:
+ ** k8s -> vol.Spec.VsphereVolume.VolumePath
+ ** format -> as is
+ ** Note: we haven't seen a tanzu env yet
+ **
+ ** CSI:
+ ** Decipher aws or azure based on the provider specific strings in driver name
+ ** pick vol.Spec.CSI.VolumeHandle and format it according to AWS or azure stitching
+ ** uuid format as described above.
+**/
+
 const (
 	awsVolPrefix   = "aws://"
 	azureVolPrefix = "azure://"
 	awsVolFormat   = "aws::%v::VL::%v"
-	azureVolFormat = "azure::VL::%v"
 
 	path            = "path"
 	proxyVolumeUUID = "Proxy_Volume_UUID"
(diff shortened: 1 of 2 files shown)


pkg/discovery/stitching/volume_stitching_manager.go
```diff
@@ -237,3 +242,31 @@ func (azure *vsphereVolumeUUIDGetter) GetVolumeUUID(vol *api.PersistentVolume) (
 
 	return string(vol.UID), nil
 }
+
+type csiVolumeUUIDGetter struct {
+}
+
+func (azure *csiVolumeUUIDGetter) Name() string {
```
enlinxu, line 249: azure? Please fix

pkg/discovery/stitching/volume_stitching_manager.go
```diff
@@ -237,3 +242,31 @@ func (azure *vsphereVolumeUUIDGetter) GetVolumeUUID(vol *api.PersistentVolume) (
...
+	return "CSI"
+}
+
+func (azure *csiVolumeUUIDGetter) GetVolumeUUID(vol *api.PersistentVolume) (string, error) {
```
enlinxu, line 253: same as above

enlinxu:
@irfanurrehman , the stitching situation gets more complicated now after adding the csi parsing. Would you please add a commend in the beginning of the file to list all the situation for future reference, before we forget the details here.

irfanurrehman:
@enlinxu updated. Also updated the test results.

enlinxu (approved):

enlinxu:
The change looks good. Question, what about CSI for vsphere?

irfanurrehman:
> The change looks good. Question, what about CSI for vsphere?

@enlinxu I don't have an example of a CSI driver with vsphere cluster. The field that we use via CSI should ideally be the same ie. `vol.Spec.CSI.VolumeHandle`, but the stitching strategy will need to be defined with the vcenter team. We haven't even come across volume representations from vcenter. So far we did agree only on using the file path as stitching property for older vsphere clusters (like what we have in our lab), where `vmdk` files are represented as volumes from Vcenter.

PR #635, opened 2022-01-16
[OM-79497]: Make timeout adaptive according to readiness probe

enlinxu:
Issue:
Sometimes user has their own readiness probe set and it could be longer than the timeout we have. Timeout such action will be premature.

Implementation:
When executing the action, we will discover the readiness probe for each container in the pod, and use the following attributes:
FailureThreshold: Number of retries before giving up
InitialDelaySeconds: Initial delay before checking readiness
PeriodSeconds: Retry interval.

If any of the attribute is larger than our default, we are going to use that value instead.
Init Container case is not yet considered.

Test:
Created additional test assets in Turbo-ScenarioAsCode (will add to that repo later) and observed expected behavior with and without readiness probe defined.

### DIFF
--- a/pkg/action/executor/move_util.go
+++ b/pkg/action/executor/move_util.go
@@ -135,9 +135,6 @@ func genNewPodName(oldPod *api.Pod) string {
 // TODO: add support for operator controlled parent or parent's parent.
 func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName, parentKind, parentName string,
 	retryNum int, failVolumePodMoves, updateQuotaToAllowMoves bool, lockMap *util.ExpirationMap) (*api.Pod, error) {
-	retryInterval := defaultPodCreateSleep
-	failureThreshold := int32(retryNum)
-	initDelay := int32(0)
 	podClient := clusterScraper.Clientset.CoreV1().Pods(pod.Namespace)
 	podQualifiedName := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 	podUsingVolume := isPodUsingVolume(pod)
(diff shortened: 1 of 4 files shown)


pkg/action/executor/move_util.go
```diff
@@ -277,9 +281,31 @@ func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName, par
 			return nil, err
 		}
 	}
+	unstructuredContainers, found, err := unstructured.NestedSlice(parentForPodSpec.Object, "spec", "template", "spec", "containers")
+	if err != nil || !found {
+		return nil, fmt.Errorf("error retrieving containers because: %v", err)
```
irfanurrehman, line 286: Please add podname here:
something like `"%s/%s", pod.Namespace, pod.Name`

pkg/action/executor/move_util.go
```diff
@@ -277,9 +281,31 @@ func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName, par
...
+	}
+
+	for _, unstructuredContainer := range unstructuredContainers {
+		var container apicorev1.Container
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredContainer.(map[string]interface{}), &container); err != nil {
+			return nil, fmt.Errorf("error converting unstructured containers to typed containers for : %v", err)
```
irfanurrehman, line 292: As above, please add some identifying name

pkg/action/executor/move_util.go
```diff
@@ -530,6 +556,15 @@ func getPodOwnersInfo(clusterScraper *cluster.ClusterScraper, pod *api.Pod,
 	return parent, nil, nsParentClient, nil, "", nil
 }
 
+func getContainerReadinessProbeDetails(container apicorev1.Container) (failureThreshold int32, initialDelaySec int32, periodSec int32) {
+	probe := container.ReadinessProbe
+	if probe == nil {
+		glog.V(4).Infof("Cannot find readiness probe for Container: %s, use default configuration instead", container.Name)
```
irfanurrehman, line 562: suggestion:
`s/Cannot find readiness probe for/Readiness probe not found for`

pkg/action/executor/move_util.go
```diff
@@ -134,6 +135,9 @@ func genNewPodName(oldPod *api.Pod) string {
 // TODO: add support for operator controlled parent or parent's parent.
 func movePod(clusterScraper *cluster.ClusterScraper, pod *api.Pod, nodeName, parentKind, parentName string,
 	retryNum int, failVolumePodMoves, updateQuotaToAllowMoves bool, lockMap *util.ExpirationMap) (*api.Pod, error) {
+	retryInterval := defaultPodCreateSleep
```
irfanurrehman, line 138: Suggestion:
Consider moving these variables near to where they are used

irfanurrehman (commented):
Thanks @enlinxu for doing this.
Some minor nits.

ading1977 (approved):
LGTM

PR #638, opened 2022-01-19
OM-79460 Refactored to use new ConnectedEntityMessage

amd-ibm:
# Issue
Some probes are using the deprecated connections (consistsOf) instead of the new XL connection (owns). The topology processor converts the deprecated connections to the new type. Removing the need for the conversion should provide a small performance boost, as well as get us off the deprecated code.

# Test
Debugged topology processor to verify that the [conversion logic](https://git.turbonomic.com/turbonomic/xl/-/blob/staging/com.vmturbo.topology.processor/src/main/java/com/vmturbo/topology/processor/stitching/TopologyStitchingGraph.java#L372) was no longer being hit.

Pulled topology from the processor and verified it matched the data pulled before the change. 

Pulled diagnostics and verified that the entities were correct. 

Ran the unit tests.

### DIFF
--- a/pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/workload_controller_entity_dto_builder.go
@@ -62,11 +62,11 @@ func (builder *workloadControllerDTOBuilder) BuildDTOs() ([]*proto.EntityDTO, er
 				kubeController.GetFullName())
 		}
 
-		// To connect WorkloadController to ContainerSpec entity, WorkloadController consistsOf the associated ContainerSpecs.
-		// The platform will translate this into the following relation:
-		// WorkloadController owns Containers
+		// To connect WorkloadController to ContainerSpec entity, WorkloadController owns the associated ContainerSpecs.
 		containerSpecsIds := builder.getContainerSpecIds(kubeController)
-		entityDTOBuilder.ConsistsOf(containerSpecsIds)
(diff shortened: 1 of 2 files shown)


marvelyue:
Hi @amd-ibm, the changes look good. I noticed we didn't have corresponding unit test for the relation. Can you help add an additional unit test in `workload_controller_entity_dto_builder_test.go` to test the `owns` relation?

For example, in `workload_controller_entity_dto_builder_test.go`, in `createKubeController` func, we can set up a fake pod on `kubeController` like:
```
pod := &api.Pod{
		Spec: api.PodSpec{
			Containers: []api.Container{
				{
					Name: "Foo",
				},
			},
		},
	}
	kubeController.Pods = append(kubeController.Pods, pod)
```

Then in the existing `TestBuildDTOs` func, when it is "controller1-UID", check something like:
```
assert.Equal(t, 2, len(entityDTO.ConnectedEntities))
assert.Equal(t, proto.ConnectedEntity_OWNS_CONNECTION, entityDTO.ConnectedEntities[1].GetConnectionType())
assert.Equal(t, "controller1-UID/Foo", entityDTO.ConnectedEntities[1].GetConnectedEntityId())
```

amd-ibm:
> createKubeController

Added unit test.

amd-ibm:
@marvelyue , I've added the requested unit test.

marvelyue (approved):
LGTM

enlinxu (approved):

PR #643, opened 2022-02-04
DOCKER_BUILDKIT=1 flag to ensure cross-platform docker builds without buildx pre-installed

chlam4:
Problem:
We support cross-platform docker builds using the `docker buildx` plugin which has been required to install to build successfully.  That was not a problem for building on Mac because `buildx` is included in "Docker Desktop".  As we migrate to "Rancher Desktop", `buildx` isn't included by default.

Solutions:
First, I tried to follow this link https://docs.docker.com/buildx/working-with-buildx/ and install `buildx` manually but that doesn't seem to work for me.  Lucky, I noticed the `DOCKER_BUILDKIT=1` flag that just works without even installing `buildx`.  @ading1977 also tested with the flag.


dblinn (approved):

ading1977 (approved):

esara (approved):

chlam4:
A few notes about `buildkit` and `buildx` after I read more about them:
- `buildkit` was the original for multi-platform support.  `buildx` was built on top of it and became a docker CLI plugin.
- Interestingly, both are required in our case to build multi-platform images.
- Without the `DOCKER_BUILDKIT=1` flag, `make docker` will fail with the "file not found" error.
- Without the buildx CLI plugin, an image will be successfully built which seems to contain all the platforms based on the image size, but the manifest only has the default one `amd64`.  Would need to dig more to understand why.

Nevertheless, the `DOCKER_BUILDKIT=1` flag will help the normal development use case.  If one needs to build multi-platform on mac, then the `buildx` plugin is still required.

PR #646, opened 2022-02-09
Add kubernetes orchestrator property to k8s nodes

irfanurrehman:
**Intent**
Implementation of the suspend functionality in Azure probe needs identification of the VM as a regular VM or an AKS VM. This implementation aids the same.
Ref https://vmturbo.atlassian.net/browse/OM-80453

**Background**
While an AKS provisioned VM already carries a `orchestrator` or an `aks-managed-orchestrator` tag, there might be clusters where this tag is missing on the AKS VM.

Example
image

**Implementation**
This implementation adds an `orchestrator`:`Kubernetes` tag property on each of the VMs to ensure availability of the same in the  `TargetSE` of `ActionItemDTO` received in the suspend action.

**Testing**
The property is available in the tags section of the VM
image

### DIFF
--- a/pkg/discovery/dtofactory/property/node_properties.go
+++ b/pkg/discovery/dtofactory/property/node_properties.go
@@ -6,6 +6,11 @@ import (
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
 )
 
+const (
+	k8sOrchestratorName  = "orchestrator"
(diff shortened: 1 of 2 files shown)


enlinxu:
@irfanurrehman Do we know in what case the tag will be missing?

enlinxu:
Discussed with @irfanurrehman and this PR is no longer needed. Closing.

PR #647, opened 2022-02-09
[OM-79845]: Only capture annotations that match configured regular expressions.

amd-ibm:
# Intent
Resolve an issue where collected annotations were displaying sensitive information due to server-side apply.

# Implementation
Updated the configuration to allow customers to specify annotation whitelists for the various entity types (currently we only appear to pull annotations for namespaces, workload controllers, and container specs). 

Refactored the _BuildLabelAnnotationProperties_ in `property_common.go` to take separate maps of labels and annotations in order to be able to distinguish between the two (they were previously merged prior to the method invocation), and refactored the callers to pass them individually.

# Test
Manually tested with local kubeturbo build and verified annotations would/wouldn't display depending on the configured regex.


enlinxu (approved):
LGTM.

irfanurrehman (approved):
Nice work @amd-ibm!

A suggestion: Update the raw yaml samples [here](https://github.com/turbonomic/kubeturbo/tree/master/deploy/kubeturbo_yamls) please, may be with some comments/examples of usage.

amd-ibm:
> Nice work @amd-ibm!
> 
> A suggestion: Update the raw yaml samples [here](https://github.com/turbonomic/kubeturbo/tree/master/deploy/kubeturbo_yamls) please, may be with some comments/examples of usage.

Updated `step4_turbo_configMap_sample.yaml` to include `annotationWhitelist` templated config and description.

PR #660, opened 2022-03-24
Cleanup k8.io/kubernetes dependencies for eviction

kevinwangcn:
# Intent
We're trying to eliminate the dependencies on the library `k8s.io/kubernetes`

# Background
Depending on `k8s.io/kubernetes` introduces huge dependencies and causes unnecessary unknown vulnerabilities and security issue. Removing these will also potentially reduce the binary size by a reasonable amount.

# Implementation
Go through the code and check out the dependencies on the `k8s.io/kubernetes` for the kubelet eviction, copy only necessary codes into a new file `kubelet_eviction.go`, keep the logic unchanged. In this case, we will get rid of `k8s.io/kubernetes` dependency when validating discovered node threshold values 

# Test Done

**1. Get the value of the threshold from api-server**
image

**2. Dump the topology using curl command**
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' "http://localhost:8080/TopologyService/broadcastAndReturnTopology" -d '{}' > /tmp/topology-new.out
```
Search the keyword `"KubernetesNodeName": "node1"` and could see the following a little above
 ```
                  "commodityType": {
                    "type": 65,
                    "key": "k8s-node-rootfs"
                  },
                  "used": 28610.66796875,
                  "peak": 28610.66796875,
                  "capacity": 40940.0,
                  "effectiveCapacityPercentage": 89.99999985098839, <------simliar to 90%
                  "isResizeable": false,
                  "isThin": true,
                  "active": true,
                  "capacityIncrement": 1048576.0,
                  "historicalUsed": {
                    "maxQuantity": 32421.88671875,
                    "histUtilization": 28592.376953125,
                    "timeSlot": []
                  },
```
**3. The threshold is dicovered and simliar to what we see from api-server prospect**

### CODE REVIEW

File: pkg/kubeclient/kubelet_eviction.go [hunk 1]
```diff
@@ -0,0 +1,318 @@
+package kubeclient
```
irfanurrehman, line 1: Please put a comment here too similar to previous work which can reference from where is this code copied over

pkg/kubeclient/kubelet_eviction.go
```diff
@@ -0,0 +1,318 @@
...
+	Operator ThresholdOperator
+	// Value is the threshold the resource is evaluated against.
+	Value ThresholdValue
+	// GracePeriod represents the amount of time that a threshold must be met before eviction is triggered.
+	GracePeriod time.Duration
+	// MinReclaim represents the minimum amount of resource to reclaim if the threshold is met.
+	MinReclaim *ThresholdValue
+}
+
+const (
```
irfanurrehman, line 43: Please merge all the constant and variable sections into 1 each. You can choose to give a lines gap in between different set of constants to put them into different groups with similar meanings.

pkg/kubeclient/kubelet_eviction.go
irfanurrehman, line 1: Please change the filename to `eviction_utils.go`

pkg/kubeclient/kubelet_eviction.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// inodes, number. internal to this module, used to account for local disk inode consumption.
+	resourceInodes v1.ResourceName = "inodes"
+	// resourcePids, number. internal to this module, used to account for local pid consumption.
+	resourcePids v1.ResourceName = "pids"
+)
+
+const (
+	NodeAllocatableEnforcementKey = "pods"
+	SystemReservedEnforcementKey  = "system-reserved"
```
irfanurrehman, line 71: This and the next 2 constants are not used anywhere

irfanurrehman (changes requested):
Thanks @kevinwangcn for doing this. Few nits.

pkg/kubeclient/kubelet_eviction.go
kevinwangcn: done

kevinwangcn: done

irfanurrehman (approved):

PR #666, opened 2022-04-08
Refine the log message for resizing action

kevinwangcn:
# Intent
Refine the log message when resizing a workload controller

# Background

1. Action execution of workload controller resize, regardless of status (succeeded, failed, etc) should include related information of container spec(s) and namespace, commodity changes, along with the workload controller name. 
2. Log message should inform the user which specific container spec has failed the merged workload controller resize action.
3. When executing actions against multiple Workload controllers with the same name that could exist in multiple namespaces, the log message should indicate the namespace info.

# Implementation
Go through the related code and pass around the namespace, pod, or container info when logging the message

# Test Done

## Case 1
**1. Execute a resize action on a workload controller and the action gets executed successfully**
image

**2. Check the kubeturbo log**
image

## Case 2
**1. Execute a resize action on a workload controller and the action failed to get executed**
image

**2. Check the kubeturbo log**
image

### DIFF
--- a/pkg/action/executor/resize_container_util.go
+++ b/pkg/action/executor/resize_container_util.go
@@ -38,7 +38,7 @@ func updateRequests(container *k8sapi.Container, patchRequests k8sapi.ResourceLi
 	}
 
 	if changed {
-		glog.V(2).Infof("Try to update container %v for the workload controller [%v] resource request from %+v to %+v",
+		glog.V(2).Infof("Try to update container %v in the workload controller [%v] resource request from %+v to %+v",
 			container.Name, objectID, container.Resources.Requests, result)
 		container.Resources.Requests = result
 	}
@@ -66,7 +66,7 @@ func updateLimits(container *k8sapi.Container, patchCapacity k8sapi.ResourceList
 	}
 
 	if changed {
-		glog.V(2).Infof("Try to update container %v for the workload controller [%v] resource limit from %+v to %v",
+		glog.V(2).Infof("Try to update container %v in the workload controller [%v] resource limit from %+v to %v",
 			container.Name, objectID, container.Resources.Limits, result)
 		container.Resources.Limits = result
 	}
--- a/pkg/action/executor/resize_container_util.go
+++ b/pkg/action/executor/resize_container_util.go
@@ -18,7 +18,7 @@ import (
 )
 
 // update the Pod.Containers[index]'s Resources.Requests
-func updateRequests(container *k8sapi.Container, patchRequests k8sapi.ResourceList) bool {
+func updateRequests(container *k8sapi.Container, patchRequests k8sapi.ResourceList, objectID string) bool {
 	glog.V(4).Infof("Begin to update Request.")
 	changed := false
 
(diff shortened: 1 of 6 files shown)


pkg/action/executor/resize_container_util.go
```diff
@@ -66,8 +66,8 @@ func updateLimits(container *k8sapi.Container, patchCapacity k8sapi.ResourceList
 	}
 
 	if changed {
-		glog.V(2).Infof("Try to update container %v resource limit from %+v to %v",
-			container.Name, container.Resources.Limits, result)
+		glog.V(2).Infof("Try to update container %v for the workload controller [%v] resource limit from %+v to %v",
```
enlinxu, line 69: nit: Change "Try to update container %v for the workload controller [%v] resource limit from %+v to %v" to "Try to update container %v limit from %+v to %v in workload controller [%v]"

pkg/action/executor/resize_container_util.go
```diff
@@ -38,15 +38,15 @@ func updateRequests(container *k8sapi.Container, patchRequests k8sapi.ResourceLi
 	}
 
 	if changed {
-		glog.V(2).Infof("Try to update container %v resource request from %+v to %+v",
-			container.Name, container.Resources.Requests, result)
+		glog.V(2).Infof("Try to update container %v for the workload controller [%v] resource request from %+v to %+v",
+			container.Name, objectID, container.Resources.Requests, result)
```
enlinxu, line 42: nit: Change "Try to update container %v for the workload controller [%v] resource request from %+v to %+v" to Try to update container %v request from %+v to %v in workload controller [%v]

irfanurrehman:
@kevinwangcn thanks for doing this. `invalid containerIndex -1` actually means that our code could not find the container name in the spec that we have (parents pod template spec).

 I think the message we want to convey is exactly some form of that.
Can we change the `invalid containerIndex -1` to `cannot find the container <container-name> in the parents pod spec`.

Additionally we use the fully qualified names at most places for k8s resources in the logging. Please use `<namespace/resource-name>` as the format at places where you have made changes or you think needs changes in the flow.

For example
This msg
`Failed to execute action RIGHT_SIZE on WORKLOAD_CONTROLLER [cpu-resize-up]: invalid containerIndex -1, the container has the name istio-proxy and is in the namespace kevwang`
could be changed to 
`Failed to execute action RIGHT_SIZE on WORKLOAD_CONTROLLER [kevwang/cpu-resize-up]: could not find container istio-proxy in parents pod template spec. It is likely an injected sidecar.`

kevinwangcn:
> 
@irfanurrehman , please have a look at new message
image

pkg/action/executor/workload_controller_resizer.go
```diff
@@ -55,8 +56,10 @@ func (r *WorkloadControllerResizer) Execute(input *TurboActionExecutorInput) (*T
 		// build resize specification
 		spec, err := cr.buildResizeSpec(item, controllerName, podSpec, getContainerIndex(podSpec, item.GetCurrentSE().GetDisplayName()))
 		if err != nil {
-			glog.Errorf("Failed to execute resize action: %v", err)
-			return &TurboActionExecutorOutput{}, err
+			glog.Errorf("Failed to build resize spec for the container %v of the workload controller %v/%v as the reason: %v",
```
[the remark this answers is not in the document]
kevinwangcn: done

enlinxu (approved):

irfanurrehman:
Thanks @kevinwangcn

PR #668, opened 2022-04-20
kubeturbo certified operator 8.5.1

esara:


deploy/kubeturbo-operator/deploy/olm-catalog/kubeturbo-operator/kubeturbo-operator.package.yaml
```diff
@@ -1,19 +1,7 @@
 packageName: kubeturbo-certified
 channels:
   - name: stable
+    currentCSV: kubeturbo-operator.v8.5.0
+  - name: v8.5.0
     currentCSV: kubeturbo-operator.v8.4.4
```
chlam4, line 6: This should be v8.5.0?

chlam4: Actually, should the current be 8.5.1 as you have as the folder name?

esara: actually the certified operator bundle no longer uses the package file - I am not sure why I updated it

chlam4 (approved):

PR #669, opened 2022-04-25
Refine the check logic beforem applying ORM feature

kevinwangcn:
# Intent
Fix the crash issue when the work controller has the `ownerReference` but doesn't have the `Controller` field

# Background
Some deployment especially on the Cloudpak enviroment doesn't have `Controller` field in the stanza `ownerReference` when the deployment has a parent. 

# Implementation
Refine the check logic and add evaluation to check if the `Controller` pointer is `nil` or not

# Test Done

## Case 1
**1. Test with a deployment which doesn't have a `Controller` field but has a `ownerReference`**
image
**2. Execute resize action on this deployment**
image
**3. Get the resize action executed successfully**
image

## Case 2

**1. Test with ORM case to make sure the change doesn't break down the original ORM feature**
image
```
[root@localvm ~]# k get orm certmanagers.operator.ibm.com -o yaml
apiVersion: turbonomic.com/v1alpha1
kind: OperatorResourceMapping
metadata:
  creationTimestamp: "2022-04-11T21:28:40Z"
  generation: 1
  name: certmanagers.operator.ibm.com
  namespace: ibm-common-services
  resourceVersion: "418139544"
  uid: 2eda1386-17a4-4176-8a55-4091536867f2
spec:
  resourceMappings:
  - resourceMappingTemplates:
    - destPath: .spec.certManagerController.resources
      srcPath: .spec.template.spec.containers[?(@.name=="cert-manager-controller")].resources
    srcResourceSpec:
      componentNames:
      - cert-manager-controller
      kind: Deployment
  - resourceMappingTemplates:
    - destPath: .spec.certManagerCAInjector.resources
      srcPath: .spec.template.spec.containers[?(@.name=="cert-manager-cainjector")].resources
    srcResourceSpec:
      componentNames:
      - cert-manager-cainjector
      kind: Deployment
```

**2. Execute the resize action**
image

**3. Check the CertManager CR**
```
[root@localvm ~]# k get certmanagers.operator.ibm.com default  -o yaml
I0425 12:33:27.436170   35437 request.go:665] Waited for 1.012576338s due to client-side throttling, not priority and fairness, request: GET:https://api.ocp47demo.openshift-turbonomic.net:6443/apis/operator.tekton.dev/v1alpha1?timeout=32s
apiVersion: operator.ibm.com/v1alpha1
kind: CertManager
metadata:
  creationTimestamp: "2022-01-17T22:40:20Z"
  finalizers:
  - certmanager.operators.ibm.com
  generation: 23
  labels:
    app.kubernetes.io/instance: ibm-cert-manager-operator
    app.kubernetes.io/managed-by: ibm-cert-manager-operator
    app.kubernetes.io/name: cert-manager
    operator.ibm.com/opreq-control: "true"
  name: default
  resourceVersion: "444233740"
  uid: 18301b3d-270e-4cc4-91e5-849d77b17c04
spec:
  certManagerCAInjector:
    resources:
      limits:
        cpu: 300m
        memory: 392Mi
      requests:
        cpu: 20m
        memory: 282Mi
```

### DIFF
--- a/pkg/action/executor/k8s_controller.go
+++ b/pkg/action/executor/k8s_controller.go
@@ -136,13 +136,20 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 	}
 
 	controllerOwnerReferences := c.obj.GetOwnerReferences()
-	if !c.shouldSkipOperator(c.obj) && len(controllerOwnerReferences) > 0 && (*controllerOwnerReferences[0].Controller || "ClusterServiceVersion" == controllerOwnerReferences[0].Kind) {
-		// If k8s controller is controlled by custom controller, update the CR using OperatorResourceMapping
-		// if SkipOperatorLabel is not set or not true.
-		if c.ormClient == nil {
-			return fmt.Errorf("failed to execute action with nil ORMClient")
+	if !c.shouldSkipOperator(c.obj) {
+		/*There is 2 situation to determine whether to use orm
+		1)the current work-contoller has an owner and the owner has the controller field which is set to true
+		2)the current work-controller has an owner and the kind of the owner is ClusterServiceVersion
+		*/
+		isControlledByOwner := (controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller) || "ClusterServiceVersion" == controllerOwnerReferences[0].Kind
+		if isControlledByOwner {
+			// If k8s controller is controlled by custom controller, update the CR using OperatorResourceMapping
+			// if SkipOperatorLabel is not set or not true.
+			if c.ormClient == nil {
+				return fmt.Errorf("failed to execute action with nil ORMClient")
+			}
+			err = c.ormClient.Update(origControllerObj, c.obj, controllerOwnerReferences[0])
 		}
-		err = c.ormClient.Update(origControllerObj, c.obj, controllerOwnerReferences[0])
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/action/executor/k8s_controller.go
```diff
@@ -136,7 +136,15 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
...
-	if !c.shouldSkipOperator(c.obj) && len(controllerOwnerReferences) > 0 && (*controllerOwnerReferences[0].Controller || "ClusterServiceVersion" == controllerOwnerReferences[0].Kind) {
+	var isControlledByOwner bool
+	if len(controllerOwnerReferences) > 0 {
+		if controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller {
+			isControlledByOwner = true
+		} else if "ClusterServiceVersion" == controllerOwnerReferences[0].Kind {
+			isControlledByOwner = true
+		}
+	}
+	if !c.shouldSkipOperator(c.obj) && isControlledByOwner {
```
pankajbatra1, line 147: The !c.shouldSkipOperator(c.obj) could be before line 140 since there is no need to figure out isControlledByOwner if the former is false.

enlinxu (approved):

pkg/action/executor/k8s_controller.go
kevinwangcn: @pankajbatra1 , I've updated my change and put check logic after `!c.shouldSkipOperator(c.obj)`

pankajbatra1 (approved):

PR #671, opened 2022-04-26
Fix index of range issue for checking owner reference

kevinwangcn:
Add a length check to make sure the owner reference does exist

## Case 1 Resize a workload controller with one container spec
**1. Resize the deployment cert-manager-cainjector which has only one container and has the owner reference set**
```
[root@localvm ~]# k get pods cert-manager-cainjector-7cff9df4f8-djd7g 
NAME                                       READY   STATUS    RESTARTS   AGE
cert-manager-cainjector-7cff9df4f8-djd7g   1/1     Running   0          23h
```
image

**2. Check the execution result and also check the ORM CR**
image

image

## Case 2 Resize a workload controller with multiple container specs
**1. Resize the deployment auth-pap which has 2 containers**
image
**2. Check the execution result**
image

## Case 3 Resize a bare pod
**1. Resize the bare pod xxx**
image

**2. Check the execution result**
image
image

### DIFF
--- a/pkg/action/executor/k8s_controller.go
+++ b/pkg/action/executor/k8s_controller.go
@@ -136,20 +136,20 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 	}
 
 	controllerOwnerReferences := c.obj.GetOwnerReferences()
-	if !c.shouldSkipOperator(c.obj) {
-		/*There is 2 situation to determine whether to use orm
-		1)the current work-contoller has an owner and the owner has the controller field which is set to true
-		2)the current work-controller has an owner and the kind of the owner is ClusterServiceVersion
-		*/
-		isControlledByOwner := (controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller) || "ClusterServiceVersion" == controllerOwnerReferences[0].Kind
-		if isControlledByOwner {
-			// If k8s controller is controlled by custom controller, update the CR using OperatorResourceMapping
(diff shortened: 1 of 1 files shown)


enlinxu:
@kevinwangcn This is a regression so please share all the tests you have performed. Tests should include at least:
Resize a workload controller with one container spec
Resize a workload controller with multiple container specs
Resize a bare pod

Also check if you can add test case in https://github.com/turbonomic/kubeturbo/blob/master/test/integration/action_execution.go

pkg/action/executor/k8s_controller.go
```diff
@@ -141,14 +141,16 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 		1)the current work-contoller has an owner and the owner has the controller field which is set to true
 		2)the current work-controller has an owner and the kind of the owner is ClusterServiceVersion
 		*/
-		isControlledByOwner := (controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller) || "ClusterServiceVersion" == controllerOwnerReferences[0].Kind
-		if isControlledByOwner {
-			// If k8s controller is controlled by custom controller, update the CR using OperatorResourceMapping
-			// if SkipOperatorLabel is not set or not true.
-			if c.ormClient == nil {
-				return fmt.Errorf("failed to execute action with nil ORMClient")
+		if len(controllerOwnerReferences) > 0 {
```
enlinxu, line 144: What do you do if len(controllerOwnerReferences) == 0?

enlinxu (approved):

PR #674, opened 2022-05-02
Refine the check logic of whether to use ORM

kevinwangcn:
# Intent
Refine the condition check for whether to use ORM when resizing a workload controller.

# Background
When a workload controller has an owner reference, in most cases, this workload controller is controlled by the CR. The CR becomes the source of the truth.  In this case, Kubeturbo will update the CR instead of updating the workload controller.

# Implementation
Add a new function `shouldUseOrm` to determine if need to use ORM or not.
There are 3 cases need to use ORM to update CR:
1)The kind of the owner of the workload controller is  `ClusterServiceVersion` ;
2)The onwer of the workload controller has the `Controller` field and this field is set to `True`;
3)The onwer of the workload controller doesn't have the `Controller` field;

# Test Done

## Case 1 : The kind of the owner of the workload controller is  `ClusterServiceVersion`
**1. The deployment `ibm-licensing-operator` has a owner with the kind `ClusterServiceVersion`**
image

image

image

**2. Execute the resize action and check the result**
image

## Case 2: The onwer of the workload controller has the `Controller` field and this field is set to `True`, the kind of owner is not  `ClusterServiceVersion`

**1. The deployment `cert-manager-controller` has the owner, its kind is `` and its `` field is set to `True`**
image
image

**2. Execute the resize action and check the result**
image

## Case 3: The onwer of the workload controller doesn't have the Controller field

**1. Manually remove the `controller` field in the `ownerReference` of the deployment ``**
image

image

**2. Execute the resize action and check the result**
image

## Case 4: The workload controller doesn't have an `onwnerReference`

**1.  Create a deployment without any `ownerReference` and generate a resize action
image
image
**2. Execute the resize action and check the result**
image

### DIFF
--- a/pkg/action/executor/k8s_controller.go
+++ b/pkg/action/executor/k8s_controller.go
@@ -136,14 +136,7 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 	}
 
 	controllerOwnerReferences := c.obj.GetOwnerReferences()
-	/*There are 2 conditions to determine whether to use ORM to update the workload controller or directly update it
-	1)the current workload contoller has the ownerReference and the owner has the controller field which is set to true
-	2)the current workload controller has the ownerReference and the kind of the owner is ClusterServiceVersion
-	*/
-	if !c.shouldSkipOperator(c.obj) && // Check if the label kubeturbo.io/skipOperator is set or not
-		len(controllerOwnerReferences) > 0 && // Check if the workcontroller has the owner
-		((controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller) || // case 1: owner exists and has a true controller field
-			"ClusterServiceVersion" == controllerOwnerReferences[0].Kind) { // case 2: owner exists and its kind is ClusterServiceVersion
+	if !c.shouldSkipOperator(c.obj) && c.shouldUseOrm(c.obj) {
 		// If k8s controller is controlled by custom controller, update the CR using OperatorResourceMapping
 		// if SkipOperatorLabel is not set or not true.
 		if c.ormClient == nil {
@@ -175,4 +168,68 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
 
 func (rc *parentController) String() string {
 	return rc.name
+}
+
+//To determine if ORM should be used or not in terms of the owner reference of the work load controller
+func (c *parentController) shouldUseOrm(controller *unstructured.Unstructured) bool {
+	if controller == nil {
+		glog.V(3).Infof("Workload controller pointer is nil, shouldUseOrm return false")
+		return false
+	}
+	controllerOwnerReferences := c.obj.GetOwnerReferences()
+	if len(controllerOwnerReferences) <= 0 {
+		glog.V(3).Infof("Workload controller <%v/%v> doesn't have an owner reference, shouldUseOrm return false", controller.GetNamespace(), controller.GetName())
+		return false
+	}
+	/*
+		In the following 3 situations, we use trigger ORM to update the CR
+	*/
+	//Case 1: The owner of the controller has the kind as ClusterServiceVersion
+	/*
+		ownerReferences:
+		- apiVersion: operators.coreos.com/v1alpha1
+		  blockOwnerDeletion: false
+		  controller: false
+		  kind: ClusterServiceVersion
+		  name: ibm-licensing-operator.v1.14.0
+		  uid: bfd3d22d-99f9-473d-bea6-dd81b8d763ff
+		resourceVersion: "438731373"
+		uid: 6e58b1e7-dd29-41d5-8b77-194797b4fe9e
+	*/
+	if "ClusterServiceVersion" == controllerOwnerReferences[0].Kind {
+		glog.V(2).Infof("Workload controller <%v/%v> has the owner with ClusterServiceVersion kind, shouldUseOrm return true", controller.GetNamespace(), controller.GetName())
+		return true
+	}
+
+	//Case 2: The owner has the controller field and the value of the filed is set to True
+	/*
+	  ownerReferences:
+	  - apiVersion: operator.ibm.com/v1alpha1
+	    blockOwnerDeletion: true
+	    controller: true
+	    kind: IBMLicensing
+	    name: instance
+	    uid: 5882b63f-bd56-4d3b-83c7-b38405bf3429
+	*/
+	if controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller {
+		glog.V(2).Infof("Workload controller <%v/%v> has the owner controller set to true, shouldUseOrm return true", controller.GetNamespace(), controller.GetName())
+		return true
+	}
+
+	//Case 3: The owner doesn't even have the controller filed
+	/*
+		ownerReferences:
+		- apiVersion: icp4a.ibm.com/v1
+		  kind: ICP4ACluster
+		  name: icp4adeploy
+		  uid: d411a4bc-29bf-4086-9c5c-c65345256dcf
+		resourceVersion: "228138709"
+		uid: 21c125c3-4b5d-4037-8088-4b00c241e16e
+	*/
+	if controllerOwnerReferences[0].Controller == nil {
+		glog.V(2).Infof("Workload controller <%v/%v> has the owner but the owner doesn't have controller field, shouldUseOrm return true", controller.GetNamespace(), controller.GetName())
+		return true
+	}
+
+	return false
 }


pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
+	/*
+		ownerReferences:
+		- apiVersion: icp4a.ibm.com/v1
+		  kind: ICP4ACluster
+		  name: icp4adeploy
+		  uid: d411a4bc-29bf-4086-9c5c-c65345256dcf
+		resourceVersion: "228138709"
+		uid: 21c125c3-4b5d-4037-8088-4b00c241e16e
+	*/
+	if controllerOwnerReferences[0].Controller == nil {
```
enlinxu, line 229: Question:
We return true if controllerOwnerReferences[0].Controller == nil, but we return false if *controllerOwnerReferences[0].Controller == false, are we sure this is the expected behavior? (I know it's the same behavior as before)

pkg/action/executor/k8s_controller.go
```diff
@@ -136,14 +136,7 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
...
 	controllerOwnerReferences := c.obj.GetOwnerReferences()
-	/*There are 2 conditions to determine whether to use ORM to update the workload controller or directly update it
-	1)the current workload contoller has the ownerReference and the owner has the controller field which is set to true
-	2)the current workload controller has the ownerReference and the kind of the owner is ClusterServiceVersion
-	*/
-	if !c.shouldSkipOperator(c.obj) && // Check if the label kubeturbo.io/skipOperator is set or not
-		len(controllerOwnerReferences) > 0 && // Check if the workcontroller has the owner
-		((controllerOwnerReferences[0].Controller != nil && *controllerOwnerReferences[0].Controller) || // case 1: owner exists and has a true controller field
-			"ClusterServiceVersion" == controllerOwnerReferences[0].Kind) { // case 2: owner exists and its kind is ClusterServiceVersion
+	if !c.shouldSkipOperator(c.obj) && c.shouldUseOrm(c.obj) {
```
pankajbatra1, line 139: c.shouldSkipOperator does not check for c.obj being null, but c.shouldUseOrm does.  You may want to update shouldSkipOperator also

pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
 }
+
+//To determine if ORM should be used or not in terms of the owner reference of the work load controller
+func (c *parentController) shouldUseOrm(controller *unstructured.Unstructured) bool {
+	if controller == nil {
+		glog.V(3).Infof("Workload controller pointer is nil, shouldUseOrm return false")
+		return false
+	}
+	controllerOwnerReferences := c.obj.GetOwnerReferences()
+	if len(controllerOwnerReferences) <= 0 {
```
pankajbatra1, line 180: Why not have == 0 check?

pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
+		- apiVersion: operators.coreos.com/v1alpha1
+		  blockOwnerDeletion: false
+		  controller: false
+		  kind: ClusterServiceVersion
+		  name: ibm-licensing-operator.v1.14.0
+		  uid: bfd3d22d-99f9-473d-bea6-dd81b8d763ff
+		resourceVersion: "438731373"
+		uid: 6e58b1e7-dd29-41d5-8b77-194797b4fe9e
+	*/
+	if "ClusterServiceVersion" == controllerOwnerReferences[0].Kind {
```
pankajbatra1, line 199: Could controllerOwnerReferences[0].Kind  be nil?

pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
 func (rc *parentController) String() string {
 	return rc.name
 }
+
+//To determine if ORM should be used or not in terms of the owner reference of the work load controller
+func (c *parentController) shouldUseOrm(controller *unstructured.Unstructured) bool {
```
irfanurrehman, line 174: The naming usage here is mixed and confusing. Lets use `obj` as the argument name. Further, you don't really need to pass it as an argument as its already available in `c *parentController`.
I understand that some of this might be technical debt, recommend correcting it while we are at it.

pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
+	if controller == nil {
```
irfanurrehman, line 175: I don't understand the point of this check. This is checking if `c.obj` is set. We have reached here after doing lot of processing on `c.obj`, isnt it?

irfanurrehman: Did you intend to pass `ownerReferences` as the argument and check `len(controllerOwnerReferences)` and `controllerOwnerReferences[0].Controller != nil` here?

irfanurrehman (commented):
I actually stopped midway putting the comments on this review. I apologise as I could not completely grasp the intent and reason of this change. Whats the background that I am missing?

pkg/action/executor/k8s_controller.go
kevinwangcn: > I don't understand the point of this check. This is checking if `c.obj` is set. We have reached here after doing lot of processing on `c.obj`, isnt it?

This check is just for double protection to avoid crash,  @pankajbatra1  also wants me to add this in the function `shouldSkipOperator`  https://github.com/turbonomic/kubeturbo/pull/674/files#r863798437

kevinwangcn: > Did you intend to pass `ownerReferences` as the argument and check `len(controllerOwnerReferences)` and `controllerOwnerReferences[0].Controller != nil` here?

Yes,  we need to check this `Controller ` and `Kind` field in the `ownerRerence` to dertermin if we should ORM to update the workload controller

pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
+		glog.V(3).Infof("Workload controller pointer is nil, shouldUseOrm return false")
```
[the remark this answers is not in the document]
kevinwangcn: Sure, I could use verbosity 5

pkg/action/executor/k8s_controller.go
kevinwangcn: the type of the controllerOwnerReferences[0].Kind  is string, it will never be nil.

pkg/action/executor/k8s_controller.go
kevinwangcn: Sure, I could add this check in the function `shouldSkipOperator `

pkg/action/executor/k8s_controller.go
```diff
@@ -176,3 +169,67 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
 func (rc *parentController) String() string {
 	return rc.name
 }
+
+//To determine if ORM should be used or not in terms of the owner reference of the work load controller
```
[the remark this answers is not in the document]
kevinwangcn: will update it to workload

pkg/action/executor/k8s_controller.go
kevinwangcn: Make sense, I would change it to `shouldUseOrm(obj *unstructured.Unstructured)`

pkg/action/executor/k8s_controller.go
irfanurrehman: @kevinwangcn You did not understand my point. You are passing and checking the object itself here, which:
1. You don't need to pass as its already available as `c.obj`
2. You don't need to check here, as it has been used extensively before. If its nil the crash will happen earlier but not here. Further we already check for errors getting this object before [here](https://github.com/turbonomic/kubeturbo/blob/master/pkg/action/executor/k8s_controller.go#L64), so I think the nil check on `obj` is not required.
3. What you should ideally check for nil here is the `ownerReferences` (`c.obj.GetOwnerReferences()`)

irfanurrehman:
> I actually stopped midway putting the comments on this review. I apologise as I could not completely grasp the intent and reason of this change. Whats the background that I am missing?

I had a chat about this change with @kevinwangcn. I did already introduce the check on logic that is needed here as part of #626 before. I think we can use [that function](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/util/owner_util.go#L22) as is here.
This is what it checks:
- Update the logic to not exclude the ownership data if the ownerRef.Controller is false, if it is the only ownerRef in the list.
- If there are multiple owners in the list, select the one with ownerRef.Controller true.
- If there are multiple owners in the list, but none with ownerRef.Controller set as true. Select the first owner in the list.

enlinxu (approved):

irfanurrehman (approved):

PR #675, opened 2022-05-04
Add integration test

kevinwangcn:
# Intent
Adding integration test into Travis

# Background
The current CI process doesn't include integration test

# Implementation
Use kind to build a 3 nodes cluster, and run the integration test in there. 

# Test Done

## Case 1
**1. Trigger the Travis job by making any new commit and then check the Travis log**
Download the binary and create the kind cluster
image
Run the integration test
image

Report the test result, 2 openshift cases are skipped
image


irfanurrehman (approved):
Thanks @kevinwangcn for doing this.

PR #678, opened 2022-05-06
add resize test

kevinwangcn:
# Intent
Adding resize action testing case into the integration test

# Background
The current integration test doesn't validate resizing action against a workload controller

# Implementation
Create a deployment with a single container and then create a resize action aginst the deployment, the resize action will adjust the `resource/limits/cpu` by making a 100m increment. at last to check if the new deployment has right `resource/limits/cpu`

# Test Done

## Case 1
```
make integration
```
image

### DIFF
--- a/test/integration/action_execution.go
+++ b/test/integration/action_execution.go
@@ -31,7 +31,10 @@ import (
 	"github.com/turbonomic/kubeturbo/pkg/discovery/dtofactory/property"
 )
 
-const openShiftDeployerLabel = "openshift.io/deployer-pod-for.name"
+const (
+	openShiftDeployerLabel = "openshift.io/deployer-pod-for.name"
+	cpuIncrement           = 100
+)
 
 var _ = Describe("Action Executor ", func() {
 	f := framework.NewTestFramework("action-executor")
@@ -184,6 +187,26 @@ var _ = Describe("Action Executor ", func() {
 
 			validateMovedPod(kubeClient, dc.Name, "deploymentconfig", namespace, targetNodeName)
 
+		})
+	})
+
+	// Test resize action execution
+	Describe("executing resize action on a deployment", func() {
+		It("should match the expected resource cpu reuqest after resize", func() {
+
+			dep, err := createDeployResource(kubeClient, depSingleContainerWithResources(namespace, "", 1, false, false, false))
+			framework.ExpectNoError(err, "Error creating test resources")
+
+			targetSE := newResizeWorkloadControllerTargetSE(dep)
+			currentSE := newContainerEntity("test-cont")
+
+			_, err = actionHandler.ExecuteAction(newResizeActionExecutionDTO(proto.ActionItemDTO_RIGHT_SIZE, targetSE, nil, currentSE), nil, &mockProgressTrack{})
+			framework.ExpectNoError(err, "Resize action failed")
+
+			_, err = waitForDeploymentToUpdateResource(kubeClient, dep)
+			if err != nil {
+				framework.Failf("Failed to check the resource request in the new deployment: %s", err)
+			}
 		})
 	})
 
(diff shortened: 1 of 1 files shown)


test/integration/action_execution.go
```diff
@@ -704,3 +748,9 @@ func getTargetSENodeName(f *framework.TestFramework, pod *corev1.Pod) string {
 func podID(pod *corev1.Pod) string {
 	return fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
 }
+
+func validateResourceResize(oldResourceQuan, newResourceQuan resource.Quantity) {
```
enlinxu, line 752: This is not generic.  Can we make it generic by changing the signature to (oldResourceVal, newResourceVal, increment)? And it should cover both cpu and memory in case we want to implement a test case for memory resize.

kevinwangcn: @enlinxu  what do you think of this definition `func validateResourceResize(oldResourceVal, newResourceVal resource.Quantity,increment int64)`?

test/integration/action_execution.go
```diff
@@ -490,6 +515,25 @@ func waitForDeployment(client kubeclientset.Interface, depName, namespace string
 	return newDep, nil
 }
 
+func waitForDeploymentToUpdateResource(client kubeclientset.Interface, depName, namespace string, oldResourceVersion string) (*appsv1.Deployment, error) {
```
irfanurrehman, line 518: This can make the test flaky.
Controllers can sometimes update resources, resulting in resource version updates.
I would rather have a the wait on what we are expecting as the result of testing here. 
`waitForResourcesUpdated()` where the wait will be on the `dep.Spec.Template.Spec.Containers[0].Resources` to be matched to what we expect after update. This also means that `validateResourceResize` will not be needed.

test/integration/action_execution.go
```diff
@@ -187,6 +190,28 @@ var _ = Describe("Action Executor ", func() {
 		})
 	})
 
+	// Test resize action execution
+	Describe("executing resize action on a deployment", func() {
+		It("should match the expected resource cpu reuqest after resize", func() {
```
irfanurrehman, line 195: Lets add another resize action test to test a merged resize action on multiple container specs

irfanurrehman (changes requested):
@kevinwangcn thanks for doing this. Some updates needed.

test/integration/action_execution.go
kevinwangcn: @irfanurrehman  I removed the function `validateResourceResize` and put the evaluation into the wait function as suggested, please have a  look at my new commit. Thanks for your review.

test/integration/action_execution.go
kevinwangcn: @enlinxu , as Irfan suggested, I remove this function, putting the logic of the validation in the function `waitForResourcesUpdated` instead.

test/integration/action_execution.go
kevinwangcn: Regarding this, I'm going to add a new sub task in the current story, and will address it in another PR, @irfanurrehman , what do you think?

irfanurrehman: Sure, that works too.

irfanurrehman (approved):

test/integration/action_execution.go
kevinwangcn: Add a new subtask to trace this: https://vmturbo.atlassian.net/browse/OM-84505

test/integration/action_execution.go
```diff
@@ -31,7 +31,10 @@ import (
 	"github.com/turbonomic/kubeturbo/pkg/discovery/dtofactory/property"
 )
 
-const openShiftDeployerLabel = "openshift.io/deployer-pod-for.name"
+const (
+	openShiftDeployerLabel = "openshift.io/deployer-pod-for.name"
+	cpuIncrement           = 100
```
pankajbatra1, line 36: Would it make sense to also test for Request (not just limit)?

kevinwangcn: sure, it is.  I add a subtask to trace it https://vmturbo.atlassian.net/browse/OM-84522

PR #680, opened 2022-05-13
Support resize on resource/request commodity

kevinwangcn:
# Intent
Improve resize test on request commodities in the integration test, also add memory case in.

# Background
The current integration test doesn't support resize test on request commodity

# Implementation
Compose resize action with memory adjustment and set the action type to assciate with request

# Test Done

## Case 1
```
make integration
```
image

### DIFF
--- a/test/integration/action_execution.go
+++ b/test/integration/action_execution.go
@@ -201,7 +201,7 @@ var _ = Describe("Action Executor ", func() {
 	})
 
 	// Test resize action execution
-	Describe("executing resize action on a deployment", func() {
+	Describe("executing resize action on a deployment with a single container", func() {
 		It("should match the expected resource request/limit after resizing on both of cpu and memory", func() {
 
 			dep, err := createDeployResource(kubeClient, depSingleContainerWithResources(namespace, "", 1, false, false, false))
@@ -209,7 +209,7 @@ var _ = Describe("Action Executor ", func() {
 
 			targetSE := newResizeWorkloadControllerTargetSE(dep)
 
-			// Resize on limit, increasing cpu/memory limit by 100m/100Mi
+			// Resize up cpu and memory on resource/limit
 			_, err = actionHandler.ExecuteAction(newResizeActionExecutionDTO(proto.ActionItemDTO_RIGHT_SIZE, targetSE, LIMIT_SINGLE_CONTAINER), nil, &mockProgressTrack{})
 			framework.ExpectNoError(err, "Resize action on limit failed")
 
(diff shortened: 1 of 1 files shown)


enlinxu:
Is there a good reason why we introduce different method for request and limit with lots of duplicate code? If we introduce more commodities we may end up with more duplicate code. Let's please clean up the code to make it less duplicate.

kevinwangcn:
> Is there a good reason why we introduce different method for request and limit with lots of duplicate code? If we introduce more commodities we may end up with more duplicate code. Let's please clean up the code to make it less duplicate.

I'm going to add a parameter in the function to differentiate the change type instead of adding a new function

enlinxu (approved):

test/integration/action_execution.go
```diff
@@ -639,33 +678,88 @@ type mockProgressTrack struct{}
...
+		}
+		aiOnMem.CurrentSE = currentSE
 
-	newCap := cap + cpuIncrement
-	ai.NewComm = &proto.CommodityDTO{
-		CommodityType: &commType,
-		Capacity:      &newCap,
+		dto.ActionItem = []*proto.ActionItemDTO{aiOnCpu, aiOnMem}
+	default:
+		framework.Errorf("The change type<%d> doesn't get supported", changeType)
```
pankajbatra1, line 760: doesn't get -> is not

test/integration/action_execution.go
```diff
@@ -192,20 +202,29 @@ var _ = Describe("Action Executor ", func() {
...
+			framework.ExpectNoError(err, "Resize action on limit failed")
 
-			_, err = waitForDeploymentToUpdateResource(kubeClient, dep)
+			_, err = waitForDeploymentToUpdateResource(kubeClient, dep, LIMIT_SINGLE_CONTAINER)
 			if err != nil {
-				framework.Failf("Failed to check the resource request in the new deployment: %s", err)
+				framework.Failf("Failed to check the change of resource/limit in the new deployment: %s", err)
+			}
+
+			// Resize on request, increasing cpu/memory request by 100m/100Mi
```
pankajbatra1, line 221: The comment here should not be specific to the actual test.

test/integration/action_execution.go
```diff
@@ -639,33 +678,88 @@ type mockProgressTrack struct{}
...
-	commType := proto.CommodityDTO_VCPU
-	cap := float64(cpuIncrement)
-	ai.CurrentComm = &proto.CommodityDTO{
-		CommodityType: &commType,
-		Capacity:      &cap,
-	}
+	switch changeType {
+	case REQUEST_SINGLE_CONTAINER:
+		currentSE := newContainerEntity("test-cont")
+		// Build action item on Cpu
```
pankajbatra1, line 687: You could refactor the code to build an action item.  There are four almost identical code blocks.

test/integration/action_execution.go
kevinwangcn: done

test/integration/action_execution.go
kevinwangcn: Add a new function `newActionItemDTO` to build the action item,  please have a look at my second commit

test/integration/action_execution.go
kevinwangcn: done

irfanurrehman (approved):

pankajbatra1 (approved):

PR #681, opened 2022-05-17
Resize down request/limit on multiple containers

kevinwangcn:
# Intent
Adding resize action testing case into the integration test, this change focuses on resize down on multiple containers

# Background
The current integration test doesn't validate resizing action against a workload controller with more than one container

# Implementation
Create a deployment with 2 containers and then create a resize down action on `resource/request` for the first container, and create a resize down action on  `resource/limit` for the second container.  At last validate those 2 containers have the right `resource/request` and `resource/limit` 

# Test Done

## Case 1
```
make integration
```
image


test/integration/action_execution.go
```diff
@@ -562,6 +645,33 @@ func waitForDeploymentToUpdateResource(client kubeclientset.Interface, dep *apps
 				newMemLimit.Value()-oldMemLimit.Value() == memIncrement*1024*1024 {
 				return true, nil
 			}
+		case REQLIM_MULTI_CONTAINER:
+			//check on the first container
+			checkContainer1 := false
+			checkContainer2 := false
+			oldCpuReq := dep.Spec.Template.Spec.Containers[0].Resources.Requests["cpu"]
```
pankajbatra1, line 652: Please refactor next four lines.  There is repetition for various cases.

test/integration/action_execution.go
```diff
@@ -562,6 +645,33 @@ func waitForDeploymentToUpdateResource(client kubeclientset.Interface, dep *apps
 				newMemLimit.Value()-oldMemLimit.Value() == memIncrement*1024*1024 {
 				return true, nil
 			}
+		case REQLIM_MULTI_CONTAINER:
+			//check on the first container
```
pankajbatra1, line 649: Move this comment down two lines

test/integration/action_execution.go
```diff
@@ -39,11 +39,14 @@ const (
 	memoryLimit            = 200
 	cpuIncrement           = 100
 	memIncrement           = 100
+	cpuDecrement           = 25
```
pankajbatra1, line 42: Please add cpuIncrement, memDecrement case.

test/integration/action_execution.go
```diff
@@ -314,6 +335,68 @@ func depSingleContainerWithResources(namespace, claimName string, replicas int32
...
+									corev1.ResourceMemory: resource.MustParse(fmt.Sprintf("%dMi", memoryRequest)),
+								},
+							},
+						},
+						{
+							Name:    "test-cont-2",
+							Image:   "busybox",
+							Command: []string{"/bin/sh"},
+							Args:    []string{"-c", "while true; do sleep 30; done;"},
+							Resources: corev1.ResourceRequirements{
```
pankajbatra1, line 380: Why not also check when each container scales differently?

test/integration/action_execution.go
```diff
@@ -543,23 +631,34 @@ func waitForDeploymentToUpdateResource(client kubeclientset.Interface, dep *apps
 		}
 		switch changeType {
 		case REQUEST_SINGLE_CONTAINER:
```
irfanurrehman, line 633: Validating this way (comparing individual values and specific methods eg. `validaIncrement()`, etc.) is unwieldy and prone to coding problems.
I would rather copy the pod template spec from the original deployment post its creation, pass it into `newResizeActionExecutionDTO()` and update the respective container resource values to the same ones (desired values) that are requested in actionExecutionDTO and simple use `reflect.DeepEqual()` to compare the `podSpecs` here.

irfanurrehman (changes requested):
@kevinwangcn thanks for doing this. Please check the comment and see if you can reasonably address it.

irfanurrehman (approved):
Thanks for doing this @kevinwangcn

enlinxu (approved):

PR #687, opened 2022-05-30
Add horizontal scale action test

kevinwangcn:
# Intent
Adding horizontal scale action execution test 

# Background
The current integration test doesn't include this case

# Implementation
Create a deployment and compose a provision action on one of its pod, check the replica number of the deployment;
Compose another suspend action and run it against the new deployment, check the replica number of the deployment;

# Test Done

## Case 1
```
make integration
./build/integration.test -k8s-kubeconfig=/root/.kube/config -k8s-context=kind-kind -ginkgo.focus="Executing horizontal scale action on a deployment"
```
image

### DIFF
--- a/test/integration/action_execution.go
+++ b/test/integration/action_execution.go
@@ -276,6 +276,48 @@ var _ = Describe("Action Executor ", func() {
 		})
 	})
 
+	// Horizontal scale test
+	Describe("Executing horizontal scale action on a deployment", func() {
+		It("should match the expected replica number", func() {
+			// create a deployment with 2 replicas
+			dep, err := createDeployResource(kubeClient, depSingleContainerWithResources(namespace, "", 2, false, false, false))
+			framework.ExpectNoError(err, "Error creating test resources")
+
+			pod, err := getDeploymentsPod(kubeClient, dep.Name, namespace, "")
+			framework.ExpectNoError(err, "Error getting deployments pod")
+			// This should not happen. We should ideally get a pod.
+			if pod == nil {
+				framework.Failf("Failed to find a pod for deployment: %s", dep.Name)
+			}
+
+			targetNodeName := getTargetSENodeName(f, pod)
+			if targetNodeName == "" {
+				framework.Failf("Failed to find a pod for deployment: %s", dep.Name)
+			}
+
+			// Test the provision action
+			_, err = actionHandler.ExecuteAction(newActionExecutionDTO(proto.ActionItemDTO_PROVISION,
+				newTargetSEFromPod(pod), nil), nil, &mockProgressTrack{})
+			framework.ExpectNoError(err, "Failed to execute provision action")
+
+			newDep, err := waitForDeployment(kubeClient, dep.Name, dep.Namespace)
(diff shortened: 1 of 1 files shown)


test/integration/action_execution.go
```diff
@@ -255,6 +255,48 @@ var _ = Describe("Action Executor ", func() {
...
+			framework.ExpectNoError(err, "Error creating test resources")
+
+			pod, err := getDeploymentsPod(kubeClient, dep.Name, namespace, "")
+			framework.ExpectNoError(err, "Error getting deployments pod")
+			// This should not happen. We should ideally get a pod.
+			if pod == nil {
+				framework.Failf("Failed to find a pod for deployment: %s", dep.Name)
+			}
+
+			targetNodeName := getTargetSENodeName(f, pod)
```
irfanurrehman, line 272: Why do we need this here? Copy paste overload?

test/integration/action_execution.go
```diff
@@ -255,6 +255,48 @@ var _ = Describe("Action Executor ", func() {
...
+			if targetNodeName == "" {
+				framework.Failf("Failed to find a pod for deployment: %s", dep.Name)
+			}
+
+			// Test the provision action
+			_, err = actionHandler.ExecuteAction(newActionExecutionDTO(proto.ActionItemDTO_PROVISION,
+				newTargetSEFromPod(pod), nil), nil, &mockProgressTrack{})
+			framework.ExpectNoError(err, "Failed to execute provision action")
+
+			newDep, err := waitForDeployment(kubeClient, dep.Name, dep.Namespace)
```
irfanurrehman, line 282: `waitForDeployment()` is time wasting check slowing down the test. It will wait for the pod to become ready.
You can simply wait on check for the deployment ->spec.replicas being updated to desired value.

test/integration/action_execution.go
```diff
@@ -255,6 +255,48 @@ var _ = Describe("Action Executor ", func() {
...
+			// As the current replica is 2, the replica should be 3 after the provision action
+			if *newDep.Spec.Replicas != 3 {
+				framework.Failf("The replica number is incorrect after executing provision action")
+			}
+
+			// Test the suspend action
+			_, err = actionHandler.ExecuteAction(newActionExecutionDTO(proto.ActionItemDTO_SUSPEND,
+				newTargetSEFromPod(pod), nil), nil, &mockProgressTrack{})
+			framework.ExpectNoError(err, "Failed to execute suspend action")
+			newDep, err = waitForDeployment(kubeClient, newDep.Name, newDep.Namespace)
```
irfanurrehman, line 292: Same as above

test/integration/action_execution.go
kevinwangcn: will delete it

test/integration/action_execution.go
kevinwangcn: Sure, I'm going to add a new function `waitForDeploymentToUpdateReplica` to wait to check the deployment replica, which will be very similar to `waitForDeployment`, but as they're checking the different fields of the deployment, it still makes sense for me.

irfanurrehman (approved):

PR #688, opened 2022-05-31
Graceful handling of multiple gitops action executions

irfanurrehman:
~~This provides the option of failing the action if a PR for the same action already exists because of a previously failed action.~~

~~The change uses clusterID/namespace/name of the resource as a unique identifier on the PR Title and adds the code to check existing PRs by the same user on the same base branch (This is up for debate).~~

~~I am still working on the portion of the code, which can identify the difference in patch content and update the already existing PR with the new content if an update has happened.~~
~~Will update the PR description and the test details after I finish the same. Opening this to get early feedback.~~

**Intent**
This implements a graceful way of handling the kubeturbo restarts when a long running gitops based action is executed on argoCD pipeline.

**Background**
Prior to this implementation a kubeturbo restart will fail the action and leave the open PR behind. If the same action is executed again, a new PR will be created. If the PR is merged after the action was failed but is reapplied for execution (the action remains but the source of truth is now aligned to the desired state), a PR with empty commit will be created.

**Implementation**
The following algorithm is implemented

1. store `cluster-unique-id/workload-ns/workload-name` as a unique identified on the PR  title that is created as part of action execution.

2. if kubeturbo restarts, the action is failed in turbo server, which is regenerated again and can be executed again.

3. on the second (or any subsequent) action execution, kubeturbo checks all open PRs of the given user, and tries to find the PR which has the unique id for the given workload controller.

4. if not found standard flow is taken which includes

    - create a new branch on the repo

    - update the yaml content with the patch from action execution

    - create a PR and

    - wait for the PR to be merged

5. if found kubeturbo first checks if there is new diff compared to the existing action commit

    - if there is new diff, kubeturbo pushes a new commit on the PR branch

    - if there is no diff, kubeturbo does not do anything in terms of the PR commit

6. kubeturbo then continues to wait on the existing PR to be merged.

**Testing**

A new PR as a result of the action has the <cluster-unique-id><namespace>/<resource-name> as the unique identifier

image

Executing an action again after the first trial failed (because of kubeturbo restart):
image

The action then continues to wait until the PR is merged
image

### DIFF
--- a/pkg/action/executor/gitops/github_manager.go
+++ b/pkg/action/executor/gitops/github_manager.go
@@ -59,21 +59,23 @@ type gitWaitData struct {
 }
 
 type GitHubManager struct {
-	gitConfig   GitConfig
-	typedClient *typedClient.Clientset
-	dynClient   dynamic.Interface
-	obj         *unstructured.Unstructured
-	managerApp  *repository.K8sApp
+	gitConfig    GitConfig
+	typedClient  *typedClient.Clientset
+	dynClient    dynamic.Interface
+	obj          *unstructured.Unstructured
+	managerApp   *repository.K8sApp
+	k8sClusterId string
 }
 
 func NewGitHubManager(gitConfig GitConfig, typedClient *typedClient.Clientset, dynClient dynamic.Interface,
-	obj *unstructured.Unstructured, managerApp *repository.K8sApp) GitopsManager {
+	obj *unstructured.Unstructured, managerApp *repository.K8sApp, clusterId string) GitopsManager {
 	return &GitHubManager{
-		gitConfig:   gitConfig,
-		typedClient: typedClient,
-		dynClient:   dynClient,
-		obj:         obj,
-		managerApp:  managerApp,
+		gitConfig:    gitConfig,
+		typedClient:  typedClient,
+		dynClient:    dynClient,
+		obj:          obj,
+		managerApp:   managerApp,
+		k8sClusterId: clusterId,
 	}
 }
 
@@ -143,8 +145,8 @@ func (r *GitHubManager) Update(replicas int64, podSpec map[string]interface{}) (
 		},
 	}
 
-	glog.Infof("Updating the source of truth at: %s on branch: %s and path: %s.", url.Path, baseBranch, path)
-	return handler.updateRemote(r.obj.GetName(), patches)
+	glog.Infof("Updating the source of truth at: %s and path: %s.", url.Path, path)
+	return handler.updateRemote(r.obj.GetName(), r.obj.GetName(), r.k8sClusterId, patches)
 }
 
 func (r *GitHubManager) WaitForActionCompletion(completionData interface{}) error {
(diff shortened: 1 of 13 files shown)


kevinwangcn (approved):

pkg/action/executor/gitops/github_manager.go
```diff
@@ -464,6 +539,26 @@ func (g *GitHandler) newPR(resName string, newBranch string) (*github.PullReques
 	return pr, nil
 }
 
+func (g *GitHandler) checkPrExists(resName, resNamespace, clusterId string) (*github.PullRequest, error) {
+	options := &github.PullRequestListOptions{
+		Base: g.baseBranch,
+		Head: g.commitUser, // ex head=turbonomic or head=turbonomic: or head=turbonomic:<branch-name>
+	}
+	prList, _, err := g.client.PullRequests.List(g.ctx, g.user, g.repo, options)
```
enlinxu, line 547: nit: You may want to pull the open PRs only.
https://docs.github.com/en/rest/pulls/pulls#list-pull-requests

pkg/action/executor/gitops/github_manager.go
```diff
@@ -353,60 +356,131 @@ func (g *GitHandler) getTree(ref *github.Reference, fileContent []byte) (tree *g
...
+	var err error
 	if g.commitMode == commitModePR {
-		branchRef, err = g.createNewBranch(newBranchName)
+		pr, err = g.checkPrExists(resName, resNamespace, clusterId)
 		if err != nil {
-			return nil, fmt.Errorf("error creating new branch %s, %v", newBranchName, err)
+			return nil, err
+		}
+		if pr != nil {
+			glog.V(3).Infof("Found an existing PR #%d while executing action for "+
```
enlinxu, line 389: nit: Maybe use V(2) here, which is the default. I think this is a proper message at v2 level for better troubleshooting if needed.

pkg/action/executor/gitops/github_manager.go
```diff
@@ -353,60 +356,131 @@ func (g *GitHandler) getTree(ref *github.Reference, fileContent []byte) (tree *g
...
+	}
+	patchedYamlContent, err := ApplyPatch([]byte(yamlContent), patches)
+	if err != nil {
+		return false, nil, fmt.Errorf("error applying patches to file %s: patches: %v, err: %v", g.path, patches, err)
+	}
+
+	// TODO: find a better way to compare yamls. As patching can reorganise the
+	// yaml content, even if there is no change and the source yaml is not
+	// alphabetically arranged we will end up creating a commit.
+	needsUpdate := strings.Compare(yamlContent, string(patchedYamlContent)) != 0
```
enlinxu, line 480: Can we just check the resource values (like cpu limit/request and mem limit/request. I am afraid there may be other noise.

irfanurrehman: Yes, something like that is the plan as part of TODO. This is similar to what we want to address for patching too.

irfanurrehman:
@enlinxu thanks for the review. Updated, PTAL!

PR #691, opened 2022-06-06
Add ORM case in the integration test

kevinwangcn:
# Intent

Add ORM case in the integration test

# Background
The current integration test doesn't include the test for the ORM feature

# Implementation
Use the pre-existing OpenShift cluster and resize one of the containers in the deployment `auth-idp` which is managed by the `authentications.operator.ibm.com` CR `example-authentication` ,  and then validate the change on the CR, which should be updated according to the ORM CR. finally, revert the change on the CR no matter the test passes or not.
Do the similar test on the deployment `cert-manager-cainjector` to test the cluster scope case.

# Test Done

## Run both of the cases 
1. Run the following command to trigger the test
```
[root@localvm kubeturbo]# ./build/integration.test -k8s-kubeconfig=/root/.kube/config -k8s-context=ocp47 -ginkgo.focus="ORM test"
```

2. Get the result at the end of the test

image

3. Check the message in the logging output:
```
I0606 12:21:04.696170   99930 operator_resource_mapping.go:306] Successfully updated CR Authentication/example-authentication '.spec.auditService.resources' from map[limits:map[cpu:120m memory:40Mi] requests:map[cpu:10m memory:20Mi]] to map[limits:map[cpu:170m memory:115Mi] requests:map[cpu:10m memory:20Mi]] for Deployment/auth-idp in namespace ibm-common-services
...
I0606 12:21:15.515440   99930 operator_resource_mapping.go:306] Successfully updated CR CertManager/default '.spec.certManagerCAInjector.resources' from map[limits:map[cpu:800m memory:392Mi] requests:map[cpu:20m memory:282Mi]] to map[limits:map[cpu:775m memory:342Mi] requests:map[cpu:20m memory:282Mi]] for Deployment/cert-manager-cainjector in namespace 
```

4. Check the CR on the cluster to see if it's changed back after test.
```
k edit authentications.operator.ibm.com example-authentication
```
image
```
k edit certmanagers.operator.ibm.com default
```
image

### DIFF
--- a/test/integration/action_execution.go
+++ b/test/integration/action_execution.go
@@ -38,16 +38,8 @@ const (
 	cpuIncrement                 = 50
 	memoryIncrement              = 75
 	cpuDecrement                 = 25
-	memoryDecrement              = 50
+	memoryDecrement              = 10
 	injectedSidecarContainerName = "istio-proxy"
-)
-
-const (
-	REQUEST_SINGLE_CONTAINER = iota
-	LIMIT_SINGLE_CONTAINER
-	REQLIM_MULTI_CONTAINER
-	ORM_RESIZE_UP
-	ORM_RESIZE_DOWN
 )
 
 const (
@@ -238,7 +230,7 @@ var _ = Describe("Action Executor ", func() {
 			framework.ExpectNoError(err, "Error creating test resources")
 
 			targetSE := newResizeWorkloadControllerTargetSE(dc)
-			resizeAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, REQLIM_MULTI_CONTAINER, &dc.Spec.Template.Spec)
+			resizeAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, &dc.Spec.Template.Spec, "test-cont-1", "test-cont-2")
 			_, err = actionHandler.ExecuteAction(resizeAction, nil, &mockProgressTrack{})
 			framework.ExpectNoError(err, "Resize action on multiple container failed")
 
@@ -284,25 +276,14 @@ var _ = Describe("Action Executor ", func() {
 
 			targetSE := newResizeWorkloadControllerTargetSE(dep)
 
-			// Resize up cpu and memory on resource/limit
-			resizeAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, LIMIT_SINGLE_CONTAINER, &dep.Spec.Template.Spec)
+			// Resize up cpu and memory
+			resizeAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, &dep.Spec.Template.Spec, "test-cont")
 			_, err = actionHandler.ExecuteAction(resizeAction, nil, &mockProgressTrack{})
 			framework.ExpectNoError(err, "Resize action on limit failed")
 
-			controllerObj, err := waitForWorkloadControllerToUpdateResource(kubeClient, dep, desiredPodSpec)
-			if err != nil {
-				framework.Failf("Failed to check the change of resource/limit in the new deployment: %s", err)
-			}
-
-			// Resize up cpu and memory on resource/request
-			dep = controllerObj.(*appsv1.Deployment)
-			resizeAction, desiredPodSpec = newResizeActionExecutionDTO(targetSE, REQUEST_SINGLE_CONTAINER, &dep.Spec.Template.Spec)
-			_, err = actionHandler.ExecuteAction(resizeAction, nil, &mockProgressTrack{})
-			framework.ExpectNoError(err, "Resize action on request failed")
-
 			_, err = waitForWorkloadControllerToUpdateResource(kubeClient, dep, desiredPodSpec)
 			if err != nil {
-				framework.Failf("Failed to check the change of the resource/request in the new deployment: %s", err)
+				framework.Failf("Failed to check the change of resource/limit in the new deployment: %s", err)
 			}
 		})
 	})
@@ -327,7 +308,7 @@ var _ = Describe("Action Executor ", func() {
 			}
 
 			targetSE := newResizeWorkloadControllerTargetSE(dep)
-			resizeAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, REQLIM_MULTI_CONTAINER, &dep.Spec.Template.Spec)
+			resizeAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, &dep.Spec.Template.Spec, "test-cont-1", "test-cont-2")
 			_, err = actionHandler.ExecuteAction(resizeAction, nil, &mockProgressTrack{})
 			framework.ExpectNoError(err, "Resize action on multiple container failed")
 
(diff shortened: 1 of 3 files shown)


test/integration/orm_action_execution.go
```diff
@@ -0,0 +1,216 @@
...
+			// 3. Build and execute resize action on the deployment
+			targetSE := newResizeWorkloadControllerTargetSE(dep)
+			resizeupAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, ORM_RESIZE_UP, &dep.Spec.Template.Spec)
+			_, err = actionHandler.ExecuteAction(resizeupAction, nil, &mockProgressTrack{})
+			if err != nil {
+				framework.Failf("Failed to execute resizing action for ORM namespace case")
+			}
+
+			// 4. Set the revert flag for afterEach
+			needRevertCR4NsScope = true
```
irfanurrehman, line 129: I recommend to let the revert logic run irrespective of action being successful or not. You don't know at what stage the action failed, it might happen that the resource was updated but action failed for a secondary reason.
There is no harm in updating the resource with its original values, even if they are still the same.

test/integration/orm_action_execution.go
```diff
@@ -0,0 +1,216 @@
...
+			operatorCR, err := dynamicClient.Resource(groupVersionRes4nsscope).Namespace(ormtest_namespace).Get(context.TODO(), namespacescope_operator_cr_name, metav1.GetOptions{})
+			framework.ExpectNoError(err, "Error gettting the operator CR")
+			origCRValue4NsScope, _, err = util.NestedField(operatorCR, "destPath", namespacescope_operand_path_in_cr)
+
+			// 3. Build and execute resize action on the deployment
+			targetSE := newResizeWorkloadControllerTargetSE(dep)
+			resizeupAction, desiredPodSpec := newResizeActionExecutionDTO(targetSE, ORM_RESIZE_UP, &dep.Spec.Template.Spec)
+			_, err = actionHandler.ExecuteAction(resizeupAction, nil, &mockProgressTrack{})
+			if err != nil {
+				framework.Failf("Failed to execute resizing action for ORM namespace case")
```
irfanurrehman, line 125: Please check in the docs, as to what happens with the afterEach block in case of failures (and explicit call to failf) (and try it practically to ensure the behavior).
We want the aftereach to run even in the case of failures.

test/integration/orm_action_execution.go
```diff
@@ -0,0 +1,216 @@
...
+				aggregation.DefaultContainerUsageDataAggStrategy, ormClient, app.DefaultDiscoveryWorkers, app.DefaultDiscoveryTimeoutSec,
+				app.DefaultDiscoverySamples, app.DefaultDiscoverySampleIntervalSec)
+			actionHandlerConfig := action.NewActionHandlerConfig("", nil, nil,
+				cluster.NewClusterScraper(kubeClient, dynamicClient, false, nil, ""),
+				[]string{"*"}, ormClient, false, true, 60, gitops.GitConfig{})
+			actionHandler = action.NewActionHandler(actionHandlerConfig)
+
+			// Kubernetes Probe Discovery Client
+			discoveryClient := discovery.NewK8sDiscoveryClient(discoveryClientConfig)
+			_, _, err = discoveryClient.DiscoverWithNewFramework("discovery-integration-test")
```
irfanurrehman, line 103: Why do we need this here? Wouldn't this slow the whole test down a lot, especially when its against a real cluster

test/integration/orm_action_execution.go
```diff
@@ -0,0 +1,216 @@
...
+	})
+})
+
+func revertOperatorCR(client dynamic.Interface, groupVersionRes schema.GroupVersionResource, nsName, crName, destPath string, value interface{}) error {
+	operatorCR, err := client.Resource(groupVersionRes).Namespace(nsName).Get(context.TODO(), crName, metav1.GetOptions{})
+	framework.ExpectNoError(err, "Error gettting the operator CR")
+	fields := strings.Split(destPath, ".")
+	err = util.SetNestedField(operatorCR.Object, value, fields[1:]...)
+	dynResourceClient := client.Resource(groupVersionRes).Namespace(nsName)
+	_, err = dynResourceClient.Update(context.TODO(), operatorCR, metav1.UpdateOptions{})
```
irfanurrehman, line 187: Why break line 186 and 187 into two different lines?

test/integration/orm_action_execution.go
```diff
@@ -0,0 +1,216 @@
...
+		operatorCR, err := client.Resource(groupVersionRes).Namespace(nsName).Get(context.TODO(), crName, metav1.GetOptions{})
+		if err != nil {
+			glog.Errorf("Unexpected error while getting the operator CR: %v", err)
+			return false, nil
+		}
+		var cpuVal, memVal string
+		newCRValue, _, err := util.NestedField(operatorCR, "destPath", resPath)
+		if resVal, ok := newCRValue.(map[string]interface{}); ok {
+			if limVal, ok := resVal["limits"].(map[string]interface{}); ok {
+				cpuVal, _ = limVal["cpu"].(string)
```
irfanurrehman, line 202: `, _` can be dropped.

irfanurrehman (changes requested):

test/integration/orm_action_execution.go
kevinwangcn: sure,  will remove this flag

test/integration/orm_action_execution.go
kevinwangcn: That's just what I think. I tested it out by explicitly failing the spec and seeing the `afterEach` block is still triggered.

test/integration/orm_action_execution.go
kevinwangcn: I have to call `NewK8sDiscoveryClient` to trigger `CacheORMSpecMap` to initialize the map for saving orm object. Probably I could call it directly `dc.config.ormClient.CacheORMSpecMap()`, not sure if it works or not, will give it a shot.

test/integration/orm_action_execution.go
kevinwangcn: will combine them together

test/integration/orm_action_execution.go
kevinwangcn: will remove it.

test/integration/orm_action_execution.go
kevinwangcn: @irfanurrehman , I can't call this function like this `dc.config.ormClient.CacheORMSpecMap()` , `config` is not visiable outside the package. so I have to keep the old way.

test/integration/orm_action_execution.go
kevinwangcn: done

test/integration/orm_action_execution.go
kevinwangcn: done

test/integration/orm_action_execution.go
irfanurrehman: I am concerned about the time wasted in discovery on this cluster. I am not sure how much time does it take on this cluster. Can you please confirm. If its lesser then a minute, I am ok with it.

kevinwangcn: @irfanurrehman , I added a function named `TimeCost` to calculate the time consumption of `DiscoverWithNewFramework`. In the Travis log, we could see the following output, it is less than 20s, I also test this on my local Kind cluster, it's around 15s. 
```
Function name<DiscoverWithNewFramework>,Run time<7.955075221s>
Function name<DiscoverWithNewFramework>,Run time<15.17894403s>
```

irfanurrehman: Did you try exporting the variables and make `dc.config.ormClient.CacheORMSpecMap()` usable here?
My point is that we have to be cautious in adding time to overall test execution run. I think its already topping 25 mins now. Every 20 seconds add to it and there still are more tests to come.

test/integration/action_execution.go
```diff
@@ -988,6 +990,47 @@ func newResizeActionExecutionDTO(targetSE *proto.EntityDTO, changeType int, podS
 		updatePodSpec(desiredPodSpec, 0, "requests", "memory", RESIZE_DOWN, memoryDecrement)
 		updatePodSpec(desiredPodSpec, 1, "limits", "cpu", RESIZE_DOWN, cpuDecrement)
 		updatePodSpec(desiredPodSpec, 1, "limits", "memory", RESIZE_UP, memoryIncrement)
+	case ORM_RESIZE_UP:
+		currentSE := newContainerEntity(namespacescope_operand_container_name)
+		containerIdx, _ := findContainerIdxInPodSpecByName(podSpec, namespacescope_operand_container_name)
+		if containerIdx != -1 {
+			// Build the action item on cpu
```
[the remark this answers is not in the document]
kevinwangcn: Refactor them with adding a new parameter for the cotainer name, put resize down and up action into one action DTO and also covers the request and limit commodity.

test/integration/orm_action_execution.go
kevinwangcn: I tried it out, if we export the variables, we call `dc.config.ormClient.CacheORMSpecMap()` directly, there is a significant improvement, you can check that in Travis log.
```
Function name<DiscoverWithNewFramework>,Run time<816.624924ms>
Function name<DiscoverWithNewFramework>,Run time<775.385891ms>
```

test/integration/action_execution.go
```diff
@@ -988,6 +993,38 @@ func newResizeActionExecutionDTO(targetSE *proto.EntityDTO, changeType int, podS
 		updatePodSpec(desiredPodSpec, 0, "requests", "memory", RESIZE_DOWN, memoryDecrement)
 		updatePodSpec(desiredPodSpec, 1, "limits", "cpu", RESIZE_DOWN, cpuDecrement)
 		updatePodSpec(desiredPodSpec, 1, "limits", "memory", RESIZE_UP, memoryIncrement)
+	case ORM_RESIZE:
+		currentSE := newContainerEntity(containerName)
```
enlinxu, line 997: Maybe I am missing something, all the code below is the same for all Cases here, why are we repeating the code in each case? 
currentSE := newContainerEntity(containerName)
		containerIdx, err := findContainerIdxInPodSpecByName(podSpec, containerName)
		framework.ExpectNoError(err, "Can't find the container<%v> in the pod's spec", containerName)

		// Build the action item on cpu
		oldCpuCap := podSpec.Containers[0].Resources.Requests.Cpu().MilliValue()
		oldCpuCap := podSpec.Containers[containerIdx].Resources.Requests.Cpu().MilliValue()
		newCpuCap := oldCpuCap + cpuIncrement
		aiOnCpu := newActionItemDTO(actionType, proto.CommodityDTO_VCPU_REQUEST, oldCpuCap, newCpuCap, currentSE, targetSE)

		// Build the action item on memory
		oldMemCap := podSpec.Containers[0].Resources.Requests.Memory().Value() / 1024
		oldMemCap := podSpec.Containers[containerIdx].Resources.Requests.Memory().Value() / 1024
		newMemCap := oldMemCap + memoryIncrement*1024

enlinxu (commented):
Looks good overall, thank you Kevin.

test/integration/action_execution.go
kevinwangcn: Originally, I was planning to generate the different sizes of action with different combination for different test scenarios, some have 2 action items, and some may have 4 action items.  On a second thought, that might not make sense, as all of them still go to the same code branch. Now, I change it to a unified one.  For each container,  a resize action is composed of  the 4 action items, it covers resize up and resize down, also covers request and limits commodity.

enlinxu (approved):

irfanurrehman (approved):
Thanks @kevinwangcn

pankajbatra1 (approved):

PR #701, opened 2022-07-13
Add additional validation and unit tests for turbo policy conversion

ading1977:
This PR:
* Add additional validation to SLO setting values to make sure they are greater than zero
* Add unit tests

### DIFF
--- a/pkg/discovery/worker/group_discovery_worker.go
+++ b/pkg/discovery/worker/group_discovery_worker.go
@@ -109,7 +109,7 @@ func (worker *k8sEntityGroupDiscoveryWorker) Do(entityGroupList []*repository.En
 	groupDTOs = append(groupDTOs, worker.buildSidecarContainerSpecGroup(sidecarContainerSpecs)...)
 
 	// Create dynamic groups for discovered Turbo policies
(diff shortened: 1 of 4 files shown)


pkg/discovery/worker/group_discovery_worker.go
```diff
@@ -257,9 +257,6 @@ func (worker *k8sEntityGroupDiscoveryWorker) createSLOPolicy(
 	}
 	var settingValue float32
 	for _, setting := range spec.Objectives {
-		if err := json.Unmarshal(setting.Value.Raw, &settingValue); err != nil {
-			return nil, err
-		}
 		switch setting.Name {
 		case v1alpha1.ResponseTime:
 			settings.AddSetting(group.NewResponseTimeSLOPolicySetting(settingValue))
```
chlam4, line 262: Isn't `settingValue` not yet initialized here?

irfanurrehman: I also dont understand this change, why add a zero setting value?

ading1977: Oh, I see where the confusion is, the following logic is moved to line 268:

```
		if err := json.Unmarshal(setting.Value.Raw, &settingValue); err != nil {
			return nil, err
		}
```

PR #705, opened 2022-07-25
Fix range loop variable issue when iterating through turbo policies and bindings

ading1977:
## Problem
This PR fixes a bug where we forgot to redefine range loop variables when iterating turbo policies and policybindings. This leads to the problem where only one policy and policybinding is cached no matter how many policies and bindings are available in the cluster.

## Fix
Redefine range loop variables

## Test
- Dev build has been verified by QA
- Add unit tests

### DIFF
--- a/pkg/discovery/processor/turbo_policy_processor.go
+++ b/pkg/discovery/processor/turbo_policy_processor.go
@@ -47,36 +47,38 @@ func (p *TurboPolicyProcessor) ProcessTurboPolicies() {
 
 	policyMap := make(map[string]*repository.TurboPolicy)
 	for _, sloScale := range turboSloScalings {
-		turboSloScaling := sloScale
-		gvk := turboSloScaling.GetObjectKind().GroupVersionKind()
+		// Create a copy as sloScale variable is reused during range loop
+		sloScaleCopy := sloScale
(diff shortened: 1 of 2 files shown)


irfanurrehman (approved):

pkg/discovery/processor/turbo_policy_processor.go
```diff
@@ -46,7 +46,8 @@ func (p *TurboPolicyProcessor) ProcessTurboPolicies() {
 	glog.V(2).Infof("Discovered %v PolicyBindings.", len(turboPolicyBindings))
 
 	policyMap := make(map[string]*repository.TurboPolicy)
-	for _, turboSloScaling := range turboSloScalings {
+	for _, sloScale := range turboSloScalings {
+		turboSloScaling := sloScale
```
chlam4, line 50: Can you elaborate why this makes a difference?  Is that because `:=` actually creates a copy?  If so, then would it make sense to perhaps name it `turboSloScalingCopy` to make it obvious?

ading1977: This is a great suggestion. Thanks @chlam4 .

chlam4 (approved):

PR #706, opened 2022-07-29
Refactor the monitoring worker to better handle concurrency

enlinxu:
Intent: 
Thanks to Cheuk's analysis on what happened at BMO, the worker go routine will be stuck in certain scenario when certain nodes are not accessible.

Fix:
Here're the changes in this PR:
1. Refactor Task to take node as field instead of array of nodes. This is the case after we refactor the discovery framework. The node array always only contains one node before the change.
2. Refactor the downstream struct like kubelet_monitor and cluster_monitor and dummy_monitor.
3. Modified all the downstream files after change in 2)
4. Change the default discovery worker count to 10 from 4.
5. Remove the boundary of max discovery worker

Test:
Unit tests passed and sanity check passed. Will check PS later.

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -57,7 +57,7 @@ const (
 	defaultDiscoveryIntervalSec       = 600
 	DefaultValidationWorkers          = 10
 	DefaultValidationTimeout          = 60
-	DefaultDiscoveryWorkers           = 4
+	DefaultDiscoveryWorkers           = 10
 	DefaultDiscoveryTimeoutSec        = 180
 	DefaultDiscoverySamples           = 10
 	DefaultDiscoverySampleIntervalSec = 60
--- a/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
+++ b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
@@ -118,7 +118,7 @@ func (m *KubeletMonitor) scrapeKubelet(node *api.Node) error {
 		} else {
 			// It's a valid case if a node is available from the full discovery but not available during sampling discoveries.
 			// Need to wait for a full discovery to fetch the available nodes.
-			glog.Warningf("Failed to get resource metrics summary sample from %s. Waiting for the next full discovery.", node.Name)
+			return fmt.Errorf("failed to get resource metrics summary sample from %s", node.Name)
 		}
 	}
 
--- a/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
+++ b/pkg/discovery/monitoring/kubelet/kubelet_monitor.go
@@ -62,14 +62,15 @@ func (m *KubeletMonitor) ReceiveTask(task *task.Task) {
 	m.node = task.Node()
 }
 
-func (m *KubeletMonitor) Do() *metrics.EntityMetricSink {
+func (m *KubeletMonitor) Do() (*metrics.EntityMetricSink, error) {
 	glog.V(4).Infof("%s has started task.", m.GetMonitoringSource())
 	err := m.RetrieveResourceStat()
 	if err != nil {
 		glog.Errorf("Failed to execute task: %s", err)
+		return m.metricSink, err
 	}
 	glog.V(4).Infof("%s monitor has finished task.", m.GetMonitoringSource())
-	return m.metricSink
+	return m.metricSink, nil
 }
 
 // RetrieveResourceStat retrieves resource stats for the received node.
@@ -77,12 +78,15 @@ func (m *KubeletMonitor) RetrieveResourceStat() error {
 	if m.node == nil {
 		return errors.New("empty node")
 	}
-	m.scrapeKubelet(m.node)
+	err := m.scrapeKubelet(m.node)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
 // Retrieve resource metrics for the given node.
-func (m *KubeletMonitor) scrapeKubelet(node *api.Node) {
+func (m *KubeletMonitor) scrapeKubelet(node *api.Node) error {
 	kc := m.kubeletClient
 
 	// Collect node cpu frequency metric only in full discovery not in sampling discovery
@@ -90,7 +94,7 @@ func (m *KubeletMonitor) scrapeKubelet(node *api.Node) {
 		nodefreq, err := kc.GetNodeCpuFrequency(node)
 		if err != nil {
 			glog.Errorf("Failed to get resource metrics (cpufreq) from %s: %s", node.Name, err)
-			return
+			return err
 		}
 		m.parseNodeCpuFreq(node, nodefreq)
 	}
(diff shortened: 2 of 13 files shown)


pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -209,7 +206,7 @@ func (worker *k8sDiscoveryWorker) executeTask(currTask *task.Task) *task.TaskRes
 					glog.V(3).Infof("A %s monitoring worker from discovery worker %v is invoked for task %s.",
 						w.GetMonitoringSource(), worker.id, resourceMonitorTask)
 					// Assign task to monitoring worker.
-					monitoringSink := w.Do(stopCh)
+					monitoringSink := w.Do()
```
ading1977, line 209: Shall we also return possible `error` for the `Do()` function? In the case of error, we don't need to merge sink later on.

chlam4 (approved):
lgtm.  thank you @enlinxu!

pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -197,7 +195,6 @@ func (worker *k8sDiscoveryWorker) executeTask(currTask *task.Task) *task.TaskRes
 			wg.Add(1)
 			go func(w monitoring.MonitoringWorker) {
 				finishCh := make(chan struct{})
-				stopCh := make(chan struct{})
 				timeoutCh := make(chan struct{}, 1)
 				defer close(finishCh)
```
ading1977, line 199: Suggest remove this line, and replace  `finishCh <- struct{}{}` on line 228 with `close(finishCh)`. It is not a best practice to close a channel on the receiving end, as explained [here](https://go101.org/article/channel-closing.html).

enlinxu: Good catch Meng. Updated. PTAL.

pkg/discovery/worker/k8s_discovery_worker.go
enlinxu: Added. PTAL

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -129,12 +119,13 @@ func (m *KubeletMonitor) scrapeKubelet(node *api.Node) {
 			// It's a valid case if a node is available from the full discovery but not available during sampling discoveries.
 			// Need to wait for a full discovery to fetch the available nodes.
 			glog.Warningf("Failed to get resource metrics summary sample from %s. Waiting for the next full discovery.", node.Name)
-			return
```
ading1977, line 132: Shall we keep the old logic and return here instead of continuing?

enlinxu: @marvelyue , what do you think? :)

marvelyue: Based on the code comment, to be safer, I'm leaning to the old logic by returning here instead of continuing. If node is not available during sampling discovery, then continuing on following logic to parse pod and node stats could lead to unexpected results.

Will take a closer look and revisit the whole logic later today.

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -57,7 +57,7 @@ const (
 	defaultDiscoveryIntervalSec       = 600
 	DefaultValidationWorkers          = 10
 	DefaultValidationTimeout          = 60
-	DefaultDiscoveryWorkers           = 4
+	DefaultDiscoveryWorkers           = 10
```
pankajbatra1, line 60: How will we know that these changes work if the two culprit nodes are back up in BMO?

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
```diff
@@ -59,66 +59,56 @@ func (m *KubeletMonitor) GetMonitoringSource() types.MonitoringSource {
...
+	return m.metricSink, nil
 }
 
-// RetrieveResourceStat retrieves resource stats for the received list of nodes.
-func (m *KubeletMonitor) RetrieveResourceStat(stopChan <-chan struct{}) error {
-	if m.nodeList == nil || len(m.nodeList) == 0 {
-		return errors.New("empty node list")
+// RetrieveResourceStat retrieves resource stats for the received node.
+func (m *KubeletMonitor) RetrieveResourceStat() error {
+	if m.node == nil {
```
pankajbatra1, line 78: nit: this check could be caught earlier in Do() or even before

pkg/discovery/monitoring/master/cluster_monitor.go
```diff
@@ -107,22 +96,20 @@ func (m *ClusterMonitor) findClusterID() error {
...
-		if key == "" {
-			glog.Warning("Invalid node")
-			continue
-		}
-		// node/pod/container cpu/mem resource capacities
-		m.genNodeResourceMetrics(node, key)
+	key := util.NodeKeyFunc(m.node)
+	if key == "" {
+		glog.Warning("Invalid node")
+		return
```
pankajbatra1, line 102: Should you return an error to caller?

cmd/kubeturbo/app/kubeturbo_builder.go
enlinxu: For the discovery worker change, we can tell from the log immediately: 
I0729 17:54:07.519167    6430 k8s_discovery_client.go:63] Number of discovery workers: 10.

For the concurrency handling, the fact is we are removing the unnecessary go routine (which caused the deadlock), so for sure we won't see the deadlock anymore.

pkg/discovery/worker/k8s_discovery_worker.go
```diff
@@ -197,9 +195,7 @@ func (worker *k8sDiscoveryWorker) executeTask(currTask *task.Task) *task.TaskRes
 			wg.Add(1)
 			go func(w monitoring.MonitoringWorker) {
 				finishCh := make(chan struct{})
-				stopCh := make(chan struct{})
 				timeoutCh := make(chan struct{}, 1)
-				defer close(finishCh)
```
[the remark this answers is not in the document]
enlinxu: It's not removed, it's moved to the Sender side.
@ading1977 had a great point about the best practice to close the channel. Channel close should always happen in the sender sider. The defer statement here is in the Receiver go routine.

pkg/discovery/monitoring/master/cluster_monitor.go
enlinxu: Updated

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
enlinxu: Updated based on the comments here. Thanks @marvelyue , I know you would notice! :)

pkg/discovery/monitoring/kubelet/kubelet_monitor.go
enlinxu: Updated, PTAL.

pankajbatra1 (approved):

PR #710, opened 2022-08-05
Gitops diff improvements

irfanurrehman:
**Intent**
Implements updates in the diff and patch mechanism to handle issue as listed in but not limited to https://vmturbo.atlassian.net/browse/OM-87593

**Background**
As part of first implementation we used the json patch library `github.com/evanphx/json-patch` to apply json patch style patches on to the source of truth yaml. We used simple string diff to compare the original yaml and the patched yaml to find if there is something updated by the patch. This could bring in multiple corner cases where although there is no real update in the live spec wrt to the source of truth (config) spec, we would still get a diff and end up creating an additional commit on an already open PR. These cases are:
- When the config (source of truth) yaml units are specified in milicores which get converted to cores in the live resource
- When we have additional defaulted fields in the live resource which are not present in the config (source of truth) yaml.

This PR handles that.

**Implementation**
We now use the diff lib from `argoproj` ie. `github.com/argoproj/gitops-engine/pkg/diff`, which uses the same libs provided by k8s that implement `kubectl diff`. These libs use server side diff logic and based on the known resource schema, can ignore defaulted fields and the unit differences.
We still use the json-patch package to apply the final patch to update the source of truth yaml, although the above diff provides a version of patched resource itself. This is because the patched/updated resource provided by the above diff includes all the additional defaulted fields which is actually meant for and is suitable for the resource update downstream (into the k8s API server) but not back upstream.

A point to note is that this brings in some of the indirect upstream kubernetes/kubernetes dependency back in, which explains the manual change including `replace directives` for all the relevant kubernetes libs in `go.mod`. This does increase the kubeturbo binary size a bit (66Mb from 57Mb) but is however not at all as bad as what we had earlier when we were directly using k/k dependency (kubeturbo bin size = 120Mb)

**Testing Done**
Added unit tests to test the diff and patch logic

Manual action execution in PR mode works as expected
Executing the action
image

creates a PR
image

merging the PR
image

marks the action succeeded
image

The patch does not add any new (defaulted) fields (an unordered yaml is still arranged into alphabetical order of keys)
image


pkg/action/executor/gitops/github_manager.go
```diff
@@ -464,21 +450,75 @@ func (g *GitHandler) updateRemote(resName, resNamespace, clusterId string, patch
 	return nil, nil
 }
 
-func (g *GitHandler) patchYamlContent(resName, branchName string, patches []PatchItem) (bool, []byte, error) {
-	yamlContent, err := g.getRemoteFileContent(resName, g.path, branchName)
+// We find the diff between the resource in the source of truth and the liveRes using
```
pankajbatra1, line 453: Is liveRes a typo?

irfanurrehman: Not really, I meant `live Resource`. I will expand for it to be more correct.

pankajbatra1 (approved):

PR #713, opened 2022-08-11
Create group of all pods using volumes per cluster

irfanurrehman:
**Intent**
Create a group per cluster for all those pods that use volumes
Implements https://vmturbo.atlassian.net/browse/OM-88242

**Implementation**
Create a string set of pods ids identifying those which use volumes and then create a group from that set

**Testing Done**
After the updated code in kubeturbo, a group is created
image

image

image


pkg/discovery/worker/group_discovery_worker.go
```diff
@@ -155,6 +157,32 @@ func (worker *k8sEntityGroupDiscoveryWorker) buildSidecarContainerSpecGroup(side
 	return groupsDTOs
 }
 
+func (worker *k8sEntityGroupDiscoveryWorker) buildPodsWithVolumesGroup(podsWithVolumes []string) []*proto.GroupDTO {
+	var groupsDTOs []*proto.GroupDTO
+	if len(podsWithVolumes) <= 0 {
+		return groupsDTOs
+	}
+	id := fmt.Sprintf("All-Pods-Using-Volumes-%s", worker.targetId)
+	displayName := "All Pods Using Volumes"
```
amd-ibm, line 166: Should the display name be dynamically generated like the ID is here? It seems like this could result in a scenario where there are multiple groups generated, all with the same display name, but with differing contents.

irfanurrehman: @amd-ibm there's a special server side magic( :) ) that differentiates the gourps with same name but coming from different clusters.
It identifies the groups uniquely by `id` field and attaches the cluster identifier with each discovered group and can differentiate the same in the UI, eg. you would see the cluster name displayed below the group:
image
We have many such auto discovered groups with static common names across different clusters. For example:
image

amd-ibm: I was unaware of that. Thanks for the info.

amd-ibm (approved):

PR #718, opened 2022-08-29
Honor the region/zone lable of the node when moving a pod that has PV attached

kevinwangcn:
# Intent
To honor the region/zone label(`topology.kubernetes.io/zone` and `topology.kubernetes.io/region`) when moving a pod that has any PV attached

# Background

# Implementation
The labels are presented as the `label` commodity in the kubeturo. When kubeturbo build the node DTOs, it's actually creating the `label` commodity for every label on that node, it also includes these two region/zone related labels. Based on this, the implementation is to have kubeturbo create new `label` commodity when it found the pod has PV attached and also the node the pods is currently running on has a zone/region label.

# Test Done
## Prepare the environment
1. Create a Fyre cluster with 3 master nodes and 3 worker nodes
```
[root@IBM-PF3888SM deploykubeturbo]# k get nodes
NAME                             STATUS   ROLES    AGE     VERSION
master0.brutes.cp.fyre.ibm.com   Ready    master   5h38m   v1.21.11+6b3cbdd
master1.brutes.cp.fyre.ibm.com   Ready    master   5h38m   v1.21.11+6b3cbdd
master2.brutes.cp.fyre.ibm.com   Ready    master   5h38m   v1.21.11+6b3cbdd
worker0.brutes.cp.fyre.ibm.com   Ready    worker   5h30m   v1.21.11+6b3cbdd
worker1.brutes.cp.fyre.ibm.com   Ready    worker   5h30m   v1.21.11+6b3cbdd
worker2.brutes.cp.fyre.ibm.com   Ready    worker   5h29m   v1.21.11+6b3cbdd
```
2. Label `worker0` with `xxx=yyy` and `topology.kubernetes.io/zone=us-east-1`, label `worker1` with `topology.kubernetes.io/zone=us-east-1` and label `worker2` with `topology.kubernetes.io/zone=us-west-1`. Basically we're trying to put `worker0` and `worker1` in the same zone and put `worker2` in a different one. 
```
[root@IBM-PF3888SM deploykubeturbo]# k get nodes --show-labels worker0.brutes.cp.fyre.ibm.com  worker1.brutes.cp.fyre.ibm.com  worker2.brutes.cp.fyre.ibm.com 
NAME                             STATUS   ROLES    AGE     VERSION            LABELS
worker0.brutes.cp.fyre.ibm.com   Ready    worker   5h33m   v1.21.11+6b3cbdd   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,disktype=ssd,kubernetes.io/arch=amd64,kubernetes.io/hostname=worker0.brutes.cp.fyre.ibm.com,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.openshift.io/os_id=rhcos,topology.kubernetes.io/zone=us-east-1,xxx=yy
worker1.brutes.cp.fyre.ibm.com   Ready    worker   5h33m   v1.21.11+6b3cbdd   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,disktype=ssd,kubernetes.io/arch=amd64,kubernetes.io/hostname=worker1.brutes.cp.fyre.ibm.com,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.openshift.io/os_id=rhcos,topology.kubernetes.io/zone=us-east-1
worker2.brutes.cp.fyre.ibm.com   Ready    worker   5h32m   v1.21.11+6b3cbdd   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,disktype=ssd,kubernetes.io/arch=amd64,kubernetes.io/hostname=worker2.brutes.cp.fyre.ibm.com,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.openshift.io/os_id=rhcos,topology.kubernetes.io/zone=us-west-1

```
3. Setup [LovalVolume](https://docs.openshift.com/container-platform/3.11/install_config/persistent_storage/persistent_storage_local.html)` by creating a StorageClass, a PV and PVC
```
[root@IBM-PF3888SM zonetest]# k get storageclasses.storage.k8s.io local-storage -o yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2022-08-25T14:52:08Z"
  name: local-storage
  resourceVersion: "49890"
  uid: 05cf668c-87ce-4c58-b08e-c564c4226693
provisioner: kubernetes.io/no-provisioner
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
```
```
[root@IBM-PF3888SM zonetest]# k get pv example-pv -o yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/bound-by-controller: "yes"
  creationTimestamp: "2022-08-25T19:23:17Z"
  finalizers:
  - kubernetes.io/pv-protection
  name: example-pv
  resourceVersion: "149405"
  uid: f95858fc-764e-4417-bbd1-e76f82eb2b66
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 1Gi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: test-move-pvc
    namespace: default
    resourceVersion: "148433"
    uid: e3158aa6-79d3-49e8-ad1a-c40938a60bd6
  local:
    path: /opt
  persistentVolumeReclaimPolicy: Delete
  storageClassName: local-storage
  volumeMode: Filesystem
status:
  phase: Bound
```
```
[root@IBM-PF3888SM zonetest]# k get pvc test-move-pvc  -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/no-provisioner
  creationTimestamp: "2022-08-25T19:23:21Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: test-move-pvc
  namespace: default
  resourceVersion: "149408"
  uid: e3158aa6-79d3-49e8-ad1a-c40938a60bd6
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Mi
  storageClassName: local-storage
  volumeMode: Filesystem
  volumeName: example-pv
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 1Gi
  phase: Bound
````
4. Create a deployment mounting the PVC and set `nodeSelector` to `xxx=yyy`
```
[root@IBM-PF3888SM deploykubeturbo]# k get deployment nginx-deployment  -o yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: "2022-08-29T18:56:00Z"
  generation: 1
  labels:
    app: nginx
  name: nginx-deployment
  namespace: default
  resourceVersion: "145446"
  uid: f130f5d9-0b4c-454f-8470-ec539d280878
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: nginx
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      annotations:
        openshift.io/scc: restricted
      creationTimestamp: null
      labels:
        app: nginx
    spec:
      containers:
      - env:
        - name: RUN_TYPE
          value: cpu
        - name: CPU_PERCENT
          value: "100"
        image: beekman9527/cpumemload:latest
        imagePullPolicy: Always
        name: nginx
        resources:
          limits:
            cpu: 1510m
          requests:
            cpu: 10m
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /mnt
          name: pxvol
      dnsPolicy: ClusterFirst
      imagePullSecrets:
      - name: turboregistrykey
      - name: turbocred
      nodeSelector:
        xxx: yyy   <---------- add this selector for the convenience of triggering a move action
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
      volumes:
      - name: pxvol
        persistentVolumeClaim:
          claimName: test-move-pvc
```
5. Make sure the pod is running on `worker0` and its PVC is bound
```
[root@IBM-PF3888SM deploykubeturbo]# k get pods -o wide
NAME                                READY   STATUS    RESTARTS   AGE   IP             NODE                             NOMINATED NODE   READINESS GATES
kubeturbo-kw-114-77f68c9978-l2wtx   1/1     Running   0          81m   10.254.18.88   worker0.brutes.cp.fyre.ibm.com   <none>           <none>
nginx-deployment-66fcf88df9-dmqfg   1/1     Running   0          59m   10.254.18.97   worker0.brutes.cp.fyre.ibm.com   <none>           <none>

```
```
[root@IBM-PF3888SM zonetest]# k get pvc
NAME            STATUS   VOLUME       CAPACITY   ACCESS MODES   STORAGECLASS    AGE
test-move-pvc   Bound    example-pv   1Gi        RWO            local-storage   22m```
```

## Case 1
**1. Label `worker2` with `xxx=yyy` and remove this label from `worker0`**
```
[root@IBM-PF3888SM deploykubeturbo]# k get nodes --show-labels worker0.brutes.cp.fyre.ibm.com  worker1.brutes.cp.fyre.ibm.com  worker2.brutes.cp.fyre.ibm.com 
NAME                             STATUS   ROLES    AGE     VERSION            LABELS
worker0.brutes.cp.fyre.ibm.com   Ready    worker   5h48m   v1.21.11+6b3cbdd   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,disktype=ssd,kubernetes.io/arch=amd64,kubernetes.io/hostname=worker0.brutes.cp.fyre.ibm.com,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.openshift.io/os_id=rhcos,topology.kubernetes.io/zone=us-east-1
worker1.brutes.cp.fyre.ibm.com   Ready    worker   5h48m   v1.21.11+6b3cbdd   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,disktype=ssd,kubernetes.io/arch=amd64,kubernetes.io/hostname=worker1.brutes.cp.fyre.ibm.com,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.openshift.io/os_id=rhcos,topology.kubernetes.io/zone=us-east-1
worker2.brutes.cp.fyre.ibm.com   Ready    worker   5h47m   v1.21.11+6b3cbdd   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,disktype=ssd,kubernetes.io/arch=amd64,kubernetes.io/hostname=worker2.brutes.cp.fyre.ibm.com,kubernetes.io/os=linux,node-role.kubernetes.io/worker=,node.openshift.io/os_id=rhcos,topology.kubernetes.io/zone=us-west-1,xxx=yyy

```

**2. Because `xxx=yyy` has been moved to `worker2` from `worker0`, before this code change, it should generate a move action for this pod, but in fact,  there is no move action generated,  the reason is the pod has a PV attached and is running on `worker0` which has zone label pointing to `us-east-1`, but `worker2` doesn't have that label, its zone label is pointing to `us-west-1`. As we're honoring this label, there is no move action for this, instead we get a reconfigure action**
image

**3. Remove the label `xxx=yyy` from `worker2` and label it to `worker1`, we can see a move action is generated, as  `worker1` is in the same zone as the `worker0`**
image

### DIFF
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder.go
@@ -23,6 +23,8 @@ import (
 
 const (
 	applicationCommodityDefaultCapacity = 1e10
+	zoneLabelName                       = "topology.kubernetes.io/zone"
+	regionLabelName                     = "topology.kubernetes.io/region"
 )
 
 var (
@@ -63,21 +65,23 @@ var (
 
 type podEntityDTOBuilder struct {
 	generalBuilder
-	stitchingManager *stitching.StitchingManager
-	nodeNameUIDMap   map[string]string
-	namespaceUIDMap  map[string]string
-	podToVolumesMap  map[string][]repository.MountedVolume
-	runningPods      []*api.Pod
-	pendingPods      []*api.Pod
+	stitchingManager  *stitching.StitchingManager
+	nodeNameUIDMap    map[string]string
+	namespaceUIDMap   map[string]string
+	podToVolumesMap   map[string][]repository.MountedVolume
+	nodeNameToNodeMap map[string]*repository.KubeNode
+	runningPods       []*api.Pod
+	pendingPods       []*api.Pod
 }
 
 func NewPodEntityDTOBuilder(sink *metrics.EntityMetricSink, stitchingManager *stitching.StitchingManager) *podEntityDTOBuilder {
 	return &podEntityDTOBuilder{
-		generalBuilder:   newGeneralBuilder(sink),
-		stitchingManager: stitchingManager,
-		nodeNameUIDMap:   make(map[string]string),
-		namespaceUIDMap:  make(map[string]string),
-		podToVolumesMap:  make(map[string][]repository.MountedVolume),
+		generalBuilder:    newGeneralBuilder(sink),
+		stitchingManager:  stitchingManager,
+		nodeNameUIDMap:    make(map[string]string),
+		namespaceUIDMap:   make(map[string]string),
+		podToVolumesMap:   make(map[string][]repository.MountedVolume),
+		nodeNameToNodeMap: make(map[string]*repository.KubeNode),
 	}
 }
 
@@ -93,6 +97,11 @@ func (builder *podEntityDTOBuilder) WithNameSpaceUIDMap(namespaceUIDMap map[stri
 
 func (builder *podEntityDTOBuilder) WithPodToVolumesMap(podToVolumesMap map[string][]repository.MountedVolume) *podEntityDTOBuilder {
 	builder.podToVolumesMap = podToVolumesMap
+	return builder
+}
+
+func (builder *podEntityDTOBuilder) WithNodeNameToNodeMap(nodeNameNodeMap map[string]*repository.KubeNode) *podEntityDTOBuilder {
+	builder.nodeNameToNodeMap = nodeNameNodeMap
 	return builder
 }
 
@@ -393,6 +402,17 @@ func (builder *podEntityDTOBuilder) getPodCommoditiesBought(
 		commoditiesBought = append(commoditiesBought, labelComm)
 	}
 
+	// Label commodities
+	// To honor the region/zone label on th node that the pod is running on if the pod has any PV attached
+	if utilfeature.DefaultFeatureGate.Enabled(features.HonorRegionZoneLabels) {
+		var anerr error
+		commoditiesBought, anerr = builder.getRegionZoneLabelCommodity(pod, commoditiesBought)
+		if anerr != nil {
+			glog.Errorf("Failed to append the region/zone label commodity")
+			return nil, anerr
+		}
+	}
+
 	// Cluster commodity.
 	clusterMetricUID := metrics.GenerateEntityStateMetricUID(metrics.ClusterType, "", metrics.Cluster)
 	clusterInfo, err := builder.metricsSink.GetMetric(clusterMetricUID)
@@ -532,4 +552,45 @@ func (builder *podEntityDTOBuilder) createContainerPodData(pod *api.Pod) *proto.
 		glog.Errorf("No IP found for pod %s", fullName)
 	}
 	return podData
+}
+
+func (builder *podEntityDTOBuilder) getRegionZoneLabelCommodity(pod *api.Pod, commoditiesBought []*proto.CommodityDTO) ([]*proto.CommodityDTO, error) {
+	if pod == nil {
+		return nil, fmt.Errorf("The pod's pointer is nil")
+	}
+	var err error
+	displayName := util.GetPodClusterID(pod)
+	mounts := builder.podToVolumesMap[displayName]
+	if len(mounts) > 0 {
+		if node, ok := builder.nodeNameToNodeMap[pod.Spec.NodeName]; ok {
+			if value, isFound := node.Labels[regionLabelName]; isFound {
+				selector := regionLabelName + "=" + value
+				commoditiesBought, err = AppendNewLabelCommodityToList(commoditiesBought, selector)
+				if err != nil {
+					return nil, err
+				}
+				glog.V(4).Infof("Added label commodity for Pod %s with key : %s", pod.Name, selector)
+			}
+			if value, isFound := node.Labels[zoneLabelName]; isFound {
+				selector := zoneLabelName + "=" + value
+				commoditiesBought, err = AppendNewLabelCommodityToList(commoditiesBought, selector)
+				if err != nil {
+					return nil, err
+				}
+				glog.V(4).Infof("Added label commodity for Pod %s with key : %s", pod.Name, selector)
+			}
+		}
+	}
+	return commoditiesBought, nil
+}
+
+func AppendNewLabelCommodityToList(commoditiesList []*proto.CommodityDTO, key string) ([]*proto.CommodityDTO, error) {
+	labelComm, err := sdkbuilder.NewCommodityDTOBuilder(proto.CommodityDTO_LABEL).
+		Key(key).
+		Create()
+	if err != nil {
+		return nil, err
+	}
+	commoditiesList = append(commoditiesList, labelComm)
+	return commoditiesList, nil
 }
--- a/pkg/discovery/dtofactory/pod_entity_dto_builder_test.go
+++ b/pkg/discovery/dtofactory/pod_entity_dto_builder_test.go
@@ -2,14 +2,17 @@ package dtofactory
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
 	api "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/turbonomic/kubeturbo/pkg/discovery/metrics"
+	"github.com/turbonomic/kubeturbo/pkg/discovery/repository"
 	"github.com/turbonomic/kubeturbo/pkg/discovery/util"
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
 )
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by ading1977** [comment]
Thanks for working on this @kevinwangcn , so effectively, we are implementing this by adding a label commodity for zone and region, right? Some questions:

* Does this work only for AWS? Is this requirement to support AWS only? Do we have a similar situation for other cloud providers? 
* Can we always trust that the `topology.kubernetes.io/zone` labels are set by AWS? Do other cloud providers set this label, or some other similar labels?
* The reconfigure action is confusing to users. It says **Reconfigure nodeSelector** while in fact this won't be the right fix. How does the user know that the reconfigure is caused by worker2 not being in the same zone?

**Review by chlam4 — COMMENTED** [review]
Are we planning to add some unit tests for the corresponding change?  Otherwise, the change looks good to me - thanks @KevinWang0204!

**Comment by kevinwangcn** [comment]
> Thanks for working on this @kevinwangcn , so effectively, we are implementing this by adding a label commodity for zone and region, right? Some questions:
> 
>     * Does this work only for AWS? Is this requirement to support AWS only? Do we have a similar situation for other cloud providers?
> 
>     * Can we always trust that the `topology.kubernetes.io/zone` labels are set by AWS? Do other cloud providers set this label, or some other similar labels?
> 
>     * The reconfigure action is confusing to users. It says **Reconfigure nodeSelector** while in fact this won't be the right fix. How does the user know that the reconfigure is caused by worker2 not being in the same zone?

Thanks for review this @ading1977 , I don't think it only works for AWS, I also checked AKS and GKE cluster,  they all have the label `topology.kubernetes.io/zone` and `topology.kubernetes.io/region` on their node. Also, the K8s document give some explanation about this label: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone, it seems to me this label is also involved by K8s default scheduler.  So I think we can trust these labels. 

In terms of `reconfigure` action, you're right, the current description of the action isn't right, the `PV nodeAffinity` feature also have this problem, because we're introducing new constraints. I remember Nirali ever fix a [bug](https://vmturbo.atlassian.net/browse/OM-81611) to make the description more accurate for the `reconfigure` action, I'm wondering if we can do similar thing in `action-orchestrator` to generate a better description.  I need spend some time investigate it, but I would appreciate any suggestion from you.

**Comment by ading1977** [comment]
> 
> Thanks for review this @ading1977 , I don't think it only works for AWS, I also checked AKS and GKE cluster, they all have the label `topology.kubernetes.io/zone` and `topology.kubernetes.io/region` on their node. Also, the K8s document give some explanation about this label: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone, it seems to me this label is also involved by K8s default scheduler. So I think we can trust these labels.
> 
> In terms of `reconfigure` action, you're right, the current description of the action isn't right, the `PV nodeAffinity` feature also have this problem, because we're introducing new constraints. I remember Nirali ever fix a [bug](https://vmturbo.atlassian.net/browse/OM-81611) to make the description more accurate for the `reconfigure` action, I'm wondering if we can do similar thing in `action-orchestrator` to generate a better description. I need spend some time investigate it, but I would appreciate any suggestion from you.

Hi @kevinwangcn , I think the description should spell out all possible missing label commodities, including those zone/region labels that are implicitly added.

**Comment by kevinwangcn** [comment]
> Are we planning to add some unit tests for the corresponding change? Otherwise, the change looks good to me - thanks @KevinWang0204!

@chlam4  thanks for reviewing this change.  I checked out the current unit test, we do have a function `Test_podEntityDTOBuilder_getPodCommoditiesBought_Error` that already cover my code change, but this function is checking if `getPodCommoditiesBought` returns any errors or not, we don't seem to have any other unit test to check if the bought commodity list is right or not. I'm wondering this would be good enough or we want to add more unit test on this.

**Comment by chlam4** [comment]
> > Are we planning to add some unit tests for the corresponding change? Otherwise, the change looks good to me - thanks @KevinWang0204!
> 
> @chlam4 thanks for reviewing this change. I checked out the current unit test, we do have a function `Test_podEntityDTOBuilder_getPodCommoditiesBought_Error` that already cover my code change, but this function is checking if `getPodCommoditiesBought` returns any errors or not, we don't seem to have any other unit test to check if the bought commodity list is right or not. I'm wondering this would be good enough or we want to add more unit test on this.

@KevinWang0204 I understand it takes extra effort to write unit tests to cover the entire `getPodCommoditiesBought()` function.  How about extracting your addition into a new routine and then just write unit tests for your routine?

**Comment by kevinwangcn** [comment]
> > > Are we planning to add some unit tests for the corresponding change? Otherwise, the change looks good to me - thanks @KevinWang0204!
> > 
> > 
> > @chlam4 thanks for reviewing this change. I checked out the current unit test, we do have a function `Test_podEntityDTOBuilder_getPodCommoditiesBought_Error` that already cover my code change, but this function is checking if `getPodCommoditiesBought` returns any errors or not, we don't seem to have any other unit test to check if the bought commodity list is right or not. I'm wondering this would be good enough or we want to add more unit test on this.
> 
> @KevinWang0204 I understand it takes extra effort to write unit tests to cover the entire `getPodCommoditiesBought()` function. How about extracting your addition into a new routine and then just write unit tests for your routine?

@chlam4   I've wrapped my change in a new function `getRegionZoneLabelCommodity` and also add a new unit test case for this new function

File: pkg/discovery/dtofactory/pod_entity_dto_builder.go [hunk 1]
```diff
@@ -533,3 +551,44 @@ func (builder *podEntityDTOBuilder) createContainerPodData(pod *api.Pod) *proto.
 	}
 	return podData
 }
+
```
pankajbatra1, line 554: Should we guard these changes with a FeatureGate?  I'm concerned about limited testing of topology.kubernetes.io/zone and topology.kubernetes.io/region on GKE and AKS and the possible need turn off the feature for certain clusters.

kevinwangcn: @pankajbatra1   thanks for reviewing this code change.  according to K8s official document [https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone), these 2 labels should be trusted, and my code change only works when the pod has a PV attached.  But for sure, I could add a new feature gate for safety. does this looks good to you?
`HonorRegionZoneLabels: {Default: false, PreRelease: featuregate.Alpha}`

kevinwangcn: @pankajbatra1 , I've updated my code by adding a feature gate for this functionality.  Its default value is `false`, if the user want to enable it, the user has to  add the following to the kubeturbo configmap, I tried it out, it works for me.
```
    "featureGates": {
        "HonorRegionZoneLabels": true
    }
```

pankajbatra1 (approved):

PR #721, opened 2022-08-31
Integration Test Unknown Node

cmedina003:
Background:
In some instances a kubelet could get in a bad state where it is unresponsive. In this case the Node should be reported as being in an unknown power state. These ginkgo specs will test that kubeturbo functionality works properly when this happens. These tests simulate the unresponsive kubelet by stoping the docker container of the node being tested on the kind cluster

Kubeturbo will report the nodes as:

unknown power state
listed in a new unknown group
Kubeturbo will report pods on this node as:

unknown power state
Implementation of ticket: https://vmturbo.atlassian.net/browse/OM-88238

Changes:

New Test Block for testing unknown node state
test for existence of unknown node group
test for proper status on the node
test for proper status on the pods
addition of a kind-worker on the kind cluster
bash util functions for stopping and starting containers
start and stop scripts for kind-workers
Testing:

only tested on a linux machine
pull kubeturbo
install [kind](https://kind.sigs.k8s.io/)
I downloaded everything manually, but you might be able to use this script to download binaries ./hack/download_test_binaries.sh
run script ./hack/create_kind_cluster.sh to create the cluster
run make integration
run ./build/integration.test -k8s-kubeconfig=$HOME/.kube/config -k8s-context=kind-kind -docker-user-name= -docker-user-password=


cmedina003:
Code is ready for review. Build failure is due to the unit tests and unrelated to my code. @kevinwangcn informed me this happens and won't affect anything.

hack/util.sh
```diff
@@ -96,3 +96,30 @@ function util::wait-for-condition() {
   fi
 }
 readonly -f util::wait-for-condition
+
+# util::stop-container stops a running docker container
+#
+# Arguments:
+#  - 1: container name for quering docker.
+function util::stop-container() {
```
kevinwangcn, line 104: As we already put `docker start/stop` in the Go code, do we still need to add this in the script?

test/integration/discovery.go
```diff
@@ -157,6 +288,129 @@ var _ = Describe("Discover Cluster", func() {
...
+
+		if condition(v) {
+			found = append(found, v)
+			howMany--
+		}
+	}
+	return found
+}
+
+func findNodeConditions(vals []corev1.NodeCondition, condition func(v corev1.NodeCondition) bool, howMany int) []corev1.NodeCondition {
```
kevinwangcn, line 336: we already have a function `NodeIsReady` in the file `node_util.go` which can help check if a node is in the ready status or not, maybe you can just use that one

test/integration/action_execution.go
```diff
@@ -379,7 +379,7 @@ func createDeployResource(client *kubeclientset.Clientset, dep *appsv1.Deploymen
 
 // This can also be bootstrapped from a test resource directory
 // which holds yaml files.
-func depSingleContainerWithResources(namespace, claimName string, replicas int32, withVolume, withGCLabel, paused bool) *appsv1.Deployment {
+func depSingleContainerWithResources(namespace, claimName string, replicas int32, withVolume, withGCLabel, paused bool, nodeName string, pullSecrets []corev1.LocalObjectReference) *appsv1.Deployment {
```
kevinwangcn, line 382: I don't think you need to add this `pullSecrets` parameter, please refer to `depMultiContainerWithResources`, you can directly add the `imagePullSecret` in pod's Spec even that secret might not exist at all. if we can do this, I guess we don't need the function `DockerImagePullSecretNames` any more.

test/integration/discovery.go
```diff
@@ -157,6 +288,129 @@ var _ = Describe("Discover Cluster", func() {
 	})
 })
 
+func findEntities(vals []*proto.EntityDTO, condition func(v *proto.EntityDTO) bool, howMany int) []*proto.EntityDTO {
```
kevinwangcn, line 291: Do we really need these `howMany` stuff?   It seems that we only need to find the first one element that match the condition and if it exists we can just return `true`, otherwise return `false` , I don't see the point to return a slice here.

hack/util.sh
cmedina003: We do not ned it for my code, but I figured it could be repurposed in case we need it. Do you think we should get rid of it?

test/integration/discovery.go
cmedina003: On line 255 I have to find multiple pods, but since most of the time I am just using it as a way to find one element I like being able to break out of the loop, since I no longer need to iterate over the range. It's more efficient really.

test/integration/discovery.go
```diff
@@ -147,6 +153,124 @@ var _ = Describe("Discover Cluster", func() {
...
+			}
+
+			// create a pod to test and attach to the node soon to be in a NotReady state
+			if deps == nil {
+				var err error = nil
+				deps, err = createDeployResource(kubeClient, depSingleContainerWithResources(namespace, "", 1, false, false, false, nodeName))
+				framework.ExpectNoError(err, "Error creating test resources")
+			}
+			// stop running node worker to simulate node in NotReady state
+			execute("docker", "stop", nodeName)
```
irfanurrehman, line 175: I would prefer to have the test fail in case of the exec error rather then simply log the error.

test/integration/discovery.go
```diff
@@ -147,6 +153,124 @@ var _ = Describe("Discover Cluster", func() {
...
+					return false, nil
+				}
+				return util.NodeIsReady(node), nil
+			}); err != nil {
+				framework.Failf("Failed to put node in a Ready state. %v", err)
+			}
+
+		})
+
+		It("unknown node should be detected", func() {
```
irfanurrehman, line 213: This is unnecessarily making the whole spec slower.
As far as I understand `BeforeEach()` and `AfterEach()` in the `Describe()` block is executed for each `It()` block, thus effectively starting and stopping the node 3 times over.
I think you can wrap all the three `It()` blocks in a single `Describe()` to make this faster and less error prone.

test/integration/discovery.go
```diff
@@ -147,6 +153,124 @@ var _ = Describe("Discover Cluster", func() {
...
+			found := getUnknownNode(entities)
+			if found == nil {
+				framework.Failf("Node with unknown state not found")
+			}
+
+			if found.GetDisplayName() == "" {
+				framework.Failf("unknown nodes should have a node name")
```
irfanurrehman, line 220: s/unknown/Unknown

test/integration/discovery.go
```diff
@@ -147,6 +153,124 @@ var _ = Describe("Discover Cluster", func() {
...
+			}
+
+		})
+
+		It("unknown nodes should be listed in the unknown node group", func() {
```
irfanurrehman, line 225: Please have each spec start with a capital letter, as each it is an individual test case.

test/integration/discovery.go
```diff
@@ -147,6 +153,124 @@ var _ = Describe("Discover Cluster", func() {
...
+			if unknownGroup.GetEntityType() != proto.EntityDTO_VIRTUAL_MACHINE {
+				framework.Failf("Unknown node group as incorrect entity type")
+			}
+
+			if unknownGroup.GetDisplayName() != "All Nodes (State Unknown)" {
+				framework.Failf("Unknown node group display name has been changed")
+			}
+
+			nodeInGroup := false
+			for _, uuid := range unknownGroup.GetMemberList().GetMember() {
```
irfanurrehman, line 242: You can make this better by checking the size of the slice (we are expecting only 1 node here) and then checking that that one members id matches

irfanurrehman (commented):
Thanks for doing this @cmedina003. Few nits and 1 timing improvement.
Timing improvement is important as the overall test run time is quite considerable already.

test/integration/discovery.go
cmedina003: > Thanks for doing this @cmedina003. Few nits and 1 timing improvement. Timing improvement is important as the overall test run time is quite considerable already.

Thank you for the review 😁 on my first kubeturbo MR!

test/integration/discovery.go
cmedina003: True, but if we keep it as is, then we would be able to handle cases where we added multiple unknown nodes. 😄

test/integration/discovery.go
cmedina003: Tried doing it in the describe, but it was causing failures, so I wrapped the setup and reset portions of the block in `It` functions before and after all the other unknown node tests.

kevinwangcn (approved):

PR #723, opened 2022-09-02
Handle the PR close event while executing gitops action

irfanurrehman:
**Intent**

Implements https://vmturbo.atlassian.net/browse/OM-88866

**Background**
As of now the PR mode action execution in the turbo gitops handling does not fail when the PR is closed, rather then being accepted by an external user. This implementation handles that and fails the action it has associated with a PR when that particular PR is closed.

**Testing**
Executed action wait on the PR to be merged
image

```
gitops-demo.
I0902 12:42:09.348923       1 github_manager.go:399] Creating new branch orig-minimal-specs-1e48bk9bg83jh.
I0902 12:42:11.352347       1 github_manager.go:460] PR #32 created. Will now wait for the PR to be merged to complete the action.
I0902 12:42:12.089676       1 remote_mediation_client.go:488] Sent action progress for 40.
```

Fails the action when PR is closed
image

### DIFF
--- a/pkg/action/executor/gitops/github_manager.go
+++ b/pkg/action/executor/gitops/github_manager.go
@@ -151,14 +151,36 @@ func (r *GitHubManager) WaitForActionCompletion(completionData interface{}) erro
 	// TODO: At some point we will need a better strategy to handle long running actions.
 	return wait.PollImmediate(20*time.Second, 24*7*time.Hour, func() (bool, error) {
 		handler := waitData.handler
-		isMerged, _, err := handler.client.PullRequests.IsMerged(handler.ctx, handler.user, handler.repo, waitData.prNum)
-		if err != nil {
-			return false, err
+		// Find out the status of the PR via the PR details and the isMerge API
+		pr, _, err := handler.client.PullRequests.Get(handler.ctx, handler.user, handler.repo, waitData.prNum)
+		if err != nil || pr == nil {
+			// We do not return error and exit here to ensure we keep retrying
+			// in case of transient network issues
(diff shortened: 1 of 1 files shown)


amd-ibm (approved):

ading1977 (approved):

pkg/action/executor/gitops/github_manager.go
```diff
@@ -151,14 +151,36 @@ func (r *GitHubManager) WaitForActionCompletion(completionData interface{}) erro
...
+					"Error accessing github API: %v.", waitData.prNum, err)
+				return false, nil
+			}
+			if isMerged {
+				return true, nil
+			}
+			return false, fmt.Errorf("the PR #%d was closed without merging", waitData.prNum)
+		default:
+			// We will retry in case of invalid PR state received
+			// TODO: a really rare scenario, but if we hit it implement a timeout expiry
```
pankajbatra1, line 179: Would this comment not also be applicable for line 161 above (err != nil || pr == nil)?  
You may want to indicate in log msg that we keep retrying.

irfanurrehman: Above, and for all such API calls at various places, we expect the API calls to succeed and we fail the action on unsuccessful API return (or call failure for example because of network).
In this particular case at line 155 above, the API failures seemingly could only be because of network partition, on which I thought it best to keep retrying until the situation resolves. The other possible failure could be auth failure, but if that is the case all previous API calls should fail (with one common credentail).
At line 180, I did put this comment, although I dont know if this will ever happen, because in case the API itself returns success but if for some reason (eg API version change) we see states different from what we know today, we should ideally handle that in advance.

PR #724, opened 2022-09-06
add controllable patched field value

cmedina003:
Ticket: https://vmturbo.atlassian.net/browse/OM-87209

Problem:
In Stitched environment pods were unable to be migrated to Google Cloud clusters and causing pod move failures. This was happening because the Google Cloud probe was setting the `controllable` field on the `ConsumerPolicy` as false, when it needed to be true. 

Fix:
This change will override the value coming from the cloud probe, making pod moves to the GKE cluster available.

Setup:
- add a GKE configuration target 

Testing:
- pull down kubeturbo
- run `export KUBETURBO_VERSION=8.6.4-SNAPSHOT`
- `make build`
- run `cd build; DOCKER_BUILDKIT=1 docker build -t <Docker Hub account name>/kubeturbo .`
- publish to your DockerHub account
- create a configmap and deployment pointed toward the same GKE cluster and deploy kubeturbo using the image you published to your docker hub. The image value should look something like this: `<account name>/kubeturbo:<image version>`
- going through the migrate container workload attempt to migrate a pod from a non-GKE cluster to the GKE cluster and notice that the move succeeds.

### DIFF
--- a/pkg/registration/supply_chain_factory.go
+++ b/pkg/registration/supply_chain_factory.go
@@ -95,7 +95,9 @@ var (
 	path                       = "path"
 	ActionEligibilityField     = "actionEligibility"
 	providerPolicyPath         = "providerPolicy"
+	consumerPolicyPath         = "consumerPolicy"
 	availableForPlacementField = "availableForPlacement"
+	controllableField          = "controllable"
 	powerStateField            = "powerState"
 )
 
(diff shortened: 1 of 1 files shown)


ading1977 (approved):
LGTM

pankajbatra1 (approved):

PR #729, opened 2022-09-16
[OM-88553]: Integration test for pod pv affinity rules

NiraliKotak:
Intent
Integration test to honor the nodeAffinity on the PVs of the pod if the pod has the PV attached when making a move action decision. Kubeturbo will add VMPM_ACCESS commodity for the pv affinity rules in Pod CommodityBoughtList. This test case validates whether the `CommodityBoughtList` in `entitydto` for the `pod` has `VMPM_ACCESS` with the correct  key (Example, `"kubernetes.io/hostname"`)
Background

Implementation
Creation of storage class with `Provisioner: "kubernetes.io/no-provisioner"`
Creation of PV with local path, affinity `key:value` as `kubernetes.io/hostname:kind-worker` and above storage class
Create a PVC with the storage class created
Create a deployment with volume and created PVC
Validate creation of test resources
Validate whether VMPM_ACCESS with correct key is present in the pod entitydto CommodityBoughtList 

Test Done
Full integration test run
Screen Shot 2022-09-16 at 8 02 43 AM

### DIFF
--- a/test/integration/action_execution.go
+++ b/test/integration/action_execution.go
@@ -10,6 +10,7 @@ import (
 	"github.com/turbonomic/turbo-go-sdk/pkg/proto"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


NiraliKotak:
Creating another PR from my repository

PR #730, opened 2022-09-16
Feature/OM:88834: Add k8s target during probe registration in secure probe

pallavidn:
Remote probes can now connect to Turbonomic XL server on a secure websocket connection.

**`kubeturbo`** was previously updated to use secret to establish the secure websocket connection while registering the probe. However, the k8s probe continues to require Turbo server admin credentials in order to configure a target when it starts up.

OM-85951 provides a mechanism in the SDK so that the probe can create a target for itself during the probe registration process. [This MR ](https://github.com/turbonomic/turbo-go-sdk/pull/140) updated the `turbo-go-sdk` to provide the same support to the remote go probes.
This MR is to implement changes in `kubeturbo` to use this capability and send target information as part of probe registration.
 
### **Changes in this review:**
Implement the sdk registration client interface to add target information. 
Target information is sent along with the probe info during probe registration when the probe starts up.

### **What is not changed:**
Retained the mechanism to add target using API when Turbo server operates in secure mode only.
If `**TargetIdentifier**` is not configured, target will be not added. Users will have to manually add the target via the UI

### **Tests with Non-Secure Server**
- Auto-add target by providing TargetIdentifier
- Auto-add target by providing TargetIdentifier & TargetType
`PreReq:` In both cases, Configure Turbo Admin User
`Observed: ` In both cases, Target is automatically added via API when the k8s probe pod starts up

- Manually add target by providing TargetType only
`PreReq: `Do not configure Turbo Admin User
`Observed: `Target is not added when the k8s probe starts up and should be added manually by the user via UI

**Other tests carried:**
- Delete target pod and check that same target is re-validated / rediscovered
- Provided client secret in the secrets. Since the server is not running secure mode, k8s probe continued to connect with the server on a non-secure websocket and used the api credentials (if present) to auto-add the target.

### Tests with Secure Server
`PreReq:` Configure the Probe client secret
`PreReq:` Do not configure Turbo Admin User

- Auto-add target by providing TargetIdentifier
- Auto-add target by providing TargetIdentifier & TargetType

`Observed:` In both cases,  Target is automatically added via Probe registration

- Manually add target by providing TargetType only
`Observed:` Target is not added when the probe starts up and should be added manually by user via UI

**Other tests carried:**
- Delete target pod and check that same target is re-validated / rediscovered
   **Note:** Server side work to allow deletion of target from the UI is in progress.
 
### Tests with Non-Secure Server+Targets Upgraded to Secure Server
Added 3 targets for 3 different clusters
- Auto-add target by providing TargetIdentifier
- Auto-add target by providing TargetIdentifier & TargetType
`PreReq:` In both cases, Configure Turbo Admin User
`Observed:` In both cases, Target is added when probe starts up via API

- Manually add target by providing TargetType only : **TODO**
`PreReq:` Do not configure Turbo Admin User
`Observed:` Target is not added when probe starts up but is manually added via UI

Then when server is upgraded, 
`Observed:` each of the targets stop to communicate

**`PreReq:`**  Obtain secret and configure secret for the k8s clusters (remove Turbo admin user - optional)
**`PreReq`:**  Redeploy the k8s image with new code

All targets should start communicating again

- Auto-add target by providing TargetIdentifier
- Auto-add target by providing TargetIdentifier & TargetType
`Observed:` In both cases, Target is registered by Probe. Since the same TargetIdentifier is used, the target is only updated, no new target is aded in the server.
`Observed:` Target resumes discovery

- Manually add target by providing TargetType only
**TODO:**

### Tests with Secure Server downgraded to Non-Secure mode
**TODO:**


pkg/k8s_tap_service.go
```diff
@@ -231,20 +233,34 @@ func NewKubernetesTAPService(config *Config) (*K8sTAPService, error) {
...
+		// Target is auto-added if an identifier is not provided.
+		// In this case, users can still add target via the UI.
+		// To ensure that the target added via the UI can communication with the server, it is necessary to configure
+		// 'TargetType' to uniquely identify the kubernetes cluster this probe going to monitor.
+		// Not configuring 'TargetType' is error-prone and not lead to correct discovery results.
+		if len(config.tapSpec.TargetType) > 0 {
+			glog.Infof("Not discovering target, add target via API or UI for target type %s", config.tapSpec.TargetType)
+		} else {
+			glog.Infof("Not discovering target, target type is not configured and discovery may not work correctly")
+		}
```
ading1977, line 255: This is a nice reminder!

pkg/k8s_tap_service.go
```diff
@@ -231,20 +233,34 @@ func NewKubernetesTAPService(config *Config) (*K8sTAPService, error) {
...
+		probeBuilder.WithSecureTargetProvider(registrationClient)
+
 		// The KubeTurbo TAP Service that will register the kubernetes target with the
 		// Turbonomic server and await for validation, discovery, action execution requests
-		glog.Infof("Should discover target %s", config.tapSpec.TargetIdentifier)
+		glog.Infof("Discovering target %s", config.tapSpec.TargetIdentifier)
 		probeBuilder = probeBuilder.DiscoversTarget(config.tapSpec.TargetIdentifier, discoveryClient)
 	} else {
-		glog.Infof("Not discovering target")
+		// Target is auto-added if an identifier is not provided.
```
ading1977, line 246: Do you mean Target is NOT auto-added if an identifier is not provided?

ading1977 (approved):

PR #732, opened 2022-09-21
Feature/om 88553 integration test pv with affinity

NiraliKotak:
Intent
Integration test to honor the nodeAffinity on the PVs of the pod if the pod has the PV attached when making a move action decision. Kubeturbo will add VMPM_ACCESS commodity for the pv affinity rules in Pod CommodityBoughtList. This test case validates whether the CommodityBoughtList in entitydto for the pod has VMPM_ACCESS with the correct key (Example, "kubernetes.io/hostname")
Background

Implementation
Creation of storage class with Provisioner: "kubernetes.io/no-provisioner"
Creation of PV with local path, affinity key:value as kubernetes.io/hostname:kind-worker and above storage class
Create a PVC with the storage class created
Create a deployment with volume and created PVC
Validate creation of test resources
Validate whether VMPM_ACCESS with correct key is present in the pod entitydto CommodityBoughtList

Test Done
Full integration test run

### DIFF
--- a/test/integration/action_execution.go
+++ b/test/integration/action_execution.go
@@ -695,7 +695,7 @@ func createVolumeClaim(client kubeclientset.Interface, namespace, storageClassNa
 		newPvc, errInternal = client.CoreV1().PersistentVolumeClaims(namespace).Create(context.TODO(), pvc, metav1.CreateOptions{})
 		if errInternal != nil {
 			glog.Errorf("Unexpected error while creating PVC for test: %v", errInternal)
-			return false, nil
+			return false, errInternal
 		}
 		return true, nil
 	}); err != nil {
@@ -751,7 +751,7 @@ func createPV(client kubeclientset.Interface, namespace, storageClassName string
 		newPV, errInternal = client.CoreV1().PersistentVolumes().Create(context.TODO(), pv, metav1.CreateOptions{})
 		if errInternal != nil {
 			glog.Errorf("Unexpected error while creating PV for test: %v", errInternal)
-			return false, nil
+			return false, errInternal
 		}
 		return true, nil
 	}); err != nil {
@@ -767,7 +767,7 @@ func createDeployment(client kubeclientset.Interface, dep *appsv1.Deployment) (*
 		newDep, errInternal = client.AppsV1().Deployments(dep.Namespace).Create(context.TODO(), dep, metav1.CreateOptions{})
 		if errInternal != nil {
 			glog.Errorf("Unexpected error while creating deployment: %v", errInternal)
-			return false, nil
+			return false, errInternal
 		}
 		return true, nil
 	}); err != nil {
@@ -783,7 +783,7 @@ func createReplicaSet(client kubeclientset.Interface, rs *appsv1.ReplicaSet) (*a
 		newRS, errInternal = client.AppsV1().ReplicaSets(rs.Namespace).Create(context.TODO(), rs, metav1.CreateOptions{})
 		if errInternal != nil {
 			glog.Errorf("Unexpected error while creating replicaset: %v", errInternal)
-			return false, nil
+			return false, errInternal
 		}
 		return true, nil
 	}); err != nil {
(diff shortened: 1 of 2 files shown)


test/integration/discovery.go
```diff
@@ -269,6 +269,56 @@ var _ = Describe("Discover Cluster", func() {
...
+			}
+
+			var bCommodityRegistered bool
+			var podEntityDTO *proto.EntityDTO
+			entityDTOs, _, err := discoveryClient.DiscoverWithNewFramework("discovery-integration-test")
+			framework.ExpectNoError(err, "Failed completing discovery of test cluster")
+			for _, entityDTO := range entityDTOs {
+				if *entityDTO.EntityType == proto.EntityDTO_CONTAINER_POD &&
+					strings.Contains(*entityDTO.DisplayName, pod.Name) {
+					podEntityDTO = entityDTO
```
irfanurrehman, line 301: You can break here once the pods DTO is found

NiraliKotak: Updated the code.

pankajbatra1 (approved):

test/integration/action_execution.go
```diff
@@ -736,14 +792,38 @@ func createReplicaSet(client kubeclientset.Interface, rs *appsv1.ReplicaSet) (*a
 	return newRS, nil
 }
 
+func createStorageClass(client kubeclientset.Interface) (*storagev1.StorageClass, error) {
```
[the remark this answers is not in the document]
NiraliKotak: The default storage class in kind cluster(i.e "standard") has VolumeBindingMode as WaitForFirstCustomer. Due to this the pv and pvc are not bound till the pod gets created. With the current implementation of deployment with `funcdepSingleContainerWithResources` and argument `withVolume` as `true` it is creates a new pv and the pvc gets bound to newly created pv. 

image

Creating new storageclass in the test with immediate binding mode, pvc gets bound to pv before deployment. The deployment is created with the pvc created.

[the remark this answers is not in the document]
irfanurrehman: I still think we can do without creating a new storage class... :-)
Nevertheless, this is done only in test setup, lets not bikeshed on this.

test/integration/discovery.go
```diff
@@ -269,6 +269,56 @@ var _ = Describe("Discover Cluster", func() {
...
+				}
+			}
+			for _, commI := range podEntityDTO.GetCommoditiesBought() {
+				for _, commI := range commI.Bought {
+					if *commI.CommodityType == proto.CommodityDTO_VMPM_ACCESS &&
+						strings.Contains(*commI.Key, "foo in (bar)") {
+						bCommodityRegistered = true
+						break
+					}
+				}
```
irfanurrehman, line 311: I am sorry for not pointing this out earlier. It did not strike me then. Shouldn't a corresponding seller side VMPM_ACCESS commodity be added on the nodes where this pod can go (nodes carrying this label). We are not checking that here?
Also, its best to move these checks in a check function (eg `checkAccessCommodities`) to make the steps in the test more readable.

test/integration/discovery.go
```diff
@@ -279,6 +324,58 @@ var _ = Describe("Discover Cluster", func() {
...
+	var podEntityDTO *proto.EntityDTO
+	var nodeEntityDTO *proto.EntityDTO
+
+	for _, entityDTO := range entityDTOs {
+		if *entityDTO.EntityType == proto.EntityDTO_CONTAINER_POD &&
+			*entityDTO.DisplayName == podName {
+			podEntityDTO = entityDTO
+			buyerRegsitered = true
+
+		} else if *entityDTO.EntityType == proto.EntityDTO_VIRTUAL_MACHINE &&
```
irfanurrehman, line 340: Thanks for doing this @NiraliKotak. This complete validation would be to check if the node(s) on which you have applied the label carries the access commodity and no other node has that commodity.

Also the whole logic here could be simpler. Look for Node dtos in a saperate section, and check its commodities in that loop. Look pod dto in another section and check its commodities (something similar to what you had earlier).

irfanurrehman: I am not too worried about the logic or your style of code. We need to however ensure that we do check for all conditions

NiraliKotak: @irfanurrehman Thanks for reviewing. In line 364 I am validating the particular node where the pod is hosted (which is the one where I have applied the label) has the Access commodity . Will it be okay if I update the code further verifying if the node which is labeled is the same as the node where pod(with pv) is hosted?

irfanurrehman: @NiraliKotak, no that is not what is needed. Because as part of the test setup, you have already ensured that the pod is on this node. What I meant is check other nodes from the cluster; they should not have the access commodity with this key in the sold commodities list.

PR #738, opened 2022-10-03
[OM-90117] Enable GoMemLimit feature to proactively trigger GC for kubeturbo

ading1977:
This PR takes advantage of the new [GoMemLimit](https://tip.golang.org/doc/gc-guide#Memory_limit) mechanism introduced in GO 1.19 to proactively trigger GC and prevent OOM.

This PR leverages an existing open source project `github.com/KimMachineGun/automemlimit` to automatically set go memory limit based on memory limit set on the `kubeturbo` container.

```
// Set Go runtime soft memory limit through the AUTOMEMLIMIT environment variable.
// AUTOMEMLIMIT configures how much memory of the cgroup's memory limit should be set as Go runtime
// soft memory limit in the half-open range (0.0,1.0].
// If AUTOMEMLIMIT is not set, it defaults to 0.9. This means 10% is the headroom for memory sources
// that the Go runtime is unaware of and unable to control.
// If GOMEMLIMIT environment variable is already set or AUTOMEMLIMIT=off, this function does nothing.
// AUTOMEMLIMIT_DEBUG environment variable enables debug logging of AUTOMEMLIMIT
```

This feature is in private preview. To enable this feature, configure:

```
"featureGates": {
    "GoMemLimit": true 
}
```


chlam4 (approved):

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 )
 
 require (
+	github.com/KimMachineGun/automemlimit v0.2.2
```
cmedina003, line 37: Are we worried about this not being supported at some point since it seems very new still?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -364,6 +365,22 @@ func (s *VMTServer) Run() {
...
+		// Set Go runtime soft memory limit: https://pkg.go.dev/runtime/debug#SetMemoryLimit
+		// Set Go runtime soft memory limit through the AUTOMEMLIMIT environment variable.
+		// AUTOMEMLIMIT configures how much memory of the cgroup's memory limit should be set as Go runtime
+		// soft memory limit in the half-open range (0.0,1.0].
+		// If AUTOMEMLIMIT is not set, it defaults to 0.9. This means 10% is the headroom for memory sources
+		// that the Go runtime is unaware of and unable to control.
+		// If GOMEMLIMIT environment variable is already set or AUTOMEMLIMIT=off, this function does nothing.
+		// AUTOMEMLIMIT_DEBUG environment variable enables debug logging of AUTOMEMLIMIT
+		_ = os.Setenv("AUTOMEMLIMIT_DEBUG", "true")
+		memlimit.SetGoMemLimitWithEnv()
```
cmedina003, line 379: Are we able to log the actual value being set?

go.mod
ading1977: These code are not being supported by anyone. They are opensource, so we just need to make sure we have viewed the source code and understand what it does. We run `go mod vendor` so the code is being preserved in our vendor directory in case it is removed from GitHub in the future.

cmd/kubeturbo/app/kubeturbo_builder.go
ading1977: Yes, they actual value is logged in the kubeturbo log.

PR #741, opened 2022-10-07
[OM-90331] Introduce limit on the number of resources returned on workload controller list API calls

irfanurrehman:
**Intent**
Introduces optimisation to handle https://vmturbo.atlassian.net/browse/OM-90331

**Background**
We use `list` api calls for query all resources including the workload controller resources from api server in our discovery cycles. In large topologies, we have seen that the numbers per cluster can be in tens of thousands for each resource, especially replicasets, where multiple older copy of the resources exist to retain history over deployment updates. This optimisations uses max items as `400` per list api call to cap the size of the buffers allocated by k8s client per API call.

**Implementation**
Use `ListOptions` `Limit` field to limit the number of items returned per query for the workload controller `List` api calls

**Testing**
The number of entities returned with and without the change are same with relationships (parent, grandparent pf pods) intact.

With code from kubeturbo latest master
image

With updated code (2 new pods came into existence)
image

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -131,6 +131,9 @@ type VMTServer struct {
 	// Garbage collection (leaked pods) interval config
 	GCIntervalMin int
 
+	// Number of workload controller items the list api call should request for
+	ItemsPerListQuery int
+
 	// The Openshift SCC list allowed for action execution
 	sccSupport []string
 
@@ -212,6 +215,7 @@ func (s *VMTServer) AddFlags(fs *pflag.FlagSet) {
 	fs.IntVar(&s.DiscoverySamples, "discovery-samples", DefaultDiscoverySamples, "The number of resource usage data samples to be collected from kubelet in each full discovery cycle. This should be no larger than 60.")
 	fs.IntVar(&s.DiscoverySampleIntervalSec, "discovery-sample-interval", DefaultDiscoverySampleIntervalSec, "The discovery interval in seconds to collect additional resource usage data samples from kubelet. This should be no smaller than 10 seconds.")
 	fs.IntVar(&s.GCIntervalMin, "garbage-collection-interval", DefaultGCIntervalMin, "The garbage collection interval in minutes for possible leaked pods from actions failed because of kubeturbo restarts. Default value is 20 mins.")
+	fs.IntVar(&s.ItemsPerListQuery, "items-per-list-query", 0, "Number of workload controller items the list api call should request for.")
 	fs.StringSliceVar(&s.sccSupport, "scc-support", defaultSccSupport, "The SCC list allowed for executing pod actions, e.g., --scc-support=restricted,anyuid or --scc-support=* to allow all. Default allowed scc is [restricted].")
 	// So far we have noticed cluster api support only in openshift clusters and our implementation works only for openshift
 	// It thus makes sense to have openshifts machine api namespace as our default cluster api namespace
@@ -365,8 +369,8 @@ func (s *VMTServer) Run() {
 		}
 	}
 
-	if utilfeature.DefaultFeatureGate.Enabled(features.GoMemLimit) {
-		glog.V(2).Info("GoMemLimit feature is enabled.")
+	if utilfeature.DefaultFeatureGate.Enabled(features.MemoryOptimisations) {
+		glog.V(2).Info("Memory Optimisations are enabled.")
 		// Set Go runtime soft memory limit: https://pkg.go.dev/runtime/debug#SetMemoryLimit
 		// Set Go runtime soft memory limit through the AUTOMEMLIMIT environment variable.
 		// AUTOMEMLIMIT configures how much memory of the cgroup's memory limit should be set as Go runtime
@@ -377,8 +381,30 @@ func (s *VMTServer) Run() {
 		// AUTOMEMLIMIT_DEBUG environment variable enables debug logging of AUTOMEMLIMIT
 		_ = os.Setenv("AUTOMEMLIMIT_DEBUG", "true")
 		memlimit.SetGoMemLimitWithEnv()
+		if s.ItemsPerListQuery == 0 {
+			limit, err := memlimit.FromCgroup()
+			if err != nil {
+				// This is very unlikely because in absense of any limit set in container resources
+				// we will get the cgroup limit as the nodes available/usable memory limit
+				glog.Errorf("Error retrieving memory limit (%v): %v.", limit, err)
+			} else if limit == 0 {
+				// This is very unlikely because in absense of any limit set in container resources
+				// we will get the cgroup limit as the nodes available/usable memory limit
+				glog.Error("Limit found set to zero (0).")
+			} else {
+				glog.V(2).Infof("Cgroup memlimit is set as %v.", limit)
+				util.ItemsPerListQuery = calculateListAPIItemsCount(float64(limit))
+			}
+		} else {
+			if s.ItemsPerListQuery < 10 {
+				glog.Warningf("Arg --items-per-list-query is set too low (%v), capping it to a minimum of 10 items.", s.ItemsPerListQuery)
+				s.ItemsPerListQuery = 10
+			}
+			util.ItemsPerListQuery = s.ItemsPerListQuery
+		}
+		glog.V(2).Infof("List Query Items Count is set to %v.", util.ItemsPerListQuery)
 	} else {
-		glog.V(2).Info("GoMemLimit feature is not enabled.")
+		glog.V(2).Info("Memory Optimisations are not enabled.")
 	}
 
 	// Collect target and probe info such as master host, server version, probe container image, etc
@@ -507,6 +533,11 @@ func handleExit(disconnectFunc disconnectFromTurboFunc) { // k8sTAPService *kube
 			disconnectFunc()
 		}
 	}()
+}
+
+// 5000 * m = where m is memlimit in GB (5000 items per GB of memlimit)
+func calculateListAPIItemsCount(memLimit float64) int {
+	return int(5000 * nodeUtil.Base2BytesToGigabytes(memLimit))
 }
 
 func discoverk8sAPIResourceGV(client *kubernetes.Clientset, resourceName string) (schema.GroupVersion, error) {
--- a/pkg/cluster/cluster_info_scraper.go
+++ b/pkg/cluster/cluster_info_scraper.go
@@ -35,8 +35,6 @@ const (
 	machineSetNodePoolPrefix = "machineset"
 	// Expiration of cached pod controller info.
 	defaultCacheTTL = 12 * time.Hour
-	// Number of max items we query in each workload controller list API calls
-	defaultMaxListQueryItems = 400
 )
 
 var (
@@ -283,8 +281,11 @@ func (s *ClusterScraper) GetAllEndpoints() ([]*api.Endpoints, error) {
 }
 
 func (s *ClusterScraper) GetResources(resource schema.GroupVersionResource) ([]unstructured.Unstructured, error) {
-	if !utilfeature.DefaultFeatureGate.Enabled(features.PaginateAPICalls) {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.MemoryOptimisations) {
 		list, err := s.DynamicClient.Resource(resource).Namespace(api.NamespaceAll).List(context.TODO(), metav1.ListOptions{})
+		if err != nil || list == nil {
+			return nil, err
+		}
 		return list.Items, err
 	}
 
@@ -293,7 +294,7 @@ func (s *ClusterScraper) GetResources(resource schema.GroupVersionResource) ([]u
 	// TODO: Is there a possibility of this loop never exiting?
 	// The documentation states that the APIs reliably return correct values for continue
 	for {
-		listOptions := metav1.ListOptions{Limit: int64(defaultMaxListQueryItems), Continue: continueList}
+		listOptions := metav1.ListOptions{Limit: int64(commonutil.ItemsPerListQuery), Continue: continueList}
 		listItems, err := s.DynamicClient.Resource(resource).Namespace(api.NamespaceAll).List(context.TODO(), listOptions)
 		if err != nil {
 			return items, err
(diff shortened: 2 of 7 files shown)


ading1977:
By the way, do we need a featureGates for this change? I believe this is still under private preview https://vmturbo.atlassian.net/browse/OM-90116.

pkg/cluster/cluster_info_scraper.go
```diff
@@ -33,6 +33,8 @@ const (
 	machineSetNodePoolPrefix = "machineset"
 	// Expiration of cached pod controller info.
 	defaultCacheTTL = 12 * time.Hour
+	// Number of max items we query in each workload controller list API calls
+	defaultMaxListQueryItems = 400
```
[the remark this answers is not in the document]
irfanurrehman: @ading1977 this is open for suggestions. I did use 400 as an initial number which seems reasonably balanced buffer size. If it makes sense to make this configurable, I can do that.

irfanurrehman:
> By the way, do we need a featureGates for this change? I believe this is still under private preview https://vmturbo.atlassian.net/browse/OM-90116.

I will add a feature flag to this.

pkg/discovery/processor/controller_processor.go
```diff
@@ -97,7 +97,7 @@ func (cp *ControllerProcessor) cacheAllControllers() {
 			}
 			continue
 		}
-		for _, item := range list.Items {
+		for _, item := range items {
```
ading1977, line 100: Is it possible to only cache the active controllers? For example:

* Skip `ReplicaSet` that has a `Deployment` owner, but has 0 replicas
* Similar logic to skip inactive `ReplicationController` that has `DeploymentConfig` owner

irfanurrehman: @ading1977  Updated the code to implement this, PTAL!

irfanurrehman:
> > By the way, do we need a featureGates for this change? I believe this is still under private preview https://vmturbo.atlassian.net/browse/OM-90116.
> 
> I will add a feature flag to this.

@ading1977 
Introduced a feature flag named `PaginateAPICalls`.

Additionally I am working on introducing the code which can deduce the number of controllers to ask for in the list API call based on the rough `Sizing formula` you have listed in https://vmturbo.atlassian.net/browse/OM-90118.
Will update the code to get the memlimit used (if there is one) and find the value of `n` using the `MEMLIMIT` via:
`2GB * n + 1GB = 0.9*MEMLIMIT`

If there is no `MEMLIMIT` set, we use some default number (I think 400 or 500 is appropriate, but you can suggest another)

pkg/discovery/processor/controller_processor.go
```diff
@@ -130,3 +133,22 @@ func (cp *ControllerProcessor) cacheAllControllers() {
 	}
 	cp.KubeCluster.ControllerMap = controllerMap
 }
+
+func cacheController(obj unstructured.Unstructured) bool {
```
ading1977, line 137: `cacheController` ->  `shouldCacheController` ?

pkg/discovery/processor/controller_processor.go
```diff
@@ -130,3 +133,22 @@ func (cp *ControllerProcessor) cacheAllControllers() {
...
+	if obj.GetKind() != util.ReplicaSetResName &&
+		obj.GetKind() != util.ReplicationControllerResName {
+		return true
+	}
```
ading1977, line 141: Shall we also check `ownerReference`? That is, we only continue to check number of replicas when:
* this is a `ReplicaSet` and the owner is a `Deployment`, or
* this is a `ReplicationController` and the owner is a `DeploymentConfig`

ading1977:
> Additionally I am working on introducing the code which can deduce the number of controllers to ask for in the list API call based on the rough `Sizing formula` you have listed in https://vmturbo.atlassian.net/browse/OM-90118. Will update the code to get the memlimit used (if there is one) and find the value of `n` using the `MEMLIMIT` via: `2GB * n + 1GB = 0.9*MEMLIMIT`
> 
> If there is no `MEMLIMIT` set, we use some default number (I think 400 or 500 is appropriate, but you can suggest another)

* First, the sizing guide should be changed to `(n+1)GB`, since we have already enabled the **GOMEMLIMIT**. @chlam4 , could you please confirm?
* Assuming the formula is now `(n+1)GB = 0.9*MEMLIMIT`, does it mean that we will get `1000*n` items for each call? I am not sure I fully understand this.
* If there is no container memory limit, I think we can assume a default 16GB memory limit and use that in the formula.

By the way, to get the container memory limit, you can use [this](https://github.com/KimMachineGun/automemlimit/blob/1a256276825df90440a33d9b6fe9385f2138c2f3/memlimit/cgroups.go#L16). We already use this for our **GOMEMLIMIT** support.

pkg/features/features.go
```diff
@@ -49,6 +49,13 @@ const (
 	// This gate enables Go runtime soft memory limit as explained in
 	// https://pkg.go.dev/runtime/debug#SetMemoryLimit
 	GoMemLimit featuregate.Feature = "GoMemLimit"
+
+	// PaginateAPICalls owner: @irfanurehman
+	// alpha:
+	//
+	// Pagination support for list API calls to API server querying workload controllers
+	// Without this feature gate the whole list is requested in a single list API call.
+	PaginateAPICalls featuregate.Feature = "PaginateAPICalls"
```
ading1977, line 58: Do you think if it makes sense to combine the `GoMemLimit` and `PaginateAPICalls` into one feature flag? I don't think it makes sense to enable one feature while disabling the other one. What do you think? I haven't thought of a good name though ...

chlam4:
> > Additionally I am working on introducing the code which can deduce the number of controllers to ask for in the list API call based on the rough `Sizing formula` you have listed in https://vmturbo.atlassian.net/browse/OM-90118. Will update the code to get the memlimit used (if there is one) and find the value of `n` using the `MEMLIMIT` via: `2GB * n + 1GB = 0.9*MEMLIMIT`
> > If there is no `MEMLIMIT` set, we use some default number (I think 400 or 500 is appropriate, but you can suggest another)
> 
> * First, the sizing guide should be changed to `(n+1)GB`, since we have already enabled the **GOMEMLIMIT**. @chlam4 , could you please confirm?
> * Assuming the formula is now `(n+1)GB = 0.9*MEMLIMIT`, does it mean that we will get `1000*n` items for each call? I am not sure I fully understand this.
> * If there is no container memory limit, I think we can assume a default 16GB memory limit and use that in the formula.
> 
> By the way, to get the container memory limit, you can use [this](https://github.com/KimMachineGun/automemlimit/blob/1a256276825df90440a33d9b6fe9385f2138c2f3/memlimit/cgroups.go#L16). We already use this for our **GOMEMLIMIT** support.

@ading1977 Technically that's right.  We could also be even a bit more conservative to leave more buffer, but what's stated above should be a good start to test.

pkg/discovery/processor/controller_processor.go
irfanurrehman: @ading1977 I think we also support the use case where a pod could be owned only by rs or rc  and not configured by a dep or dc (although almost nobody uses this).

irfanurrehman:
> * First, the sizing guide should be changed to `(n+1)GB`, since we have already enabled the **GOMEMLIMIT**. @chlam4 , could you please confirm?
> * Assuming the formula is now `(n+1)GB = 0.9*MEMLIMIT`, does it mean that we will get `1000*n` items for each call? I am not sure I fully understand this.
> * If there is no container memory limit, I think we can assume a default 16GB memory limit and use that in the formula.

I am not really sure either. This literally means that each resource takes approximately 1Mb in memory for the overall processing of the list API call, keeping the 1Gb additional buffer aside. I don't have the insight into the profiling data, but I am willing to take some baseline and work with it. This is what I m going to do:
- Use your recommendation ie n + 1 Gb = .9*MEMLIMIT (each n = 1K resources)
- To add some more buffer space to the same, convert the above to .8n + 1 Gb = .9*MEMLIMIT ie each Gb of memory limit that we have can be used to process 800 resources.

chlam4:
> > * First, the sizing guide should be changed to `(n+1)GB`, since we have already enabled the **GOMEMLIMIT**. @chlam4 , could you please confirm?
> > * Assuming the formula is now `(n+1)GB = 0.9*MEMLIMIT`, does it mean that we will get `1000*n` items for each call? I am not sure I fully understand this.
> > * If there is no container memory limit, I think we can assume a default 16GB memory limit and use that in the formula.
> 
> I am not really sure either. This literally means that each resource takes approximately 1Mb in memory for the overall processing of the list API call, keeping the 1Gb additional buffer aside.

Actually I realized my above estimation was for deployments (not replica sets).  There's a factor of 8 between deployments and replica sets in the environment that I made the estimation.  With that said, a better formula would be `(n/8+1) = 0.9*MEMLIMIT`.

irfanurrehman:
> Actually I realized my above estimation was for deployments (not replica sets). There's a factor of 8 between deployments and replica sets in the environment that I made the estimation. With that said, a better formula would be `(n/8+1) = 0.9*MEMLIMIT`.
> 
> More details of the estimation:
> 
> * In that environment we did estimation, the raw json data of the 16K replica sets were measured to 190MB.  Usually the GO structure required to deserialize is about 4x-5x.  In this case, it will be 760MB-950MB.  In other words, about 16K resources per 1GB memory.
> * On the other hand, the observed peak memory usage in that environment was 1.7GB.  That's about 10K resources per GB, but that's a conservative estimation.
> 
> As discussed just now offline, we will set the per GB query limit to 5K to make it even more conservative to account for the size and structure differences between replica sets and other workload controllers.

Thanks @chlam4 as discussed offline, I have updated the calculation to use 5k resources per GB of memlimit available and provided `--list-query-items-count` to be able to override all calculations and defaults and set a spcific value

pkg/features/features.go
irfanurrehman: Made these into one flag `MemoryOptimisations=true`

irfanurrehman:
@ading1977 @chlam4 updated this as per our discussions
The image is available as `irfdocker/kubeturbo:list-optimisation`

pkg/discovery/processor/controller_processor.go
ading1977: That is my understanding too. But with the current logic, if an rs or rc does not have owner, but has 0 replicas, then we won't be able to cache it. But in fact we do want to cache these controllers. Did I miss anything?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -131,6 +131,9 @@ type VMTServer struct {
 	// Garbage collection (leaked pods) interval config
 	GCIntervalMin int
 
+	// Number of workload controller items the list api call should request for
+	ListQueryItemsCount int
```
ading1977, line 135: Would be it better to call this `ItemsPerQuery`, or `ItemsPerListQuery`?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -377,8 +381,22 @@ func (s *VMTServer) Run() {
...
+			if err != nil || limit == 0 {
+				// This is very unlikely because in absense of any limit set in container resources
+				// we will get the cgroup limit as the nodes available/usable memory limit
+				glog.Errorf("Error retrieving memory limit (%v) or limit is not set: %v.", limit, err)
+			} else {
+				glog.V(2).Infof("Cgroup memlimit is set as %v.", limit)
+				util.ListQueryItemsCount = calculateListAPIItemsCount(float64(limit))
+			}
+		} else {
+			util.ListQueryItemsCount = s.ListQueryItemsCount
```
ading1977, line 395: Do we want to do some validation on this number (negative value, for example)? Maybe set some minimum value?

cmd/kubeturbo/app/kubeturbo_builder.go
irfanurrehman: > That is my understanding too. But with the current logic, if an rs or rc does not have owner, but has 0 replicas, then we won't be able to cache it. But in fact we do want to cache these controllers. Did I miss anything?

I don't understand the reason to cache a controller which has 0 replicas, in effect no pod associated. I think we don't build a DTO for such controllers. It might be that I am missing something.. :-)

> Would be it better to call this ItemsPerQuery, or ItemsPerListQuery?

Updated to `ItemsPerListQuery`

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -377,8 +381,22 @@ func (s *VMTServer) Run() {
 		// AUTOMEMLIMIT_DEBUG environment variable enables debug logging of AUTOMEMLIMIT
 		_ = os.Setenv("AUTOMEMLIMIT_DEBUG", "true")
 		memlimit.SetGoMemLimitWithEnv()
+		if s.ListQueryItemsCount == 0 {
+			limit, err := memlimit.FromCgroup()
+			if err != nil || limit == 0 {
+				// This is very unlikely because in absense of any limit set in container resources
+				// we will get the cgroup limit as the nodes available/usable memory limit
+				glog.Errorf("Error retrieving memory limit (%v) or limit is not set: %v.", limit, err)
```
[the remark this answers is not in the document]
irfanurrehman: The thought behind this was if err found then the second condition would not be checked (l to r evaluation of OR). But now I am not sure if golang does evaluate or as left to right. Updated to handle this as 2 separate conditions thus.

cmd/kubeturbo/app/kubeturbo_builder.go
ading1977: > > That is my understanding too. But with the current logic, if an rs or rc does not have owner, but has 0 replicas, then we won't be able to cache it. But in fact we do want to cache these controllers. Did I miss anything?
> 
> I don't understand the reason to cache a controller which has 0 replicas, in effect no pod associated. I think we don't build a DTO for such controllers. It might be that I am missing something.. :-)
> 
You are right, I forgot that we don't create DTO for controllers with 0 replicas (the same applies to services with no replicas). Then this is not a concern any more.

ading1977 (approved):
LGTM. Thanks for the great work @irfanurrehman !

irfanurrehman:
> LGTM. Thanks for the great work @irfanurrehman !

Thanks for your ardent review @ading1977

PR #743, opened 2022-10-18
init impl of static pod daemon sets

cmedina003:
bug: https://vmturbo.atlassian.net/browse/OM-90831

# References:
https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/
https://cloud.google.com/kubernetes-engine/docs/concepts/network-overview

# background:
 We discovered that rather than implementing kube-proxy with daemon sets, GKE has used static pods on every node. This change will either check every node or every node pool for pods with the same intent. 
 
 
# description:
 Before dispatching workers the code will find all the static pods acting as daemons and mark them as such. We will find the different groups of pods by a common prefix that the user would have to consistently use during creation on the node. After all are found the code keeps a mapping in memory of the static pods that are also daemons for creation during pod building.

# Testing:
- build the docker image
- deploy to your docker hub account
- deploy to any GKE cluster
- move all pods except the mirror pod to other nodes
    - Use kubectl cordon or kubectl drain to remove all pods except for the mirror pod (You wouldn't be able to move the mirror pod).
 - notice that the VM will have a suspend action for both the pod and the VM 
 
 
 Afterwords the UI should show something like below:
 
Screen Shot 2022-10-18 at 9 50 58 AM


pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -115,6 +117,11 @@ func (builder *podEntityDTOBuilder) WithPendingPods(pendingPods []*api.Pod) *pod
 	return builder
 }
 
+func (builder *podEntityDTOBuilder) WithStaticPodToDaemonMap(staticPodToDaemonMap map[string]bool) *podEntityDTOBuilder {
```
pankajbatra1, line 120: Please run integration tests and report results.

pkg/discovery/util/node_util.go
```diff
@@ -217,3 +218,21 @@ func DetectHARole(node *api.Node) bool {
...
+				nodePools[pool] = sets.NewString(node.Name)
+			} else {
+				nodePools[pool].Insert(node.Name)
+			}
+		}
+	}
+	glog.V(4).Infof("Found %+v node pool keys.", len(nodePools))
+	glog.V(5).Info(nodePools)
+	return nodePools
+}
```
ading1977, line 238: How about when the `nodePools` map is empty, you return a map with the key: **Cluster**, value: a set of all nodes in the cluster (i.e., a fake pool with all nodes in the cluster). Then you don't need to have the logic later on to check if there are node pools or not.

pkg/discovery/repository/kube_cluster.go
```diff
@@ -192,6 +195,67 @@ func (summary *ClusterSummary) computePodToServiceMap() {
...
+	mirrorPods := util.GetMirrorPods(summary.Pods)
+	// common name prefix is our assumed conventions for showing intent of creating
+	// daemon sets with static pods
+	prefixToNodeNames := util.GetMirrorPodPrefixToNodeNames(mirrorPods)
+	nodePoolToNodeNames := util.MapNodePoolToNodeNames(summary.Nodes)
+	anyPools := len(nodePoolToNodeNames) != 0
+	if anyPools {
+		glog.V(3).Info("Conducting static pod daemon check with node pools.")
+	} else {
+		glog.V(3).Info("Conducting static pod daemon check with nodes.")
```
ading1977, line 211: As mentioned, if `MapNodePoolToNodeNames` returns a fake pool to represent the entire cluster, then you don't need the logic of `anyPools`, etc.

pkg/discovery/repository/kube_cluster.go
```diff
@@ -192,6 +195,67 @@ func (summary *ClusterSummary) computePodToServiceMap() {
...
 }
 
+func (summary *ClusterSummary) computeStaticPodToDaemonMap() {
+	glog.V(2).Info("creating mapping for static pods modeled as daemons.")
+	// static pods are only exposed through the API via mirror pods
+	// https://kubernetes.io/docs/reference/glossary/?all=true#term-mirror-pod
+	mirrorPods := util.GetMirrorPods(summary.Pods)
+	// common name prefix is our assumed conventions for showing intent of creating
+	// daemon sets with static pods
+	prefixToNodeNames := util.GetMirrorPodPrefixToNodeNames(mirrorPods)
```
ading1977, line 205: Can we combine the two calls to eliminate one round of iteration?

pkg/discovery/util/pod_util.go
```diff
@@ -89,16 +89,54 @@ func HasController(pod *api.Pod) bool {
...
+func GetMirrorPods(pods []*api.Pod) []*api.Pod {
+	glog.V(4).Info("Getting mirror pods.")
+	mirrorPods := []*api.Pod{}
+	for _, pod := range pods {
+		if isMirrorPod(pod) {
+			mirrorPods = append(mirrorPods, pod)
+		}
+	}
+	glog.V(4).Infof("Found %+v mirror pods", len(mirrorPods))
+	glog.V(5).Info(mirrorPods)
```
ading1977, line 109: V(4) -> V(3)
V(5) -> V(4) (we don't have V(5))

pkg/discovery/util/pod_util.go
```diff
@@ -89,16 +89,54 @@ func HasController(pod *api.Pod) bool {
...
+		prefix, ok := GetMirrorPodPrefix(pod)
+		if ok && prefixToNodeNames[prefix] == nil {
+			prefixToNodeNames[prefix] = sets.NewString(pod.Spec.NodeName)
+		} else if ok {
+			prefixToNodeNames[prefix].Insert(pod.Spec.NodeName)
+		}
+	}
+
+	glog.V(4).Infof("Found %+v static pod prefix keys.", len(prefixToNodeNames))
+	glog.V(5).Info(prefixToNodeNames)
```
ading1977, line 126: Same as before.

pkg/discovery/repository/kube_cluster.go
cmedina003: I could, but it would be same amount of work and this way it separates functionality for future use.

pkg/discovery/util/pod_util.go
cmedina003: Thank you! Good catch!

pkg/discovery/util/node_util.go
cmedina003: I like this potentially being reusable and if I add a Cluster key, that would be an implementation detail the caller would have to know about prior to using this function.

pkg/discovery/repository/kube_cluster.go
ading1977: If creating a fake pool inside `MapNodePoolToNodeNames` is a concern, maybe we can create it here when `antyPools` is false? Then we can simplify the logic below.

pkg/discovery/repository/kube_cluster.go
```diff
@@ -192,6 +195,67 @@ func (summary *ClusterSummary) computePodToServiceMap() {
...
+				}
+			}
+		} else {
+			for _, node := range summary.Nodes {
+				if !nodeNames.Has(node.GetName()) {
+					daemon = false
+					break
+				}
+			}
+		}
```
ading1977, line 244: Again, this part of the logic could be simplified if we treat the entire cluster as a single node pool, it should not be treated any different than other node pools.

Also, I believe the logic should be:

* Initialize `daemon` to false
* As long as one node pool matches the nodeSet that corresponds to a prefix, then set `daemon` to true and break

pkg/discovery/util/node_util.go
cmedina003: adding the cluster key to the calling code.

pkg/discovery/util/pod_util.go
```diff
@@ -89,16 +89,54 @@ func HasController(pod *api.Pod) bool {
...
+	glog.V(3).Info("maping mirror pod prefixes to node names.")
+	prefixToNodeNames := make(map[string]sets.String)
+	for _, pod := range pods {
+		prefix, ok := GetMirrorPodPrefix(pod)
+		if ok && prefixToNodeNames[prefix] == nil {
+			prefixToNodeNames[prefix] = sets.NewString(pod.Spec.NodeName)
+		} else if ok {
+			prefixToNodeNames[prefix].Insert(pod.Spec.NodeName)
+		}
+	}
```
ading1977, line 123: Nit:

```
		prefix, ok := GetMirrorPodPrefix(pod)
		if !ok {
			// Not a mirror pod
			continue
		}
		nodeNames, found := prefixToNodeNames[prefix]
		if !found {
			prefixToNodeNames[prefix] = sets.NewString(pod.Spec.NodeName)			
		} else {
			nodeNames.Insert(pod.Spec.NodeName)
		}

```

pkg/discovery/util/node_util.go
```diff
@@ -217,3 +218,43 @@ func DetectHARole(node *api.Node) bool {
...
+
+func IsSuperset[T NodeLike](nodeNames sets.String, nodes []T, extractor func(T) string) bool {
+	for _, node := range nodes {
+		if !nodeNames.Has(extractor(node)) {
+			return false
+		}
+	}
+	return true
+}
```
ading1977, line 252: This is interesting, I have never used Generics with Go. Just out of curiosity, why do we need to have a generic type here if the only type we need is `*v1.Node`? 

By the way, thanks for checking super set, I think this is better than doing equal.

pkg/discovery/repository/kube_cluster.go
```diff
@@ -192,6 +196,59 @@ func (summary *ClusterSummary) computePodToServiceMap() {
...
+	nodePoolToNodes := util.MapNodePoolToNodes(summary.Nodes, summary.MachineSetToNodesMap)
+	// additionally includes every node as its own node group with a generated key to ensure uniqueness
+	nodePoolToNodes["cluster-"+base64.StdEncoding.EncodeToString([]byte(summary.Name))] = summary.Nodes
+	staticPodToDaemonMap := make(map[string]bool)
+	// cache of static pod prefix to flag for quicker look up
+	prefixToDaemon := make(map[string]bool)
+	daemonCount := 0
+	for _, pod := range mirrorPods {
+		prefix, ok := util.GetMirrorPodPrefix(pod)
+		nodeNames := prefixToNodeNames[prefix]
```
ading1977, line 218: Could `nodeNames` ever be `nil`, or the `prefix` is guaranteed to exist in the map?

pkg/discovery/util/node_util.go
cmedina003: So I was originally using this function for checking node names in a set with node names from an slice, but then I needed it for a node slice so I figured I would keep the generic version. Over time we might even use something like a Byte[] or NodeMap.

ading1977:
By the way, I wonder why there are so many changes in go mod and vendor, did we do the following?

`go mod tidy`
`go mod vendor`

cmedina003:
> By the way, I wonder why there are so many changes in go mod and vendor, did we do the following?
> 
> `go mod tidy` `go mod vendor`

I ran both. It seems like this happened, because I updated the go mod to 1.19.

pkg/discovery/repository/kube_cluster.go
cmedina003: There is no chance, but I can write some defensive code.

ading1977 (approved):

PR #744, opened 2022-10-18
Revert "[OM-90331] Introduce limit on the number of resources returned on workload controller list API calls"

ading1977:
Reverts turbonomic/kubeturbo#741


ading1977:
Close as the build failed for unknown reason.

PR #746, opened 2022-10-19
[OM-90331] Memory optimisations with updated feature flag name

irfanurrehman:
**Intent**
This is resubmitting the changes as merged in https://github.com/turbonomic/kubeturbo/pull/741 and later reverted by https://github.com/turbonomic/kubeturbo/pull/745. This was done to ensure correct feature flag compatibility and ensure that the changes go in the next release and not the current one.

The original PR introduces optimisations to handle https://vmturbo.atlassian.net/browse/OM-90331

**Background**
We use `list` api calls for query all resources including the workload controller resources from api server in our discovery cycles. In large topologies, we have seen that the numbers per cluster can be in tens of thousands for each resource, especially replicasets, where multiple older copy of the resources exist to retain history over deployment updates. This optimisations uses max items as `400` per list api call to cap the size of the buffers allocated by k8s client per API call.

**Implementation**
Use `ListOptions` `Limit` field to limit the number of items returned per query for the workload controller `List` api calls

**Testing**
The number of entities returned with and without the change are same with relationships (parent, grandparent pf pods) intact.

With code from kubeturbo latest master
image

With updated code (2 new pods came into existence)
image

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -131,6 +131,9 @@ type VMTServer struct {
 	// Garbage collection (leaked pods) interval config
 	GCIntervalMin int
 
+	// Number of workload controller items the list api call should request for
+	ItemsPerListQuery int
+
 	// The Openshift SCC list allowed for action execution
 	sccSupport []string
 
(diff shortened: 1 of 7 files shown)


irfanurrehman:
rebased on latest!

ading1977:
@irfanurrehman , feel free to merge this. We discussed in today's standup that we don't have to wait for the test result from our SaaS environment.

chlam4 (approved):

irfanurrehman:
Thanks @ading1977 and @chlam4.

PR #750, opened 2022-10-31
[OM-91845]: Log the expected mount path for kubeturbo secret

pallavidn:
When reviewing [OM-91845](https://vmturbo.atlassian.net/browse/OM-91845) we saw a customer who took the kubeturbo yamls and customized them to fit within their CD deployment process.

For this bug, the customer changed the mount path of the secret into a variable. The secret name itself can be customized but not the mount path which requires /etc/turbonomic-credentials only.

This PR contains the change to log the expected mount path when the Turbo server credentials are not found by kubeturbo.

### DIFF
--- a/pkg/k8s_tap_service.go
+++ b/pkg/k8s_tap_service.go
@@ -26,6 +26,7 @@ import (
 const (
 	defaultUsername      = "defaultUser"
 	defaultPassword      = "defaultPassword"
+	credentialsDirPath   = "/etc/turbonomic-credentials"
 	usernameFilePath     = "/etc/turbonomic-credentials/username"
 	passwordFilePath     = "/etc/turbonomic-credentials/password"
 	clientIdFilePath     = "/etc/turbonomic-credentials/clientid"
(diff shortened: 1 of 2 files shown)


pkg/k8s_tap_service.go
```diff
@@ -67,6 +68,10 @@ func ParseK8sTAPServiceSpec(configFile string, defaultTargetName string) (*K8sTA
 		tapSpec.TargetIdentifier = defaultTargetName
 	}
 
+	if _, err := os.Stat(credentialsDirPath); os.IsNotExist(err) {
+		glog.V(2).Infof("credentials mount path is unavailable. Checked path: %s", credentialsDirPath)
```
chlam4, line 72: Nit: will it be a bit more explicit to say "unavailable" -> "does not exist"?

pkg/k8s_tap_service.go
```diff
@@ -93,21 +98,21 @@ func loadOpsMgrCredentialsFromSecret(tapSpec *K8sTAPServiceSpec) error {
 	// Return unchanged if the mounted file isn't present
 	// for backward compatibility.
 	if _, err := os.Stat(usernameFilePath); os.IsNotExist(err) {
-		glog.V(3).Infof("credentials from secret unavailable. Checked path: %s", usernameFilePath)
+		glog.V(2).Infof("server api credentials from secret unavailable. Checked path: %s", usernameFilePath)
 		return nil
 	}
 	if _, err := os.Stat(passwordFilePath); os.IsNotExist(err) {
-		glog.V(3).Infof("credentials from secret unavailable. Checked path: %s", passwordFilePath)
+		glog.V(2).Infof("sever api credentials from secret unavailable. Checked path: %s", passwordFilePath)
```
chlam4, line 105: typo: "sever" -> "server".

chlam4 (approved):
A couple of nit comments.  Otherwise, it looks good to me!

PR #751, opened 2022-10-31
[OM-92009]: add flag to enable pod move across cluster

libai98:
code for adding a flag option to enable pod move across cluster.

### DIFF
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -64,7 +64,6 @@ const (
 	DefaultDiscoverySampleIntervalSec = 60
 	DefaultGCIntervalMin              = 10
 	DefaultReadinessRetryThreshold    = 60
-	DefaultEnablePodClusterMoves      = false
 )
 
 var (
@@ -135,9 +134,8 @@ type VMTServer struct {
 	// The Openshift SCC list allowed for action execution
 	sccSupport []string
 
-	// Allow pod move across cluster
-	// The default is false
-	enablePodClusterMoves bool
+	// Injected Cluster Key to enable pod move across cluster
+	ClusterKeyInjected string
 
 	// Force the use of self-signed certificates.
 	// The default is true.
@@ -188,7 +186,7 @@ func NewVMTServer() *VMTServer {
 
 // AddFlags adds flags for a specific VMTServer to the specified FlagSet
 func (s *VMTServer) AddFlags(fs *pflag.FlagSet) {
-	fs.BoolVar(&s.enablePodClusterMoves, "enable-pod-cluster-move", false, "Enable pod move across cluster")
+	fs.StringVar(&s.ClusterKeyInjected, "cluster-key-injected", "", "Injected cluster key to enable pod move across cluster")
 	fs.IntVar(&s.Port, "port", s.Port, "The port that kubeturbo's http service runs on.")
 	fs.StringVar(&s.Address, "ip", s.Address, "the ip address that kubeturbo's http service runs on.")
 	// TODO: The flagset that is included by vendoring k8s uses the same names i.e. "master" and "kubeconfig".
--- a/cmd/kubeturbo/app/kubeturbo_builder.go
+++ b/cmd/kubeturbo/app/kubeturbo_builder.go
@@ -64,6 +64,7 @@ const (
 	DefaultDiscoverySampleIntervalSec = 60
 	DefaultGCIntervalMin              = 10
 	DefaultReadinessRetryThreshold    = 60
+	DefaultEnablePodClusterMoves      = false
 )
 
 var (
@@ -134,6 +135,10 @@ type VMTServer struct {
 	// The Openshift SCC list allowed for action execution
 	sccSupport []string
 
+	// Allow pod move across cluster
+	// The default is false
+	enablePodClusterMoves bool
+
 	// Force the use of self-signed certificates.
 	// The default is true.
 	ForceSelfSignedCerts bool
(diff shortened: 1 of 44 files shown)


pkg/discovery/dtofactory/cluster_dto_builder.go
```diff
@@ -14,29 +14,37 @@ import (
...
 	}
 }
 
 // GetClusterKey constructs the commodity key sold by the cluster entity, by adding a prefix to the cluster id.
 // Use of a prefix is to distinguish the same key used by the node entities in the cluster
-func GetClusterKey(clusterId string) string {
+func GetClusterKey(clusterId string, enablePodClusterMoves bool) string {
+	if enablePodClusterMoves {
+		return ClusterFreeKey
+	}
```
chlam4, line 47: I'm not sure if we should allow customization of the key between cluster and namespace entities.  Let's discuss.  I think we want such a customization between pods and nodes/VMs (so pods can move across VMs in different clusters).

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -134,6 +135,10 @@ type VMTServer struct {
 	// The Openshift SCC list allowed for action execution
 	sccSupport []string
 
+	// Allow pod move across cluster
+	// The default is false
+	enablePodClusterMoves bool
```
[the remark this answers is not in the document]
libai98: That is a good suggestion! We can discuss on this in the standup meeting.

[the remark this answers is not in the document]
libai98: code modified based on our discussion.

pkg/discovery/dtofactory/cluster_dto_builder.go
libai98: Code modified based on our discussion. Thanks!

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -221,7 +228,7 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
 		if isResizeable, exists := resizableCommodities[commSold.GetCommodityType()]; exists {
 			commSold.Resizable = &isResizeable
 		}
-	}
```
chlam4, line 224: Accidentally added?

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -239,8 +246,16 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
 	}
 
 	// Cluster commodity.
+	var clusterCommKey string
+	if len(strings.TrimSpace(builder.clusterKeyInjected)) != 0 {
+		clusterCommKey = builder.clusterKeyInjected
+		glog.V(5).Infof("Injecting cluster key for Node %s with key : %s", node.Name, clusterCommKey)
```
chlam4, line 252: I think @ading1977 said we don't have V5.  Maybe changing it to V4?

pkg/discovery/dtofactory/node_entity_dto_builder.go
```diff
@@ -239,8 +246,16 @@ func (builder *nodeEntityDTOBuilder) getNodeCommoditiesSold(node *api.Node, clus
...
+	} else {
+		clusterCommKey = clusterId
+		glog.V(5).Infof("Adding cluster key for Node %s with key : %s", node.Name, clusterCommKey)
```
chlam4, line 255: Similar comment on the V5.

pkg/discovery/dtofactory/pod_entity_dto_builder.go
```diff
@@ -423,7 +430,16 @@ func (builder *podEntityDTOBuilder) getPodCommoditiesBought(
 	if err != nil {
 		glog.Errorf("Failed to get %s used for current Kubernetes Cluster %s", metrics.Cluster, clusterInfo)
 	} else {
-		clusterCommodityKey, ok := clusterInfo.GetValue().(string)
+		var clusterCommodityKey string
+		var ok bool
+		if len(strings.TrimSpace(builder.clusterKeyInjected)) != 0 {
```
chlam4, line 435: This is the only place using `TrimSpace`.  We should probably make it consistent across all places.  Perhaps we do the trimming at the very top and propagate only the trimmed version down to every place?

chlam4 (commented):
Thanks @libai98!  Looks good to me other than a few nit comments.  Shall we test it out before merging?

pkg/discovery/dtofactory/node_entity_dto_builder.go
libai98: Yes, fixed.  :)  Thanks!

pkg/discovery/dtofactory/pod_entity_dto_builder.go
libai98: There is a long chain for the injected cluster key to be passed down from the command arg to pod and node build.
Even I do not assigned a value, the default empty string will be passed along. I am trying not to do the checking and trimming for empty string until this key is used for creating cluster access commodity in pod and node DTO builder.

libai98:
> Thanks @libai98! Looks good to me other than a few nit comments. Shall we test it out before merging?

Yes, I will try to test that out before merging. Thanks for all the help on this!

pkg/discovery/dtofactory/node_entity_dto_builder.go
libai98: Yes.

pkg/discovery/dtofactory/node_entity_dto_builder.go
libai98: Yes. Thanks for the quick review!

chlam4 (approved):

PR #752, opened 2022-11-01
[OM-92054]: create static pod group

cmedina003:
# description 
  We need a way to easily identify all mirror pods within a cluster in order to identify the static pods.
Issue: https://vmturbo.atlassian.net/browse/OM-92054

# What Changed
- created mirror pod group for display in the UI
- renamed the staticPodToDaemonMap to mirrorPodToDaemonMap

# Testing
- Deploy this version of the kubeturbo operator to any cluster with static nodes.
  - if your cluster does not have any static nodes use the k8 documentation for creating a static pod https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/
  
- ensure there should be a suspend action for VMs that implement a daemon set with static pods
- ensure that the group for mirror pods exists
Screen Shot 2022-11-01 at 4 36 02 PM
Screen Shot 2022-11-01 at 4 45 04 PM

### DIFF
--- a/pkg/discovery/worker/group_discovery_worker.go
+++ b/pkg/discovery/worker/group_discovery_worker.go
@@ -196,7 +196,7 @@ func (worker *k8sEntityGroupDiscoveryWorker) buildMirrorPodGroup(mirrorPodUids [
 	groupDTO, err := group.StaticRegularGroup(fmt.Sprintf("Mirror-Pods-%s", worker.targetId)).
 		OfType(proto.EntityDTO_CONTAINER_POD).
 		WithEntities(mirrorPodUids).
-		WithDisplayName("Mirror Pods").
+		WithDisplayName("Mirror Pods " + worker.targetId).
 		Build()
 
 	if err != nil {
--- a/pkg/discovery/worker/group_discovery_worker.go
+++ b/pkg/discovery/worker/group_discovery_worker.go
@@ -196,7 +196,7 @@ func (worker *k8sEntityGroupDiscoveryWorker) buildMirrorPodGroup(mirrorPodUids [
 	groupDTO, err := group.StaticRegularGroup(fmt.Sprintf("Mirror-Pods-%s", worker.targetId)).
 		OfType(proto.EntityDTO_CONTAINER_POD).
 		WithEntities(mirrorPodUids).
-		WithDisplayName(fmt.Sprintf("All Mirror Pods of [%s]", worker.targetId)).
+		WithDisplayName("Mirror Pods").
 		Build()
 
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/discovery/worker/group_discovery_worker.go
```diff
@@ -186,6 +188,24 @@ func (worker *k8sEntityGroupDiscoveryWorker) buildPodsWithVolumesGroup(podsWithV
 	return groupsDTOs
 }
 
+func (worker *k8sEntityGroupDiscoveryWorker) buildMirrorPodGroup(mirrorPodUids []string) []*proto.GroupDTO {
+	if len(mirrorPodUids) <= 0 {
+		return []*proto.GroupDTO{}
```
ading1977, line 193: We should return `nil` instead of an empty list here.

pkg/discovery/worker/group_discovery_worker.go
```diff
@@ -186,6 +188,24 @@ func (worker *k8sEntityGroupDiscoveryWorker) buildPodsWithVolumesGroup(podsWithV
...
+
+	groupDTO, err := group.StaticRegularGroup(fmt.Sprintf("Mirror-Pods-%s", worker.targetId)).
+		OfType(proto.EntityDTO_CONTAINER_POD).
+		WithEntities(mirrorPodUids).
+		WithDisplayName("Mirror Pods").
+		Build()
+
+	if err != nil {
+		glog.Errorf("Error creating group of mirror pods for target [%s]. Error: %v", worker.targetId, err)
+	}
```
ading1977, line 204: We should return `nil` here.

cmedina003: good catch

pkg/discovery/worker/group_discovery_worker.go
```diff
@@ -196,7 +196,7 @@ func (worker *k8sEntityGroupDiscoveryWorker) buildMirrorPodGroup(mirrorPodUids [
 	groupDTO, err := group.StaticRegularGroup(fmt.Sprintf("Mirror-Pods-%s", worker.targetId)).
 		OfType(proto.EntityDTO_CONTAINER_POD).
 		WithEntities(mirrorPodUids).
-		WithDisplayName("Mirror Pods").
+		WithDisplayName("Mirror Pods " + worker.targetId).
```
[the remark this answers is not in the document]
cmedina003: I think we are going to review that for all the group names: https://vmturbo.atlassian.net/browse/OM-92330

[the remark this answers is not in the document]
ading1977: OK, let's review that in that ticket.

ading1977 (approved):

PR #756, opened 2022-11-06
OM-92056 Integration test for static pod analysis settings

NiraliKotak:
References
https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/

Background:
Created static pod by adding the staic pod yaml like in [link](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) to the nodes while creating kind cluster as part of test setup.
With [PR](https://github.com/turbonomic/kubeturbo/pull/743) kubeturbo will identify static pod and mark daemon as true if all nodes in the nodepool have static pod.

Testing:
Added test cases 

- Multiple nodepools having static pods in all nodes. Expedted dameon:true for the pod and 1 mirror group 
- One nodepool with multiple nodes, all have static pods Expected dameon:true and 1 mirror group
- Nodepool with some nodes not having static pos Expected dameon:false and 1 mirror group

### DIFF
--- a/test/integration/discovery.go
+++ b/test/integration/discovery.go
@@ -210,7 +210,7 @@ var _ = Describe("Discover Cluster", func() {
 
 		It("NotReady nodes should be listed in the NotReady node group", func() {
 
-			notReadyGroup := getNotReadyNodeGroup(groups)
+			notReadyGroup := getGroup(groups, "NotReady-Nodes-")
 			if notReadyGroup == nil {
 				framework.Failf("Could not find NotReady Node group")
 			}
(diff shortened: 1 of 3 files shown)


test/integration/discovery.go
```diff
@@ -412,6 +412,125 @@ var _ = Describe("Discover Cluster", func() {
 		})
 	})
 
+	Describe("test mirror pods", func() {
+		testName := "discovery-integration-test"
+		mirrorpod_prefix := "static-web"
+		var entities []*proto.EntityDTO = nil
+		//var groups []*proto.GroupDTO = nil
```
cmedina003, line 419: you could delete this line.

NiraliKotak: Done

ading1977 (approved):

pankajbatra1 (approved):

PR #757, opened 2022-11-08
[OM-90969]: Updated helm chart config maps to support new feature gates structure.

amd-ibm:
## Intent
Update helm chart and operator to support new feature gates structure.

## Background
We recently changed how feature gates are structured in the configuration which made the un-configurable via helm or the operator. This aims to fix that.

## Testing
Tested against appliance with valid and invalid configurations.

When a syntax occurs reading the configuration, we will now produce an error similar to the following: 
```
F1108 10:03:13.589145   77578 kubeturbo_builder.go:358] Failed to generate correct TAP config: error processing configuration: invalid character 'T' looking for beginning of value
```

When we fail to unmarshal the configuration, we will not produce an error with the following format: 
```
F1108 10:03:39.703458   77672 kubeturbo_builder.go:358] Failed to generate correct TAP config: error processing configuration property 'featureGates':json: cannot unmarshal string into Go struct field K8sTAPServiceSpec.featureGates of type bool
```

When an unexpected feature gate is provided, kubeturbo will crash with the following error (this logic remains unchanged): 
```
F1108 10:16:37.840194   79450 kubeturbo_builder.go:364] Invalid Feature Gates: unrecognized feature gate: InvalidFeatureGate
```

### DIFF
--- a/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
+++ b/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
@@ -73,6 +73,16 @@ HANodeConfig:
 #   namespace: ""
 #   workloadController: ""
 
+# The featureGates property defines a map of string-to-boolean values that provides users with a mechanism 
+# to enable # features that are disabled by default. For a list of supported feature gates, see 
+# https://github.com/turbonomic/kubeturbo/blob/master/pkg/features/features.go
+# featureGates:
+#   PersistentVolumes: true
+#   ThrottlingMetrics: false
+#   HonorAzLabelPvAffinity: true
+#   GitopsApps: false
+#   GoMemLimit: true
+
 args:
   # logging level
   logginglevel: 2
--- a/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
+++ b/deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
@@ -74,14 +74,14 @@ HANodeConfig:
 #   workloadController: ""
 
 # The featureGates property defines a map of string-to-boolean values that provides users with a mechanism 
-# to enable # features that are disabled by default. For a list of supported feature gates, see 
+# to enable/disable features. For a list of supported feature gates, see 
 # https://github.com/turbonomic/kubeturbo/blob/master/pkg/features/features.go
 # featureGates:
-#   PersistentVolumes: true
-#   ThrottlingMetrics: false
-#   HonorAzLabelPvAffinity: true
-#   GitopsApps: false
-#   GoMemLimit: true
+#   PersistentVolumes: true (default: true)
+#   ThrottlingMetrics: false (default: true)
+#   HonorAzLabelPvAffinity: true (default: false)
+#   GitopsApps: false (default: true)
+#   GoMemLimit: true (default: false)
 
 args:
   # logging level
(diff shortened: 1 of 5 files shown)


pankajbatra1 (approved):

deploy/kubeturbo/templates/configmap.yaml
```diff
@@ -20,8 +20,12 @@ data:
...
       {{- if .Values.featureGates }}
       "featureGates": {
-        "disabledFeatures": [{{ .Values.featureGates.disabledFeatures }}]
-      }
+        {{$first := true}}
+        {{- range $key, $val := .Values.featureGates }}
+        {{- if $first}}{{$first = false}}{{else}},{{-end}}
+        {{ $key | quote }}: {{ $val }}
+        {{- end }}
+      },
```
ading1977, line 28: Have we considered something like:
`{{  .Values.featureGates | toJson }}` ?

See: [toJson](https://helm.sh/docs/chart_template_guide/function_list/#tojson-musttojson)

deploy/kubeturbo/values.yaml
```diff
@@ -73,6 +73,16 @@ HANodeConfig:
...
 
+# The featureGates property defines a map of string-to-boolean values that provides users with a mechanism 
+# to enable # features that are disabled by default. For a list of supported feature gates, see 
+# https://github.com/turbonomic/kubeturbo/blob/master/pkg/features/features.go
+# featureGates:
+#   PersistentVolumes: true
+#   ThrottlingMetrics: false
+#   HonorAzLabelPvAffinity: true
+#   GitopsApps: false
+#   GoMemLimit: true
```
ading1977, line 84: * Shall we specify the current default values to these features?
* For the following:
  > provides users with a mechanism to enable features that are disabled by default
  
  it is also true that we can disable features that are enabled by default ... So may be just say provides users with a mechanism to enable or disable a feature?

amd-ibm: Sure, I can add the default values.

amd-ibm: Updated description and added default values.

deploy/kubeturbo/templates/configmap.yaml
amd-ibm: Changed to this. Thanks for the recommendation.

deploy/kubeturbo-operator/helm-charts/kubeturbo/values.yaml
```diff
@@ -73,6 +73,16 @@ HANodeConfig:
 #   namespace: ""
 #   workloadController: ""
 
+# The featureGates property defines a map of string-to-boolean values that provides users with a mechanism 
+# to enable # features that are disabled by default. For a list of supported feature gates, see
```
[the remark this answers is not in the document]
amd-ibm: Fixed.

ading1977 (approved):

pankajbatra1 (approved):

amd-ibm:
> 

@ading1977 , if the user were to pass a non-boolean value, kubeturbo will crash due to an error marshaling the data. When that happens, the kubeturbo logs will show an error such as the following: 

```
F1108 10:03:39.703458   77672 kubeturbo_builder.go:358] Failed to generate correct TAP config: error processing configuration property 'featureGates':json: cannot unmarshal string into Go struct field K8sTAPServiceSpec.featureGates of type bool
```

PR #758, opened 2022-11-10
[OM-87493]: Updated preActionCheck to fail actions when the destination node is not in a Ready state

amd-ibm:
## Intent
Updated preActionCheck to fail actions when the destination node is not in a Ready state.
Fixed warning logs for nodes with other potentially bad conditions that were displaying false-positives.

## Background
During the preActionCheck for pod moves we incorrectly generating warning logs for DiskPressure, MemoryPressure, etc. For example: 

```
W1006 21:20:29.404068       1 rescheduler.go:153] Move action: pod[default/pod-move-auto-20221006210832-6c95b788cb-tbwkm]'s new host(okd-auto-node-2) in bad condition: MemoryPressure
W1006 21:20:29.404081       1 rescheduler.go:153] Move action: pod[default/pod-move-auto-20221006210832-6c95b788cb-tbwkm]'s new host(okd-auto-node-2) in bad condition: DiskPressure
W1006 21:20:29.404090       1 rescheduler.go:153] Move action: pod[default/pod-move-auto-20221006210832-6c95b788cb-tbwkm]'s new host(okd-auto-node-2) in bad condition: PIDPressure
```

Looking at the way the logs are generated and what a health node's conditions look like, it was clear that we were warning when we shouldn't be in most cases. Most node conditions suggest a bad state when the value is `True`. For `Ready` status, on the other hand, `True` indicates a good state (`False` or `Unknown` indicate an unhealthy state in this case).

Here's an example of a health node's conditions:

Screen Shot 2022-11-10 at 12 12 54 PM

## Testing
Ran kubeturbo locally, proxied to my appliance. 
Flipped the boolean flags in the logic to mimic various scenarios. 
Ran the build and unit tests.

### DIFF
--- a/pkg/action/executor/rescheduler.go
+++ b/pkg/action/executor/rescheduler.go
@@ -151,7 +151,7 @@ func (r *ReScheduler) preActionCheck(pod *api.Pod, node *api.Node) error {
 	for _, cond := range conditions {
 		if cond.Type == api.NodeReady {
 			// If the destination node is NOT in a Ready state, return an error to fail the action
-			if cond.Status == api.ConditionFalse {
+			if cond.Status != api.ConditionTrue {
 				return fmt.Errorf("Move action: pod[%v]'s new host (%v) is in bad condition: %v",
 					fullName, node.Name, cond.Message)
 			}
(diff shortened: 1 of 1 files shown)


pkg/action/executor/rescheduler.go
```diff
@@ -149,13 +149,19 @@ func (r *ReScheduler) preActionCheck(pod *api.Pod, node *api.Node) error {
 	}
 
 	for _, cond := range conditions {
-		if cond.Status != api.ConditionTrue {
-			glog.Warningf("Move action: pod[%v]'s new host(%v) in bad condition: %v", fullName, node.Name, cond.Type)
+		if cond.Type == api.NodeReady {
+			// If the destination node is NOT in a Ready state, return an error to fail the action
+			if cond.Status != api.ConditionTrue {
+				return fmt.Errorf("Move action: pod[%v]'s new host (%v) is in bad condition: %v",
```
ading1977, line 155: We should just say that the new host is NotReady

ading1977 (approved):

pkg/action/executor/rescheduler.go
amd-ibm: Updated the error message.

PR #768, opened 2022-11-22
update setting controllable flag with additional mirror pod logic

cmedina003:
# Description

During Node Provisioning, action for starting/provisioning of mirror pods were not being created when they should be treated the same as daemon pods. Prior to this MR, if the mirror pod was detected the logic would set the pod as `controllable: false`. When the mirror pod is representing a static pod that is being modeled as a daemon we will now set `controllable: true`, which is what we do for daemon pods.

Screen Shot 2022-11-21 at 4 30 00 PM
Screen Shot 2022-11-22 at 11 42 35 AM

# Testing
 Run a plan that adds 110 x (number of nodes) on a cluster and when you see a provision action for a node, then you should also see a provision action for the static pod.


ading1977:
@cmedina003 , we should only change the `controllable` flag for container pods. For containers and applications, they should not be controlled, i.e., we do NOT want resize or horizontal scale for app and containers that sit on static pods.

ading1977 (approved):
LGTM.

PR #769, opened 2022-11-25
[OM-93059] Improve the calculation of number of items per query

ading1977:
## Goal
This PR improves the calculation of number of items per query based on the research done by @chlam4 . According to the investigation, the ideal number of items per query depends on both memory limit of `kubeturbo`, and also the number of pods running in the cluster.

## Implementation
Currently, the items per query is calculated at `kubeturbo` startup time, and never changes afterwards. This needs to be changed because:
* The number of pods in the cluster changes over time.
* The memory limit of `kubeturbo` may also change over time when in-place pod resource update is enabled

As a result, the following logic is implemented:
* During `kubeturbo` startup time, only do sanity check of the command line argument `--items-per-list-query`:
  * If the value is negative, set it to `DefaultItemsPerGBMemory`
  * If the value is too low (i.e., below `DefaultItemsPerGBMemory`), set it to `DefaultItemsPerGBMemory`
* At the start of each main discovery `K8sDiscoveryClient.Discover`:
  * Set `GoMemLimit` based on the current memory limit 
* During `ClusterProcessor.DiscoverCluster`:
  * Calculate items per query based on memory limit and the number pods in the cluster
  * The result will be passed to `ControllerProcessor` which performs paginated query through `ClusterInfoScraper.GetResourcesPaginated` using the result as the page size 

## Test
* Deployed `kubeturbo` dev build in GKE, OCP 4.8 in AWS without memory limit set
  * In GKE:
    `I1125 15:59:36.437826       1 cluster_processor.go:228] Set items per list query to 5678 based on memory limit of 2.84 GB and pod count of 100.`
  * In OCP 4.8 in AWS:
    `I1125 15:48:13.028831       1 cluster_processor.go:228] Set items per list query to 126557 based on memory limit of 29.95 GB and pod count of 997.`
 
* Deployed `kubeturbo` dev build in OCP 4.8 in AWS with memory limit set to a low number of 1.2GB
 
 ```
I1129 19:57:47.159190       1 cluster_processor.go:193] Discovering cluster with 7 nodes and 1001 pods.
W1129 19:57:47.160248       1 cluster_processor.go:200] Cannot calculate items per list API call: the memory limit of kubeturbo 1.20 GB is too low to calculate a reasonable number of items per list API call. Kubeturbo may run into OOM. Consider increasing the memory limit of kubeturbo to at least 2.93 GB.
I1129 19:57:47.160270       1 cluster_processor.go:201] Set items per list API call to the default value of 5000.
  ```
* Deployed `kubeturbo` dev build in AKS. The error in the log is irrelevant to this issue, and has been address in [this](https://github.com/KimMachineGun/automemlimit/pull/2) PR.

```
W1129 19:57:45.893731       1 cluster_processor.go:200] Cannot calculate items per list API call: error retrieving memory limit (0): open /sys/fs/cgroup/memory/memory.memsw.usage_in_bytes: no such file or directory.
I1129 19:57:45.893831       1 cluster_processor.go:201] Set items per list API call to the default value of 5000.
```

### DIFF
--- a/pkg/discovery/processor/cluster_processor.go
+++ b/pkg/discovery/processor/cluster_processor.go
@@ -19,10 +19,8 @@ import (
 const (
 	// DefaultItemsPerGBMemory defines number of items to retrieve for each GB of memory
 	DefaultItemsPerGBMemory                     = 5000
-	DefaultMemoryLimitInGB                      = 8
 	DefaultAutoMemLimitPct                      = 0.9
 	DefaultGenericMetricSizePerThousandPodsInGB = 0.24
-	DefaultNumberOfPodsInThousands              = 1
 	DefaultCadvisorMetricSizePerPodInGB         = 0.004
 	DefaultMaxPodsPerNode                       = 250
 	DefaultExtraPerNodeUsageInGB                = 0.2
@@ -32,39 +30,6 @@ const (
 var (
 	workers       = 10
 	totalWaitTime = 60 * time.Second
-	// DefaultItemsPerListQuery defines default number of items that should be requested in each
-	// workload controller list API to ensure no OOMs occur.
-	//
-	// This value is calculated using the following expression:
-	//
-	// (kubeturbo_mem_limit_gb * default_automemlimit_pct
-	//    - generic_metric_size_gb_per_thousand_pods * number_of_pods_in_thousands_in_cluster
-	//    - (cadvisor_metric_size_gb_per_pod * max_pods_per_node + extra_per_node_usage_gb)
-	//    - extra_cluster_wide_usage_gb) * items_per_gb
-	//
-	// In the absence of any MEMLIMIT set, we assume the following default values
-	// to calculate the default items per query:
-	//   * mem_limit_gb: 8
-	//   * default_automemlimit_pct: 0.9 (percentage of memory limit that can be controlled by Go Runtime)
-	//   * generic_metric_size_gb_per_thousand_pods: 0.24
-	//   * number_of_pods_in_thousands_in_cluster: 1 (1000 pods)
-	//   * cadvisor_metric_size_gb_per_pod: 0.004
-	//   * max_pods_per_node: 250
-	//   * extra_per_node_usage_gb: 0.2
-	//   * extra_cluster_wide_usage_gb: 0.2
-	//   * items_per_gb: 5000
-	//
-	// which results in the value of:
-	//   (8 * 0.9 - 0.24 * 1 - (0.004 * 250 + 0.2) - 0.2) *  5000 = 27800
-	//
-	// This default is unlikely to be used on Linux based system as Cgroup limit will always
-	// be available. If the container limit is not set, Cgroup limit will be set to nodes limit.
-	//
-	// This value will be used for non-linux systems, e.g., kubeturbo local run on MacOS.
-	DefaultItemsPerListQuery = (DefaultMemoryLimitInGB*DefaultAutoMemLimitPct -
-		DefaultGenericMetricSizePerThousandPodsInGB*DefaultNumberOfPodsInThousands -
-		(DefaultCadvisorMetricSizePerPodInGB*DefaultMaxPodsPerNode + DefaultExtraPerNodeUsageInGB) -
-		DefaultExtraClusterWideUsageInGB) * DefaultItemsPerGBMemory
 )
 
 // ClusterProcessor defines top level object that will connect to the Kubernetes cluster and all the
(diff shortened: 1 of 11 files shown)


pkg/discovery/processor/cluster_processor.go
```diff
@@ -2,31 +2,85 @@ package processor
...
+	//   * max_pods_per_node: 250
+	//   * extra_per_node_usage_gb: 0.2
+	//   * extra_cluster_wide_usage_gb: 0.2
+	//   * items_per_gb: 5000
+	//
+	// which results in the value of:
+	//   (8 * 0.9 - 0.24 * 1 - (0.004 * 250 + 0.2) - 0.2) *  5000 = 27800
+	//
+	// This default is unlikely to be used on Linux based system as Cgroup limit will always
+	// be available. If the container limit is not set, Cgroup limit will be set to nodes limit.
```
chlam4, line 61: I wonder if it'd be too aggressive to use the node memory limit to compute the items per query.  Perhaps we should use the half of the node limit instead, assuming there are others on the node too which could assume a significant amount of memory.  We can consider doing that in a subsequent PR, and leave the current one as is.

pkg/discovery/processor/cluster_processor.go
```diff
@@ -198,3 +271,52 @@ func (p *ClusterProcessor) DiscoverCluster() (*repository.ClusterSummary, error)
...
+	limit, err := memlimit.FromCgroup()
+	if err != nil {
+		// This is very unlikely because in absense of any limit set in container resources
+		// we will get the cgroup limit as the nodes available/usable memory limit
+		return 0, 0, fmt.Errorf("error retrieving memory limit (%v): %v", limit, err)
+	}
+	if limit == 0 {
+		// This is very unlikely because in absense of any limit set in container resources
+		// we will get the cgroup limit as the nodes available/usable memory limit
+		return 0, 0, fmt.Errorf("limit found set to zero (0")
```
chlam4, line 309: Nit: missing a closing paranthesis inside quotes?

pkg/discovery/processor/cluster_processor.go
```diff
@@ -198,3 +271,52 @@ func (p *ClusterProcessor) DiscoverCluster() (*repository.ClusterSummary, error)
...
+	}
+	podsInThousands := float64(podCount) / 1000
+	limitInGB := nodeUtil.Base2BytesToGigabytes(float64(limit))
+	availMem := limitInGB*DefaultAutoMemLimitPct -
+		DefaultGenericMetricSizePerThousandPodsInGB*podsInThousands -
+		(DefaultCadvisorMetricSizePerPodInGB*DefaultMaxPodsPerNode + DefaultExtraPerNodeUsageInGB) -
+		DefaultExtraClusterWideUsageInGB
+	if availMem < 1.0 {
+		return 0, 0, fmt.Errorf("unable to calculate reasonable items per list query based on "+
```
chlam4, line 318: While I agree that it will be too rough an estimate when `availMem` becomes too small, it's still probably better than setting it to the default (which is 27800 and is much bigger than 5000).  I'd suggest we consider the followings:
- I think we can set a minimum, e.g., 2000 items per query (instead of the default which is 27800).
- If `availMem` is larger than 0.4 for example, we still use `DefaultItemsPerGBMemory * availMem`.
- If smaller, then we use the minimum instead of the default.  In the log message, we should say this represents not enough memory allocated for kubeturbo and one should increase it.  Furthermore, we can estimate how many round trips it would take and warn the user that it could take a long time to discover and it could also run into OOM.

chlam4 (commented):
Nice work @ading1977!  A couple of comments for your consideration.

pkg/discovery/processor/cluster_processor.go
ading1977: How about further simplify this, and get rid of the `DefaultItemsPerListQuery` of 27800, and just make the default to 5000?

* If `availMem` is smaller than 1, use  `DefaultItemsPerGBMemory` (5000), and show logs to suggest increase the memory limit.
* Otherwise, use `DefaultItemsPerGBMemory * availMem`

ading1977: This has been agreed.

pkg/discovery/processor/cluster_processor.go
ading1977: From @ading1977 

Regarding halve the node limit, the challenge is that it is not straightforward to tell if the limit is actually a node limit, or a user specified container limit …

From @chlam4 

I think regardless it’s user specified or inherited from the node, it presents a higher risk if the container limit is too close to the node’s (unless its request is also as high).  This may sound like crazy, but a formula like the following to adjust the container limit could work well in practice:
`effective_container_limit = container_limit * ( 1 - 2 ^ ((container_limit - container_request) / node_limit) )`
As this is definitely crazy and probably not easy to explain, we can just leave as is and see what happens in practice.  Perhaps the 2x buffer is enough to minimize any OOMs.

chlam4 (approved):

PR #775, opened 2022-12-02
[OM-93675]: Updated turbo-go-sdk vendor with new proto bindings 

pallavidn:
Update the turbo-go-sdk vendor due to new go bindings as per [PR](https://github.com/turbonomic/turbo-go-sdk/pull/144)


chlam4 (approved):

ading1977:
Not valid anymore.

PR #777, opened 2022-12-08
[OM-93635] Provide a feature flag to ignore processing affinities

irfanurrehman:
**Intent**
This is to provide a temporary option to ignore processing affinities during discovery of a k8s cluster.
This will help us validate our analysis of the root cause of https://vmturbo.atlassian.net/browse/OM-93635 and help progress the POV ahead.


irfanurrehman:
Sanity test on the above shows all entities discovered are same with and without this flag. All actions are also same, and infact all the pod moves are also same if the cluster does not have workloads defining any `affinity` or `antiaffinity` rules.

The image is available as `irfdocker/kubeturbo:ignore-affinities`

chlam4 (approved):
Thank you @irfanurrehman!

ading1977 (approved):

pankajbatra1 (approved):

PR #781, opened 2023-01-03
[OM-88273]: GitOps Configuration Overrides via CustomResource

amd-ibm:
## Intent
Integrate the [new GitOps CustomResource](https://github.com/turbonomic/turbo-gitops) into kubeturbo

## Background
Recently, support was added for applications controlled via GitOps (ArgoCD). The configuration for this functionality is provided as runtime parameters, and limits users to a single git account (only one set of credentials can be provided), a single commit mode type for all applications, etc. The changes here, integrate a new GitOps CustomResource into kubeturbo that allows users to override these default configurations, such as overriding the commit mode or providing a separate set of credentials.

NOTE: The GitOps CustomResources are namespace-scoped, so overrides will only apply to apps within the same namespace as  the resource is deployed.

## Testing
Spun up multiple GitOps apps in appliance, and tested kubeturbo locally against remote appliance. Tests included the following configuration scenarios:
* tested custom resource with selector
* tested custom resource with whitelist
* tested custom resource with both selector and whitelist
* tested custom resource without either selector or whitelist
* tested custom resource with credentials

Additionally, added and updated unit tests.


pankajbatra1 (approved):

pkg/cluster/cluster_info_scraper.go
```diff
@@ -66,17 +69,21 @@ type ClusterScraper struct {
 	DynamicClient           dynamic.Interface
 	ControllerRuntimeClient runtimeclient.Client
 	cache                   turbostore.ITurboCache
+	GitOpsConfigCache       *map[string][]*gitopsv1alpha1.Configuration
```
irfanurrehman, line 72: maps in go are already pointers and can be exactly used as pointers. You would almost never need a pointer to a map.

pkg/action/executor/k8s_controller_updater.go
```diff
@@ -74,11 +74,13 @@ func newK8sControllerUpdater(clusterScraper *cluster.ClusterScraper, ormClient *
...
-			managerApp:   managerApp,
-			gitConfig:    gitConfig,
-			k8sClusterId: clusterId,
+			name:                  kind,
+			ormClient:             ormClient,
+			managerApp:            managerApp,
+			gitConfig:             gitConfig,
+			k8sClusterId:          clusterId,
+			gitOpsConfigCache:     *clusterScraper.GitOpsConfigCache,
+			gitOpsConfigCacheLock: &clusterScraper.GitOpsConfigCacheLock,
```
irfanurrehman, line 83: I dont generally define a pointer variable until there is a need for one. I don't understand the need here.

pkg/discovery/processor/cluster_processor.go
```diff
@@ -229,6 +229,14 @@ func (p *ClusterProcessor) DiscoverCluster() (*repository.ClusterSummary, error)
 	// Discover Turbo Policies
 	NewTurboPolicyProcessor(p.clusterInfoScraper, kubeCluster).ProcessTurboPolicies()
 
+	// Discover GitOps configuration overrides
+	NewGitOpsConfigProcessor(p.clusterInfoScraper, kubeCluster).ProcessGitOpsConfigs()
+
+	// Update the GitOps configuration override cache
+	if clusterScraper, ok := p.clusterInfoScraper.(*cluster.ClusterScraper); ok {
+		clusterScraper.UpdateGitOpsConfigCache(kubeCluster.GitOpsConfigurations)
```
irfanurrehman, line 237: This is an odd usage. We are sort of duplicating the cache data in memory. Effectively storing the copy in `kubeCluster` and then in `clusterScraper.GitopsConfigCache`. We need only one copy. I would simply call `GetAllGitOpsConfigurations()` here and pass the data on. We don't need the copy in `kubeCluster`. Or do we, I might be missing something.. :-)

pkg/discovery/processor/cluster_processor.go
```diff
@@ -229,6 +229,14 @@ func (p *ClusterProcessor) DiscoverCluster() (*repository.ClusterSummary, error)
 	// Discover Turbo Policies
 	NewTurboPolicyProcessor(p.clusterInfoScraper, kubeCluster).ProcessTurboPolicies()
 
+	// Discover GitOps configuration overrides
+	NewGitOpsConfigProcessor(p.clusterInfoScraper, kubeCluster).ProcessGitOpsConfigs()
+
+	// Update the GitOps configuration override cache
+	if clusterScraper, ok := p.clusterInfoScraper.(*cluster.ClusterScraper); ok {
```
irfanurrehman, line 236: Why do we need the type check here?  We already defined and implemented the `GetAllGitOpsConfigurations()` interface.

pkg/discovery/processor/gitops_config_processor.go
```diff
@@ -0,0 +1,30 @@
+package processor
```
irfanurrehman, line 1: I am not sure if we really need this processor.

irfanurrehman (changes requested):
Thanks @amd-ibm for doing this.
Few changes suggested.

pkg/action/executor/k8s_controller.go
```diff
@@ -251,3 +258,76 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
+}
+
+// Returns a flag indicating if there exists a GitOps configuration override for the supplied application
+func isGitOpsConfigOverridden(config *gitopsv1alpha1.Configuration, objName string) bool {
+	return isAppSelectorMatch(config.Selector, objName) || isAppInWhitelist(config.Whitelist, objName)
+}
+
+// Returns the credentials to be used for a GitOps operation.
+func getGitOpsCredentials(overrideConfig *gitopsv1alpha1.Configuration, defaultConfig gitops.GitConfig) (string, string, string, string) {
+	secretNamespace := defaultConfig.GitSecretNamespace
```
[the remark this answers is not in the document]
amd-ibm: Updated

pkg/action/executor/k8s_controller.go
```diff
@@ -251,3 +258,76 @@ func (c *parentController) shouldSkipOperator(controller *unstructured.Unstructu
...
+		glog.V(4).Infof("Overriding GitOps credentials to use username [%v] instead of the default.", username)
+	}
+	return email, username, secretName, secretNamespace
+}
+
+// Returns the GitOps configuration for the supplied application. If an override exists, it will return it. Otherwise,
+// it will return the default configuration supplied that was supplied as a runtime parameter.
+func ( c *parentController) GetGitOpsConfig(obj *unstructured.Unstructured) gitops.GitConfig {
+	appName := c.managerApp.Name
+	glog.V(2).Infof("Checking for GitOps configuration override for %v...", appName)
```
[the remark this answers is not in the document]
amd-ibm: Updated the log level

pkg/cluster/cluster_info_scraper.go
amd-ibm: Removed the pointer.

pkg/action/executor/k8s_controller_updater.go
amd-ibm: Are you referring to the lock here?

pkg/discovery/processor/cluster_processor.go
amd-ibm: Updated to call `UpdateGitOpsConfigCache` which in turn calls `GetAllGitOpsConfigurations`.

pkg/discovery/processor/cluster_processor.go
amd-ibm: Removed this code.

pkg/discovery/processor/gitops_config_processor.go
amd-ibm: Removed the processor.

pkg/action/executor/k8s_controller_updater.go
irfanurrehman: Ignore this plz.. :-)

irfanurrehman (approved):
Thanks @amd-ibm for the updates.
I am not sure, why I see multiple individual commits with the same commit msg. Please ensure to squash the commits before merge.

PR #797, opened 2023-02-03
[OM-95986] Allow changing image tag with certified kubeturbo-operator

ading1977:
## Background

After switching to `kubeturbo` image registry to `icr.io`, user cannot change the `kubeturbo` version using the `icr.io/cpopen/turbonomic` repository. This is a critical issue as user cannot change `kubeturbo` version to match with old server version.

## Solution

* Remove the logic to set hardcoded image digest for `kubeturbo` images
* Allow build script to set default `kubeturbo` version in `values.yaml`

With the above changes, the build script should specify a `VERSION` build argument when building `kubeturbo-operator` container image, for example:

```
docker buildx build --platform "linux/amd64,linux/arm64,linux/ppc64le,linux/s390x" --build-arg VERSION=8.8.0 --push -t icr.io/cpopen/turbonomic/kubeturbo-operator:8.8.0 .
```

The `Dockerfile` will replace the `VERSION` placeholder in the `values.yaml` file with the provided build argument.

## Test
* Create a dev build `kubeturbo-operator` container image with `VERSION=8.8.0-SNAPSHOT` , and use that to create and publish a certified kubeturbo operator bundle into the beta channel:
 
  image

* Deploy the above certified operator bundle in an Openshift v4.11 cluster, and verify that `icr.io/cpopen/turbonomic/kubeturbo:8.8.0-SNAPSHOT` instance is created and working as expected

* Add the following into the operator yaml:
```
  image:
    tag: 8.7.6
```
* Verify that `icr.io/cpopen/turbonomic/kubeturbo:8.7.6` is created and working as expected
  
  image

### DIFF
--- a/deploy/kubeturbo-operator/Dockerfile
+++ b/deploy/kubeturbo-operator/Dockerfile
@@ -1,26 +1,32 @@
 FROM quay.io/operator-framework/helm-operator:v1.25
 MAINTAINER Turbonomic <turbodeploy@turbonomic.com>
+ARG VERSION
+
+RUN echo "Building kubeturbo-operator:$VERSION"
 
 # Required OpenShift Labels
 LABEL name="Kubeturbo Operator" \
-      vendor="Turbonomic" \
-      version="8" \
-      release="6" \
+      vendor="IBM" \
+      version=$VERSION \
+      release=$VERSION \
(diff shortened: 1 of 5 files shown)


SumanthKodali999 (approved):
LGTM

PR #802, opened 2023-02-13
[OM-96836]:Update go-restful dependency to solve CVE issue

kevinwangcn:
## Background 
The `go-restful` library(version 2.9.5) is causing a security vulnerability issue. the fix will update it to version 2.16.0 which could solve the issue
https://vmturbo.atlassian.net/browse/OM-96836

## Build the custom image and scan it with  `twistlock`, no `critical` issue anymore
```
./tt images pull-and-scan --user xxx --url "https://w3twistlock.sos.ibm.com" --control-group public docker.io/kevin0204/kubeturbo:security0214
```
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	github.com/coreos/go-systemd/v22 v22.3.2 // indirect
 	github.com/docker/distribution v2.8.0+incompatible // indirect
 	github.com/docker/go-units v0.4.0 // indirect
-	github.com/emicklei/go-restful v2.9.5+incompatible // indirect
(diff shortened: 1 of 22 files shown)


go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/docker/distribution v2.8.0+incompatible // indirect
 	github.com/docker/go-units v0.4.0 // indirect
 	github.com/emicklei/go-restful v2.9.5+incompatible // indirect
```
pankajbatra1, line 35: Why do you still have this line?

PR #810, opened 2023-02-27
[OM-97758] The deploy/kubeturbo folder gets ignored by git

tian-ma:
gitignore was ignoring folder `deploy/kubeturbo` which is not correct.

Not in jira


ading1977 (approved):
Thanks for working on this. LGTM.

PR #812, opened 2023-02-28
[OM-97759] Kubeturbo helm deployment fails if turbonomicCredentialsSecretName gets commented out by user

tian-ma:
This fixes the error

```
Error: INSTALLATION FAILED: Deployment.apps "kubeturbo-..." is invalid: [spec.template.spec.volumes[1].secret.secretName: Required value, spec.template.spec.containers[0].volumeMounts[1].name: Not found: "turbonomic-credentials-volume"]
```

when the value for `turbonomicCredentialsSecretName` is not used

### DIFF
--- a/deploy/kubeturbo/templates/deployment.yaml
+++ b/deploy/kubeturbo/templates/deployment.yaml
@@ -87,7 +87,7 @@ spec:
         secret:
           defaultMode: 420
           optional: true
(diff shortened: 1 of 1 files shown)


ading1977 (approved):
LGTM.
This is only needed for helm install. For operator install, `turbonomicCredentialsSecretName` is defined in the default `values.yaml`, which cannot be removed.

pankajbatra1 (approved):

PR #815, opened 2023-03-09
[OM-94865] Affinity processing unit tests.

amd-ibm:
## Intent

Add/update unit tests for code affected by recent affinity processing improvements.

## Background

We recently merged some affinity processing improvements (https://github.com/turbonomic/kubeturbo/pull/791). This did not include updates to unit tests, as the work was split. This covers _some_ of the testing. There will be additional tests added in the near future.


pkg/discovery/dtofactory/pod_entity_dto_builder_test.go
```diff
@@ -272,3 +273,306 @@ func TestGetRegionZoneLabelCommodity(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.NotEmpty(t, commoditiesBought)
+	assert.Equal(t, 2, len(commoditiesBought))
+
+	for _, commodity := range commoditiesBought {
+		assert.Equal(t, proto.CommodityDTO_LABEL, *commodity.CommodityType)
+	}
+}
+
+func Test_getPodCommoditiesBought_PodAffinityToSpread(t *testing.T) {
```
kevinwangcn, line 432: It looks to me all the new cases with the length of the commodity as 1 are going to the same code branch. 
The `getPodCommoditiesBought` function doesn't seem to treat different kinds of affinity`(pod(node) affinity/anti-affinity)` differently.  Maybe I misunderstand something.

pkg/discovery/processor/namespace_processor_test.go
```diff
@@ -81,7 +82,14 @@ func TestProcessNamespaces(t *testing.T) {
 		KubeCluster:        ks,
 	}
 
-	namespaceProcessor.ProcessNamespaces()
+	kubeNamespaceMap := namespaceProcessor.ProcessNamespaces()
```
kevinwangcn, line 85: Does this case relate to affinity? I don't get it.

amd-ibm: It does, in a way. The `namespaceProcessor.ProcessNamespaces()` method was updated to return this map as a part of the affinity work. All I added was a few assertions that that map was generating the expected values.

pkg/discovery/dtofactory/pod_entity_dto_builder_test.go
amd-ibm: I was just trying to ensure that the commodities would be generated for each scenario.

pkg/discovery/processor/namespace_processor_test.go
kevinwangcn: Make sense to me

pkg/discovery/dtofactory/pod_entity_dto_builder_test.go
kevinwangcn: Got your point.  thanks for the clarification.

kevinwangcn (approved):

PR #821, opened 2023-03-16
[OM-94867] Unit tests for new podaffinity processing algorithm

irfanurrehman:
This implements units tests for the new algorithm merged in #791
The test cases are all that exist upstream and the idea is to ensure tht all those test cases do pass, even after all the changes that we have done to adapt and import the algorithm from k8s upstream or will do in future.

This does not include tests for nodeaffiinity and pv affinity and can be taken up as a separate task at some point later.


irfanurrehman:
@amd-ibm @pankajbatra1 @kevinwangcn Review/Approval needed. thanks!

PR #790, opened 2023-01-13
[OM-94895]:Replace busybox image with cpufreqgetter and remove the useless code

kevinwangcn:
# Intent
As we're moving all of the images to icr.io, the third-party image `busybox` which is used by `kubeturbo` to get CPU frequency from the nodes of one k8s cluster will be replaced by `cpufreqgetter` that is written in Go(only use the standard library) and will be maintained by ourselves. 

# Background
`kubeturbo` relies on `busybox` image to get CPU frequency 

# Implementation
The new `cpufreqgetter` will deal with different platforms(x86/Power/Z), and it will provide multi-arch images also.  The pod of `cpufreqgetter` will print out the final value of CPU frequency in the log. `kubeturbo` will not need to distinguish between different platforms and parse the value according to CPU arch, it will read the value from the pod log directly.

# Test Done

## X86:Rediscovery and check the kubeturbo log
```
[root@api.ocp410kev.cp.fyre.ibm.com git]# k logs kubeturbo-kw-114-59bc97479c-729wc |grep -i "CPU frequency"
W0113 17:21:58.906634       1 kubelet_client.go:339] Failed to get CPU frequency for node master0.ocp410kev.cp.fyre.ibm.com from kubelet: "https://10.22.119.197:10250/spec" was not found. Will try job based getter.
W0113 17:21:58.906641       1 kubelet_client.go:339] Failed to get CPU frequency for node master1.ocp410kev.cp.fyre.ibm.com from kubelet: "https://10.22.121.188:10250/spec" was not found. Will try job based getter.
W0113 17:21:58.907355       1 kubelet_client.go:339] Failed to get CPU frequency for node worker2.ocp410kev.cp.fyre.ibm.com from kubelet: "https://10.22.127.187:10250/spec" was not found. Will try job based getter.
W0113 17:21:58.907387       1 kubelet_client.go:339] Failed to get CPU frequency for node master2.ocp410kev.cp.fyre.ibm.com from kubelet: "https://10.22.122.188:10250/spec" was not found. Will try job based getter.
W0113 17:21:58.907446       1 kubelet_client.go:339] Failed to get CPU frequency for node worker0.ocp410kev.cp.fyre.ibm.com from kubelet: "https://10.22.123.185:10250/spec" was not found. Will try job based getter.
[...]

PR #825, opened 2023-03-21
[OM-99059] Ensure backward compatibility of --busybox-image parameter

irfanurrehman:
**Intent**
Fixes https://vmturbo.atlassian.net/browse/OM-99059

**Background**
We did leave behind `--busybox-image` parameter with the feature implemented in #790 but using the same does not work with the latest builds
This PR fixes the same


amd-ibm (approved):

kevinwangcn:
Looks good to me. Thanks for helping fix this  @irfanurrehman

kevinwangcn (approved):

PR #828, opened 2023-03-23
[OM-96144]: Adding some unit tests related to Vertical scale CRD in k…

libai98:
…ubeturbo


amd-ibm:
It looks like you didn't add tests for all of the utility functions. Was there any reason not to test them all?

Also, I think you should add additional positive and negative tests for the discovery logic. The utility tests are nice to have, but we should ensure that discovery is behaving the way we expect it to.

libai98:
The util methods I did not add unit tests are addCVSLimitSettings and addCVSRequestSettings. Those are just some common code I used for build CVS policy. Yes, it is good to have unit tests for this and discovery logic.
My unit tests here mainly focus on testing on converting resource.quantity value of CPU and memory to the unit we specified in their respective DTOs.

amd-ibm (approved):

PR #830, opened 2023-03-24
[OM-98560]: add more log information for container vertical scale pol…

libai98:
…icy discovery

### DIFF
--- a/pkg/discovery/worker/group_discovery_policy_cvs.go
+++ b/pkg/discovery/worker/group_discovery_policy_cvs.go
@@ -52,7 +52,6 @@ func resolveCVSPolicyTargets(
 	var resolvedIds []string
 	for _, target := range targets {
 		namespace := policyBinding.GetNamespace()
-		controllerRegex := target.Name
 
 		// All targets must be container in a workload controller
 		controllerRegex := target.Name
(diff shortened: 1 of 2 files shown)


pkg/discovery/worker/group_discovery_policy_cvs.go
```diff
@@ -89,24 +95,26 @@ func resolveCVSPolicyTargets(
...
+		return nil, fmt.Errorf("the ContainerVerticalScaleSpec is nil in policy binding %v", pbName)
 	}
 
 	settings := group.NewSettingsBuilder()
 
 	aggressiveness := spec.Settings.Aggressiveness
 	if aggressiveness != nil {
 		val := PercentileAggressiveness[(string(*aggressiveness))]
-		glog.V(2).Infof("Aggressiveness %f", val)
+		glog.V(4).Infof("Aggressiveness:  %f", val)
```
amd-ibm, line 110: Should we include the policy name with these to add more context? Perhaps it's unnecessary 🤷‍♂️ we don't process these in parallel, do we?

amd-ibm (approved):

pkg/discovery/worker/group_discovery_policy_cvs.go
libai98: We do not process that in parallel.

PR #838, opened 2023-04-05
[OM-99775]: Support multi-stage build and update GO to 1.20

kevinwangcn:
# Intent
To get rid of the dependence on the local build environment when building the docker image. This change will make it easier to update the Go version, there is no need to make any change to the Jenkins environment when we want to update Go to another new version. Instead, what we only need to do is simply change the builder version in the Dockerfile.  This also avoids problems caused by inconsistency of the Dev and Jenkins environments

https://docs.docker.com/build/building/multi-stage/

# Background
We are continually finding more and more security issues on the Go, which requires us to update with a newer version. That always needs to involve the DevOps team to update the Go version in the pipeline.  

# Implementation
Add a Dockerfile which supports the multi-stage build. By this means, DevOps can easily run the command `make docker-buildx` to build Kubeturbo image which also supports multi-platforms. 

# Test

1. Run the command `make docker-buildx` to build multi-arch images and push it to docker(or icr)
Could see `Kubeturbo` image is built and push to docker with multi-archs
image

3. Scan the image with `twistlock`
The report doesn't have any CVE issue on Go
[twistlock-scan-results-20230405-220642-N-UTC-0BD8E3AE.results.csv](https://github.com/turbonomic/kubeturbo/files/11165414/twistlock-scan-results-20230405-220642-N-UTC-0BD8E3AE.results.csv)

5. Deploy the image in a K8s cluster and check if the pod could run 
The pod could run with the new image and there is  no outstanding error in the log

### DIFF
--- a/build/Dockerfile.multi-archs
+++ b/build/Dockerfile.multi-archs
@@ -1,0 +1,60 @@
+# Building base image
+FROM --platform=$BUILDPLATFORM golang:1.20.3 AS builder
+ARG version
+ENV KUBETURBO_VERSION $version
+WORKDIR /workspace
+ADD . ./
+RUN make build
+
+
+FROM registry.access.redhat.com/ubi8
+MAINTAINER Turbonomic <turbodeploy@turbonomic.com>
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+
+RUN echo "Running on $BUILDPLATFORM, and building for $TARGETPLATFORM"
+
+### Atomic/OpenShift Labels - https://github.com/projectatomic/ContainerApplicationGenericLabels
+LABEL name="Hybrid Cloud Container" \
+      vendor="Turbonomic" \
+      version="v8.0.0" \
+      release="1" \
+      summary="Performance assurance for the applications in Openshift" \
+      description="Hybrid Cloud Container leverages Turbonomic control platform, to assure the performance of micro-services running in OpenShift, as well as the efficiency of underlying infrastructure." \
+### Required labels above - recommended below
+      url="https://www.turbonomic.com" \
+      io.k8s.description="Hybrid Cloud Container will monitor and control the entire stack, from OpenShift down to your underlying infrastructure. " \
+      io.k8s.display-name="Hybrid Cloud Container" \
+      io.openshift.expose-services="" \
+      io.openshift.tags="turbonomic,Hybrid Cloud Container"
+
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/turbonomic/kubeturbo
 
-go 1.19
+go 1.20
```
ading1977, line 3: Do we want to match this version with the one specified in `build/Dockerfile.multi-archs` (i.e., `1.20.3`) ? 

Shall we also have a comment here to say that if this version is changed, we need to make sure the golang base image needs to changed as well?

ading1977 (approved):

go.mod
kevinwangcn: Yes, that's not allowed. if I put `1.20.3` in the go.mod, it reports:
```
[kevinw@Kevins-MacBook-Pro.local kubeturbo]$ go mod tidy
go: errors parsing go.mod:
/opt/git/kubeturbo/go.mod:3: invalid go version '1.20.3': must match format 1.23
```
I added a comment in `go.mod` to remind syncing the change to the `Dockerfile.multi-archs`

Thanks @ading1977  @amd-ibm  for helping review this change.

build/Dockerfile.multi-archs
```diff
@@ -0,0 +1,60 @@
...
+
+FROM registry.access.redhat.com/ubi8
+MAINTAINER Turbonomic <turbodeploy@turbonomic.com>
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+
+RUN echo "Running on $BUILDPLATFORM, and building for $TARGETPLATFORM"
+
+### Atomic/OpenShift Labels - https://github.com/projectatomic/ContainerApplicationGenericLabels
+LABEL name="Hybrid Cloud Container" \
```
pankajbatra1, line 18: Does this section need to change?   It seems specific to `suspend` team

pankajbatra1: 'Hybrid Cloud Container' === suspend

kevinwangcn: That section is copied from our current [Dockerfile](https://github.com/turbonomic/kubeturbo/blob/50fa8ecba43289b66bae2163756782bd87b34c48/build/Dockerfile#L8), it was written by Enlin 5 years ago based on git history.  I could change it if you want it.

pankajbatra1 (approved):

PR #840, opened 2023-04-10
[OM-99775]:Add buildx target in the Makefile of kubeturbo-operator

kevinwangcn:
For the background, please refer to #838 , this change will update the kubeturbo-operator build to have a very similar process as the kubeturbo


SumanthKodali999 (approved):
LGTM

pankajbatra1 (approved):

PR #843, opened 2023-04-13
build for target architecture

tian-ma:
The builder is using the architecture of the build CPU which is not correct and results in a binary that cannot run on the target CPU.
This PR adds the compiler flags to fix it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,10 +29,10 @@ debug-product: clean
 
 build: clean
 	@ if [ -z '${TARGETPLATFORM}' ]; then \
-	    echo "target is current platform"; \
+		echo "target is current platform"; \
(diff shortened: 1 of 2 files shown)


kevinwangcn:
@tian-ma Thanks for pointing this out.  I did miss passing the `GOARCH` to the build command. I notice your fix changed the existing target `clean` in the `Makefile`. I think we'd better keep it as it is now. Instead, I suggest adding a new target for fixing this, it will not change any existing target and impact anyone who is using this target. You can refer to my change [here](https://github.com/kevinwangcn/kubeturbo/commit/052b4fa773ed0e9a1ed231b570f987e81d1de194).  Also, we need to sync the fix to the other 2 repos `turbodif` and `prometurbo`.

tian-ma:
Use this instead https://github.com/turbonomic/kubeturbo/pull/844

PR #846, opened 2023-04-24
Support of cpufreqgetterRepository value for the helm chart 

pvlasov:
Similar to OM-99059 for the operator


ading1977:
This is a duplicate of #856

PR #871, opened 2023-06-01
[TRB-23205]: Filter out duplicate commodities

LionelWuW:
# Intent
Filter out duplicated commodities that have the same commodity type and the key from the entity DTO objects.

# Implementation

- Adding filter implementation to prevent appending duplicated sold commodities to the sold commodity list
- Adding logic to prevent appending existing sold commodities to the sold commodity list
- - Adding logic to prevent appending existing buy commodities to the buy commodity list


irfanurrehman:
Thanks @LionelWuW for doing this.
Please push these changes to the correct repo (https://github.com/turbonomic/turbo-go-sdk) and I will review them.

PR #873, opened 2023-06-02
[TRB-39717] Stitching volumes from AWS to kubeturbo fails for manually created volumes

tian-ma:
https://jsw.ibm.com/browse/TRB-39717
For AWS:
Original code requires that a volume ID is in the format `aws://us-east-2c/vol-0e4eaa3ef79bcb5a9`

This PR aims to remove this restriction as discussed in the bug report.
Try to extract volume id by parsing Labels if the volume ID doesn't explicitly contain the zone identifier

Before the fix is applied:
The volume "stitching-bug-validation" appears twice and total number of volumes is 4

After the fix is applied:
Total number of volumes is 3

image

image

### DIFF
--- a/pkg/discovery/stitching/volume_stitching_manager.go
+++ b/pkg/discovery/stitching/volume_stitching_manager.go
@@ -227,10 +227,10 @@ func extractAWSVolumeUuid(volID string, vol *api.PersistentVolume) (string, erro
 					zone = labelV
 				}
 			}
-			if region != "" {
-				parts[0] = region
-			} else if zone != "" {
-				parts[0] = zone + "-"
+			if zone != "" {
+				parts[0] = zone
+			} else if region != "" {
+				parts[0] = region + "-"
 			}
 		}
 	}
--- a/pkg/discovery/stitching/volume_stitching_manager.go
+++ b/pkg/discovery/stitching/volume_stitching_manager.go
@@ -198,16 +198,41 @@ func (aws *awsVolumeUUIDGetter) GetVolumeUUID(vol *api.PersistentVolume) (string
 		return "", fmt.Errorf("not a valid AWS provisioned volume: %v", vol.Name)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/discovery/stitching/volume_stitching_manager.go
```diff
@@ -198,16 +198,41 @@ func (aws *awsVolumeUUIDGetter) GetVolumeUUID(vol *api.PersistentVolume) (string
...
-	suffix := volID[len(awsVolPrefix):]
+	suffix := volID
+	if strings.HasPrefix(volID, awsVolPrefix) {
+		suffix = volID[len(awsVolPrefix):]
+	}
+
 	parts := strings.Split(suffix, "/")
-	if len(parts) != 2 {
-		return "", fmt.Errorf("failed to split uuid (%d): %v, for volume %v", len(parts), parts, volName)
+	if len(parts) < 2 {
```
irfanurrehman, line 214: Line 246 below can cause a problem if this condition is true.
`parts[1]` does not seem to be set (nil pointer dereference) (`len(parts) < 2`)

irfanurrehman (approved):
Thanks for doing this.
Please squash the commits to those with more meaningful messages before merging.

pkg/discovery/stitching/volume_stitching_manager.go
```diff
@@ -198,16 +198,41 @@ func (aws *awsVolumeUUIDGetter) GetVolumeUUID(vol *api.PersistentVolume) (string
...
+				if strings.HasSuffix(labelKey, "zone") {
+					zone = labelV
+				}
+			}
+			if zone != "" {
+				parts[0] = zone
+			} else if region != "" {
+				parts[0] = region + "-"
+			}
+		}
```
ading1977, line 235: Nit:

Remove the `else` brackets.

ading1977: Also can you put an example of the volume labels?

ading1977 (approved):

PR #877, opened 2023-06-13
[TRB-35519]: Support for new merged SLO action in Action executor

libai98:
Support for new merged SLO action in Action executor.

This is the code change for the SLO action merge in kubeturbo action execution.
The background wiki: https://vmturbo.atlassian.net/wiki/spaces/AE/pages/3668901989/SLO+Horizontal+Scale+Action+Merging#Replicas-Represented-as-Attribute-on-WorkloadController

I am still waiting for @amd-ibm to finalize on his code to send action from server to the kubeturbo to be executed.
So this is just a initial draft to get some feedback.

Developer Checks
[x] Full build with unit tests and Checkstyle / SpotBugs tests
[x] Unit tests added / updated
[x]No unlicensed images, no third-party code (such as from StackOverflow)
Integration (Robot Framework) tests added / updated
Manual testing done (and described)
Product sweep run and passed
Developer wiki updated (and linked to this description)

### DIFF
--- a/pkg/action/executor/horizontal_scaler.go
+++ b/pkg/action/executor/horizontal_scaler.go
@@ -2,6 +2,7 @@ package executor
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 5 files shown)


pkg/action/executor/horizontal_scaler.go
```diff
@@ -30,32 +30,63 @@ func (h *HorizontalScaler) Execute(input *TurboActionExecutorInput) (*TurboActio
...
-		return 0, fmt.Errorf("action %v is not a scaling action", atype.String())
+	if atype != proto.ActionItemDTO_HORIZONTAL_SCALE {
+		return 0, fmt.Errorf("action %v is not a horizontal scaling action", atype.String())
+	}
+
+	data := action.GetContextData()
+	var oldValue, newValue string
+
+	for _, item := range data {
+		switch *item.ContextKey {
```
irfanurrehman, line 65: Please add nil checks here and where ever else you might dereference a pointer value

pkg/action/executor/horizontal_scaler.go
```diff
@@ -30,32 +30,63 @@ func (h *HorizontalScaler) Execute(input *TurboActionExecutorInput) (*TurboActio
...
+		case "OLD_REPLICAS":
+			oldValue = *item.ContextValue
+		case "NEW_REPLICAS":
+			newValue = *item.ContextValue
+		}
+	}
+
+	if oldValue == "" {
+		return 0, fmt.Errorf("key 'OLD_REPLICAS' not found in context data")
```
irfanurrehman, line 74: `context data` is unlikely to be understood by anybody.
Probably make it more generic, for example "OLD_REPLICAS value not found in action DTO"
and similar for further error messages.

irfanurrehman: I would have another comment on replica values.
Earlier it was always a provision or a suspend, so it did not matter whats the current number of replicas in the cluster, but I think it does now.
What if the old replica value that was considered for the action by market is now different from the one live in the cluster. Do we still increase/decrease by this `diff` or should we be more `declarative`
cc @amd-ibm @evat-pm

pkg/action/executor/horizontal_scaler.go
```diff
@@ -30,32 +30,63 @@ func (h *HorizontalScaler) Execute(input *TurboActionExecutorInput) (*TurboActio
...
 	}
+
+	if newValue == "" {
+		return 0, fmt.Errorf("key 'NEW_REPLICAS' not found in context data")
+	}
+
+	oldInt, err := strconv.Atoi(oldValue)
+	if err != nil || oldInt <= 0 {
+		return 0, fmt.Errorf("value of 'OLD_REPLICAS' must be a positive integer")
```
irfanurrehman, line 83: We are losing the error info in case Atoi returns err.
```
if err != nil {
    return 0, err
}
if oldInt <= 0 {
....
}

irfanurrehman (changes requested):
Thanks @libai98 for doing this.

pkg/action/executor/horizontal_scaler.go
libai98: I did a new implementation with Andrew's change of injecting replica data in https://github.ibm.com/turbonomic/xl/pull/368.

pkg/action/executor/horizontal_scaler.go
libai98: I did a new implementation with Andrew's change of injecting replica data in https://github.ibm.com/turbonomic/xl/pull/368.

irfanurrehman (dismissed):
Squash the commits to something meaningful before merging please
Thanks for doing this!

irfanurrehman:
> Squash the commits to something meaningful before merging please
> Thanks for doing this!

May be include the changes into  this PR to ensure the older mechanism of provision and suspend pods still works

pkg/action/executor/horizontal_scaler.go
libai98: Discussed this in the team standup. @irfanurrehman will open a jira for this. This discussion will not have the impact on this implementation at this moment.

PR #879, opened 2023-06-14
[TRB-41307]:Check the violation of the limitrange and revert the change on parent controller if execution fails

kevinwangcn:
# Intent
We want to fail the resize action execution ASAP when a **forbidden** error is detected while waiting for the new replica to be created. 

# Background
The customer set up a `limitrange` with a maximum CPU defined in a namespace,  the target value in resize up action is bigger than the maximum value of the `limitrange`. That will cause a `forbidden` error when the new replica set is trying to pull up a new pod.  It's meaningless to keep Kubeturbo waiting for the new pods to be created in that case. 

# Implementation
Check the `limitrange` violation before executing the action.  If there is a violation, simply fail the execution and return to the server with the detailed reason. Also, the new implementation needs to handle the case that no resource requirement is defined in the deployment but there is a default resource requirement defined in the `limitrange`.  Meanwhile, this change also adds a revert logic to the workload controller. If any error happens in waiting for the new pods to be created,  undo the change on the workload controller. 

# Test Done

## Case 1: Check if a forbidden error is detected and if the execution get failed quickly
**1. Create deployment/limitrange/resourcequota in a namespace**
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: quota-with-space-1
  namespace: testquota
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: quota-with-space-1
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: quota-with-space-1
      name: quota-with-space-1
    spec:
      containers:
      - env:
        - name: RUN_TYPE
          value: cpu
        - name: CPU_PERCENT
          value: "100"
        image: beekman9527/cpumemload:latest
        imagePullPolicy: Always
        name: quota-with-space-1-spec
        resources:
          limits:
            cpu: 110m
            memory: 328Mi
          requests:
            cpu: 10m
            memory: 10Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
---
apiVersion: v1
kind: LimitRange
metadata:
  name: cpu-resource-constraint
  namespace: testquota
spec:
  limits:
  - max:
      cpu: 100m
      memory: 4Gi
    min:
      memory: 5Mi
    type: Pod
  - default:
      cpu: 50m
      memory: 1365Mi
    defaultRequest:
      cpu: 50m
      memory: 128Mi
    max:
      cpu: 100m
      memory: 4Gi
    min:
      memory: 5Mi
    type: Container
---
apiVersion: v1
kind: ResourceQuota
metadata:
  name: test-quota
  namespace: testquota
spec:
  hard:
    limits.cpu: 800m
    limits.memory: 1Gi
    pods: "50"
    requests.cpu: "10"
    requests.memory: 20Gi
status:
  hard:
    limits.cpu: 800m
    limits.memory: 1Gi
    pods: "50"
    requests.cpu: "10"
    requests.memory: 20Gi
  used:
    limits.cpu: 110m
    limits.memory: 328Mi
    pods: "1"
    requests.cpu: 10m
    requests.memory: 10Mi
```
**2. Wait for the resize up action to be generated and execute it through UI**
image

**3. The action execution should fail immediately and the log should say**
```
E0620 00:57:13.753020       1 action_handler.go:236] Failed to execute action RIGHT_SIZE on WORKLOAD_CONTROLLER [testquota/quota-with-space-2]: limitrange violation:[maximum cpu usage per Pod is 100m, but limit is 110m, maximum cpu usage per Container is 100m, but limit is 110m]E0620 00:57:13.753034       1 action_handler.go:181] action execution error: limitrange violation:[maximum cpu usage per Pod is 100m, but limit is 110m, maximum cpu usage per Container is 100m, but limit is 110m]
```
image

### DIFF
--- a/pkg/action/executor/k8s_controller.go
+++ b/pkg/action/executor/k8s_controller.go
@@ -3,9 +3,11 @@ package executor
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"regexp"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/turbonomic/kubeturbo/pkg/resourcemapping"
@@ -38,6 +40,7 @@ import (
 type k8sController interface {
 	get(name string) (*k8sControllerSpec, error)
 	update(updatedSpec *k8sControllerSpec) error
+	revert() error
 }
 
 // k8sControllerSpec defines a set of objects that we want to update:
@@ -61,6 +64,7 @@ type kubeClients struct {
 type parentController struct {
 	clients               kubeClients
 	obj                   *unstructured.Unstructured
+	backupObj             *unstructured.Unstructured
 	name                  string
 	ormClient             *resourcemapping.ORMClientManager
 	managerApp            *repository.K8sApp
@@ -99,6 +103,7 @@ func (c *parentController) get(name string) (*k8sControllerSpec, error) {
 	}
 
 	c.obj = obj
+	c.backupObj = c.obj.DeepCopy()
 	int32Replicas := int32(replicas)
 	return &k8sControllerSpec{
 		replicas:       &int32Replicas,
@@ -165,6 +170,7 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 		}
 		return c.ormClient.UpdateOwners(c.obj, ownerInfo, ownerResources)
 	} else {
+		start := time.Now()
 		_, err = c.clients.dynNamespacedClient.Update(context.TODO(), c.obj, metav1.UpdateOptions{})
 		if err != nil {
 			return err
@@ -187,7 +193,7 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 			err4Waiting := c.waitForAllNewReplicasToBeCreated(c.obj.GetName())
 			if err4Waiting != nil {
 				glog.V(2).Infof("Get error while waiting for the new replicas to be created for the workload controller %s/%s:%v", c.obj.GetNamespace(), c.obj.GetName(), err4Waiting)
-				if hasWarningEvent, warningInfo := c.getLatestWarningEvents(c.obj.GetNamespace(), c.obj.GetName()); hasWarningEvent {
+				if hasWarningEvent, warningInfo := c.getLatestWarningEventsSinceUpdate(c.obj.GetNamespace(), c.obj.GetName(), start); hasWarningEvent {
 					return fmt.Errorf(warningInfo)
 				}
 				return err4Waiting
@@ -196,6 +202,32 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 		}
 	}
 	return err
+}
+
+func (c *parentController) revert() error {
+	oldPodSpecUnstructured, found, err := unstructured.NestedFieldCopy(c.backupObj.Object, "spec", "template", "spec")
+	if err != nil || !found {
+		return err
+	}
+	currentObj, err := c.clients.dynNamespacedClient.Get(context.TODO(), c.obj.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	currentPodSpecUnstructured, found, err := unstructured.NestedFieldCopy(currentObj.Object, "spec", "template", "spec")
+	if err != nil || !found {
+		return err
+	}
+	if reflect.DeepEqual(oldPodSpecUnstructured, currentPodSpecUnstructured) {
+		glog.V(4).Infof("There is no change between the original pod spec and the current one for the workload controller %s/%s, no need to revert", c.obj.GetNamespace(), c.obj.GetName())
+		return nil
+	}
+	if err := unstructured.SetNestedField(currentObj.Object, oldPodSpecUnstructured, "spec", "template", "spec"); err != nil {
+		return err
+	}
+	if _, err = c.clients.dynNamespacedClient.Update(context.TODO(), currentObj, metav1.UpdateOptions{}); err != nil {
+		return err
+	}
+	return nil
 }
 
 func (c *parentController) rolloutDeploymentConfig(dc *osv1.DeploymentConfig) error {
@@ -277,7 +309,7 @@ func (c *parentController) waitForAllNewReplicasToBeCreated(name string) error {
 }
 
 // get the latest warning event for a workload controller
-func (c *parentController) getLatestWarningEvents(namespace, name string) (bool, string) {
+func (c *parentController) getLatestWarningEventsSinceUpdate(namespace, name string, start time.Time) (bool, string) {
 	// Get events that belong to the given workload controller
 	events, err := c.clients.typedClient.CoreV1().Events(namespace).List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


irfanurrehman (commented):
I started a review on this, but looking at the implementation, I am personally not in favour of introducing reverts.
Why not check the limitrange, if one exists, and if any of the resource values breaks the limitrange we fail the action before even trying to update the workload controller spec.

kevinwangcn:
> I started a review on this, but looking at the implementation, I am personally not in favour of introducing reverts. Why not check the limitrange, if one exists, and if any of the resource values breaks the limitrange we fail the action before even trying to update the workload controller spec.

@irfanurrehman , thanks for reviewing this change.  I thought about this before I did this implantation. I mentioned it in the [JIRA](https://jsw.ibm.com/browse/TRB-41307?focusedId=12630262&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12630262). We could do a pre-check on `limitrange` before really executing the action. But there are still possibilities that the new pod can't be created due to other reasons even if it passes the limitrange check. The reason could be resource requirements can't be met or somehow timeout.   In those cases, we may still undo the change on the workload controller.

pkg/action/executor/k8s_controller.go
```diff
@@ -241,12 +275,21 @@ func (c *parentController) shouldRolloutDeploymentConfig() (bool, *osv1.Deployme
...
 	return wait.Poll(DefaultRetrySleepInterval, DefaultWaitReplicaToBeScheduled, func() (bool, error) {
 		obj, errInternal := c.clients.dynNamespacedClient.Get(context.TODO(), name, metav1.GetOptions{})
 		if errInternal != nil {
 			return false, errInternal
 		}
 
+		if hasWarningEvent, warningInfo := c.getLatestWarningEventsSinceUpdate(c.obj.GetNamespace(), c.obj.GetName(), start); hasWarningEvent {
+			if strings.Contains(warningInfo, "forbidden") {
+				// if there is a forbidden error, return a error whill will exit the meaningless waiting
+				glog.Errorf("Failed to create new replia for the workload controller %s/%s as there is a forbidden error: %s", c.obj.GetNamespace(), c.obj.GetName(), warningInfo)
```
amd-ibm, line 288: typo: "replia"

amd-ibm (approved):

pkg/action/executor/k8s_controller.go
```diff
@@ -198,6 +204,31 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
...
+	}
+	currentObj, err := c.clients.dynNamespacedClient.Get(context.TODO(), c.obj.GetName(), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	currentPodSpecUnstructured, found, err := unstructured.NestedFieldCopy(currentObj.Object, "spec", "template", "spec")
+	if err != nil || !found {
+		return err
+	}
+	if reflect.DeepEqual(oldPodSpecUnstructured, currentPodSpecUnstructured) {
```
irfanurrehman, line 220: Probably drop an info msg here (probably v4). This should not happen if we have reached revert, right?

pkg/action/executor/k8s_controller.go
```diff
@@ -184,10 +189,11 @@ func (c *parentController) update(updatedSpec *k8sControllerSpec) error {
 		}
 
 		if utilfeature.DefaultFeatureGate.Enabled(features.AllowIncreaseNsQuota4Resizing) {
+			start := time.Now()
```
irfanurrehman, line 192: The start time should be before we start the parent controller update, isnt it?

pkg/action/executor/limitrange_util.go
```diff
@@ -0,0 +1,317 @@
+package executor
+
```
irfanurrehman, line 2: if this is coming from upstream, please drop a proper referene header here. [Example](https://github.com/turbonomic/kubeturbo/blob/master/pkg/discovery/worker/compliance/podaffinity/filtering.go#L2)

pkg/action/executor/limitrange_util.go
```diff
@@ -0,0 +1,317 @@
...
+				for k, v := range limit.MaxLimitRequestRatio {
+					if err := limitRequestRatioConstraint(string(limitType), string(k), v, container.Resources.Requests, container.Resources.Limits); err != nil {
+						errs = append(errs, err)
+					}
+				}
+			}
+		}
+
+		// enforce pod limits on init containers
+		if limitType == corev1.LimitTypePod {
```
irfanurrehman, line 86: same as the comment on line 65

pkg/action/executor/workload_controller_resizer.go
```diff
@@ -74,6 +74,14 @@ func (r *WorkloadControllerResizer) Execute(input *TurboActionExecutorInput) (*T
 		resizeSpecs = append(resizeSpecs, spec)
 	}
 
+	// Verify if the desired podSpec viloates the limitrange
+	desiredPod := buildDesiredPod4QuotaEvaluation(namespace, resizeSpecs, *podSpec)
+	limitrangeViolateErr := CheckLimitrangeViolationOnPod(r.clusterScraper.Clientset, namespace, desiredPod)
+	if limitrangeViolateErr != nil {
+		glog.Errorf("Failed to execute aciton on the workload controller %v/%v due to limitrange violation: %v", namespace, controllerName, limitrangeViolateErr)
```
irfanurrehman, line 81: s/aciton/action

pkg/action/executor/limitrange_util.go
```diff
@@ -0,0 +1,317 @@
...
+						errs = append(errs, err)
+					}
+				}
+				for k, v := range limit.MaxLimitRequestRatio {
+					if err := limitRequestRatioConstraint(string(limitType), string(k), v, container.Resources.Requests, container.Resources.Limits); err != nil {
+						errs = append(errs, err)
+					}
+				}
+			}
+			for j := range pod.Spec.InitContainers {
```
irfanurrehman, line 65: I am not sure if we actually do give resize recommentations for init containers. Is this check needed here?

I can think of a possible environment scenario which can fail our action, even when its not something because of our resize recommendation. A pod was created with a specific limitrange (or not enforceable limit range at all) and the limit range was later created smaller then what the pod is now live with. I am not sure if we should treat that as an actual failure. cc @evat-pm

pkg/action/executor/limitrange_util.go
kevinwangcn: done

pkg/action/executor/k8s_controller.go
kevinwangcn: Yes, I move it in front of the parent controller update

pkg/action/executor/limitrange_util.go
kevinwangcn: This piece of code is copied from K8s admission plugin. For dealing with `InitContainers`,  it's only doing some comparison and map updates. There is no remote call or deep loop in it. And usually, most pods don't have `InitContainers`.  I kind of leaning towards keeping them if that makes sense to you @irfanurrehman . 
Thanks for your time in reviewing this.

irfanurrehman: ACK!

irfanurrehman (approved):

PR #881, opened 2023-06-15
[TRB-41643] Improve logging of node scaling action

tian-ma:
add some extra logging for node scaling (via machine set in openshift)
https://jsw.ibm.com/browse/TRB-41643

### DIFF
--- a/pkg/action/executor/machine_scale_executor.go
+++ b/pkg/action/executor/machine_scale_executor.go
@@ -52,13 +52,19 @@ func (s *MachineActionExecutor) Execute(vmDTO *TurboActionExecutorInput) (*Turbo
 		return nil, fmt.Errorf("unsupported action type %v", vmDTO.ActionItems[0].GetActionType())
 	}
 	// Get on with it.
-	glog.V(3).Infof("Starting execution of machine scaling action: %+v on node %v.", actionType, nodeName)
 	controller, key, err := newController(s.cAPINamespace, nodeName, diff, actionType, s.executor.clusterScraper)
 	if err != nil {
 		return nil, err
 	} else if key == nil {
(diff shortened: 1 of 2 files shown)


pkg/action/executor/machine_scale_executor.go
```diff
@@ -80,5 +79,6 @@ func (s *MachineActionExecutor) Execute(vmDTO *TurboActionExecutorInput) (*Turbo
 	if err != nil {
 		return nil, err
 	}
+	glog.V(3).Infof("Completed execution of machine scaling action: %+v on node %v.", actionType, nodeName)
```
ading1977, line 82: Similar to the above changes.

pkg/action/executor/machine_scale_executor.go
```diff
@@ -45,15 +45,14 @@ func (s *MachineActionExecutor) Execute(vmDTO *TurboActionExecutorInput) (*Turbo
...
-		break
 	case proto.ActionItemDTO_SUSPEND:
 		actionType = SuspendAction
 		diff = -1
-		break
 	default:
 		return nil, fmt.Errorf("unsupported action type %v", vmDTO.ActionItems[0].GetActionType())
 	}
 	// Get on with it.
+	glog.V(3).Infof("Starting execution of machine scaling action: %+v on node %v.", actionType, nodeName)
```
ading1977, line 55: Suggest:

* V(2)
* Depending on the action type, can we say something like the following:
  `Starting to scale up the machineset xxx by 1 replica`, or
  `Starting to scale down the machineset xxx by 1 replica`

tian-ma: Done

ading1977 (approved):

PR #888, opened 2023-06-26
[TRB-23668] Reduce Affinity commodities

irfanurrehman:
***Intent***
This implements couple of improvements to try to reduce the total number of commodities added after processing pod to pod affinities and anti-affinities. Implements solution for https://jsw.ibm.com/browse/TRB-23668

***Background***
We traditionally were seeing huge kubetubro side discovery delays which we drilled down to affinity processing logic and improved the same in https://github.com/turbonomic/kubeturbo/pull/791. After this we saw Turbo server side performance issues in super large enviroments which we could attribute to large number of commodities that are added as a result of kubeturbo side affinity processing. This PR makes few improvements:
1. By using the workload controller unique name in place of the pod unique name as the key in these commodities. This will reduce the total number of commodities by the ratio of average number of replicas per workload in the given cluster.
2. Uses segmentation commodities for the spread workload replicas(pod antiaffinity to itself to ensure spread across topology) wrt to hostname as topology. 
3. Uses simple label commodities with pod unique name for the spread workload replicas(pod antiaffinity to itself to ensure spread across topology) wrt to topologies other then hostname.


amd-ibm (commented):
This looks good to me, but I see that your build is failing. Please fix the build and I'll take a second look.

pkg/discovery/worker/compliance/podaffinity/processor.go
```diff
@@ -101,11 +107,12 @@ func GetNamespaceLabelsSnapshot(ns string, nsLister NamespaceLister) (nsLabels l
 
 // ProcessAffinities returns a map of nodes to list of pods which can be placed on each
 // respective node and a set of pods which have affinities.
```
kevinwangcn, line 109: @irfanurrehman  Could you please add some explanation here to tell what the newly added 2 `sets.String` in the returned results are used for?

irfanurrehman: Updated! Thanks for taking a look.

amd-ibm (approved):

PR #900, opened 2023-07-10
[TRB-42587]:Support dynamic logging level configuration

kevinwangcn:
# Intent

To make Kubeturbo support dynamic logging level configuration

# Background

https://jsw.ibm.com/browse/TRB-42580

# Testing

## Case 1:  Override the logging level value defined in the command line of the Kubeturbo deployment
**1.Set the verbosity to 6 in the Kubeturbo deployment**
```[root@api.ocp410kev.cp.fyre.ibm.com ~]# k describe deploy kubeturbo-kw-114 
Name:                   kubeturbo-kw-114
Namespace:              default
CreationTimestamp:      Mon, 05 Dec 2022 14:44:51 -0500
Labels:                 app=kubeturbo-kw-114
Annotations:            deployment.kubernetes.io/revision: 62
Selector:               app=kubeturbo-kw-114
Replicas:               1 desired | 1 updated | 1 total | 1 available | 0 unavailable
StrategyType:           RollingUpdate
MinReadySeconds:        0
RollingUpdateStrategy:  25% max unavailable, 25% max surge
Pod Template:
  Labels:           app=kubeturbo-kw-114
  Annotations:      kubeturbo.io/monitored: false
  Service Account:  turbo-user
  Containers:
   kubeturbo-kw-114-cnt:
    Image:      kevin0204/kubeturbo:loadcm
    Port:       <none>
    Host Port:  <none>
    Args:
      --turboconfig=/etc/kubeturbo/turbo.config
      --v=6 <------------change here
      --kubelet-https=true
      --kubelet-port=10250
```
**2. Set the logging level to 2 in the configmap of Kubeturbo**
```
[root@api.ocp410kev.cp.fyre.ibm.com ~]# k describe cm turbo-config-kw-114 
Name:         turbo-config-kw-114
Namespace:    default
Labels:       <none>
Annotations:  <none>

Data
====
turbo-autoreload.config:
----
{
    "logging": {
       "level": 2 <------here
    }
}
turbo.config:
----
```
**3. Start the Kubeturbo, and expect to see the logging level set to 2**
```
[root@api.ocp410kev.cp.fyre.ibm.com ~]# k logs -f kubeturbo-kw-114-6f575b7ffb-zggkm |grep -i logging
I0710 14:59:51.679557       1 kubeturbo_builder.go:1045] Logging level is changed from 6 to 2
```

## Case 2:  Change the logging level in the configmap and check if the new verbosity is applied in the Kubetrubo without a pod restart
**1. Edit the configmap of the Kubeturbo and change the logging level from 2 to 5**
```
[root@api.ocp410kev.cp.fyre.ibm.com ~]# k describe cm turbo-config-kw-114 
Name:         turbo-config-kw-114
Namespace:    default
Labels:       <none>
Annotations:  <none>

Data
====
turbo-autoreload.config:
----
{
    "logging": {
       "level": 5
    }
}
```
**2. Check the Kubeturbo and search its logs with the keyword `logging`**
```
[root@api.ocp410kev.cp.fyre.ibm.com ~]# k logs -f kubeturbo-kw-114-6f575b7ffb-zggkm |grep -i logging
I0710 14:59:51.679557       1 kubeturbo_builder.go:1045] Logging level is changed from 6 to 2
I0710 15:11:05.737864       1 kubeturbo_builder.go:1045] Logging level is changed from 2 to 5 <----here
````
**3. Do a discovery and check if there are similar output in the logs, these logs has the verbosity as `5`**
```
[root@api.ocp410kev.cp.fyre.ibm.com ~]# k logs -f kubeturbo-kw-114-6f575b7ffb-zggkm |grep -i "Adding"
I0710 15:17:28.052107       1 node_entity_dto_builder.go:253] Adding label commodity for Node worker2.ocp410kev.cp.fyre.ibm.com with key : kubernetes.io/hostname=worker2.ocp410kev.cp.fyre.ibm.com
I0710 15:17:28.052118       1 node_entity_dto_builder.go:253] Adding label commodity for Node worker2.ocp410kev.cp.fyre.ibm.com with key : kubernetes.io/os=linux
I0710 15:17:28.052122       1 node_entity_dto_builder.go:253] Adding label commodity for Node worker2.ocp410kev.cp.fyre.ibm.com with key : node-role.kubernetes.io/worker=
...
```

## Case 3:  Deploy Kubeturbo through helm and check the configmap 
**1.Deply the Kubeturbo through helm chart in the `test1` namespace and specify the logging level with 6**
```
helm install --debug kubeturbo-test-5 ./kubeturbo  --set serverMeta.turboServer=http://9.46.1.1 --set image.tag=loadcm --set image.repository=kevin0204/kubeturbo --set logging.level=6 --set targetConfig.targetName=helm-test-5  --set roleName=turbo-cluster-admin --set roleBinding=helm-test5 --namespace test1
```
**2.Check the pod status and the configmap**
```
[kevinw@Kevins-MacBook-Pro.local deploy]$ k get pods -n test1
NAME                               READY   STATUS    RESTARTS   AGE
kubeturbo-test-5-fbfcf8c67-7rg24   1/1     Running   0          8m18s
[kevinw@Kevins-MacBook-Pro.local deploy]$ k get cm turbo-config-kubeturbo-test-5 -n test1 -oyaml
apiVersion: v1
data:
  turbo-autoreload.config: |-
    {
      "logging": {
        "level": 6  <-------here
      }
    }
  turbo.config: |-
    {
      "communicationConfig": {
        "serverMeta": {
          "version": "8.0",
          "turboServer": "http://9.46.1.1"
        },
        "restAPIConfig": {
          "opsManagerUserName": "Turbo_username",
          "opsManagerPassword": "Turbo_password"
        },
        "sdkProtocolConfig": {
           "registrationTimeoutSec": 300,
           "restartOnRegistrationTimeout": false
        }
      },
      "HANodeConfig": {
        "nodeRoles": ["master"]
      },
      "targetConfig": {
        "targetName": "helm-test-5"
      }
    }
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: kubeturbo-test-5
    meta.helm.sh/release-namespace: test1
  creationTimestamp: "2023-07-17T19:36:00Z"
  labels:
    app.kubernetes.io/managed-by: Helm
  name: turbo-config-kubeturbo-test-5
  namespace: test1
  resourceVersion: "119698893"
  uid: 23c0f204-55f0-4f33-9156-e95d4fdb3d26
```
# Checklist

These are the items that must be done by the developer and by reviewers before the change is ready to merge. Please ~~strikeout~~ any items that are not applicable, but don't delete them

- [ ] Developer Checks
    - [x] Full build with unit tests and fmt and vet checks
    ~~- [ ] Unit tests added / updated~~
    - [x] No unlicensed images, no third-party code (such as from StackOverflow)
    ~~- [ ] Integration tests added / updated~~
    - [x] Manual testing done (and described)
    ~~- [ ] Product sweep run and passed~~
    ~~- [ ] Developer wiki updated (and linked to this description)~~<---there is a separate task for this
- [ ] Reviewer Checks
    - [ ] Merge request description clear and understandable
    - [ ] Developer checklist items complete
    - [ ] Functional code review (how is the code written)
    - [ ] Architectural review (does the code try to do the right thing, in the right way)
    - [ ] Defensive coding (incoming data checked / sanitized, exceptions logged, clear error messages)
    - [ ] No unlicensed images, no third-party code (such as from StackOverflow)
    - [ ] Security review checklist complete.

# Audience

_(@ mention any `review/...` groups or people that should be aware of this merge request)_

### CODE REVIEW

**Comment by tian-ma** [comment]
It is strange that we need an extra file called dynamic to do dynamic config.
This is redundant. I understand we already debated this but it still looks weird

File: deploy/kubeturbo/templates/configmap.yaml [hunk 1]
```diff
@@ -55,4 +55,10 @@ data:
         "workloadController": "{{ default "" .Values.annotationWhitelist.workloadController }}"
       {{- end }}
       }
+    }
+  turbo-dynamic.config: |-
```
ading1977, line 59: I wonder if it will be better to call this `turbo-autoreload.config`?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -1009,3 +1011,47 @@ func reviewSCCAccess(namespace string, kubeClient kubernetes.Interface) bool {
...
+	viper.SetConfigType("json")
+	viper.SetConfigName(dynamicConfigfileName)
+	for {
+		verr := viper.ReadInConfig()
+		if verr == nil {
+			break
+		} else {
+			glog.V(5).Infof("Can't find the dynamic config file %s/%s, will retry in 3 seconds", dynamicConfigfilePath, dynamicConfigfileName)
+			time.Sleep(3 * time.Second)
+		}
```
ading1977, line 1030: We don't have V(5) logging, the maximum is V(4).

I have a generic question. If the dynamic config does not exist initially when `kubeturbo` starts, but user adds that in the configmap later on, does `kubeturbo` need to restart in order to pick up this file?

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -1009,3 +1011,47 @@ func reviewSCCAccess(namespace string, kubeClient kubernetes.Interface) bool {
...
+	updateConfig := func() {
+		newLoggingLevel := viper.GetString("logging.level")
+		currentLoggingLevel := pflag.Lookup("v").Value.String()
+		if newLoggingLevel != currentLoggingLevel {
+			if newLogVInt, err := strconv.Atoi(newLoggingLevel); err != nil || newLogVInt < 0 {
+				glog.V(1).Infof("Invalid log verbosity %v in the dynamic config file", newLoggingLevel)
+			} else {
+				err := pflag.Lookup("v").Value.Set(newLoggingLevel)
+				if err != nil {
+					glog.V(1).Infof("Can't apply the new logging level setting due to the error:%v", err)
```
ading1977, line 1043: This should be an `Errorf`

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -1009,3 +1011,47 @@ func reviewSCCAccess(namespace string, kubeClient kubernetes.Interface) bool {
...
+		}
+	}
+
+	glog.V(1).Infof("Start watching the dynamic config file %s/%s", dynamicConfigfilePath, dynamicConfigfileName)
+	updateConfig := func() {
+		newLoggingLevel := viper.GetString("logging.level")
+		currentLoggingLevel := pflag.Lookup("v").Value.String()
+		if newLoggingLevel != currentLoggingLevel {
+			if newLogVInt, err := strconv.Atoi(newLoggingLevel); err != nil || newLogVInt < 0 {
+				glog.V(1).Infof("Invalid log verbosity %v in the dynamic config file", newLoggingLevel)
```
ading1977, line 1039: This should be an `Errorf`

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: no need to restart, that's why I use a `for` loop to keep checking if the `autorealod-config` exists or not.  Once it's added, the new logging level will be picked up on the fly.

deploy/kubeturbo/templates/configmap.yaml
kevinwangcn: Change it to `autoreload`, and also updated the related variable's name and logging message.  Thanks for the review @ading1977

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: done

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: done

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -1009,3 +1011,47 @@ func reviewSCCAccess(namespace string, kubeClient kubernetes.Interface) bool {
...
+	viper.AddConfigPath(autoReloadConfigFilePath)
+	viper.SetConfigType("json")
+	viper.SetConfigName(autoReloadConfigFileName)
+	for {
+		verr := viper.ReadInConfig()
+		if verr == nil {
+			break
+		} else {
+			glog.V(4).Infof("Can't read the autoreload config file %s/%s due to the error: %v, will retry in 3 seconds", autoReloadConfigFilePath, autoReloadConfigFileName, verr)
+			time.Sleep(3 * time.Second)
```
ading1977, line 1029: I don't think this will be a common case. Suggest increasing the interval to 60 seconds.

deploy/kubeturbo/values.yaml
```diff
@@ -89,6 +89,10 @@ HANodeConfig:
 #   GitopsApps: false (default: true)
 #   GoMemLimit: true (default: false)
 
+# Dynamic logging level
```
ading1977, line 92: `Logging level. Changing this value does not require restart of kubeturbo.`

ading1977 (approved):

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: When the user adds this dynamic logging level the first time,  if we change this interval to 60s, the user may need to wait for around 2 minutes to see the logging level is finally changed, 1 minute on this waiting, 1 minute waiting for the file change to be reflected inside the container. Is it a bit longer?

irfanurrehman: I am not sure how viper works, but with inotify we cannot install a watch on a non existent file. 
We first check if the actual file exists, if it does, we load it and install a watch on it. If not, we do install the fsnotify watch on the parent directory first, which in effect means when a new file is created we get a notification. Once we have a new file event for our file, we can load that and start watching for that file instead.


If viper is handling the above logic internally, for the case you seem to be handling above, if the file exists at start, but is malformed, we can simply post an error and move on. Whenever the file is corrected next, our notify watch should catch the update event and we then proceed with that. I don't see a point in retry.
What I mean is all this handling should ideally be in you callback `updateConfig()` below.

cmd/kubeturbo/app/kubeturbo_builder.go
```diff
@@ -1009,3 +1011,47 @@ func reviewSCCAccess(namespace string, kubeClient kubernetes.Interface) bool {
...
+
+			}
+		}
+	}
+	updateConfig() //update the logging level during startup
+	viper.OnConfigChange(func(in fsnotify.Event) {
+		updateConfig()
+	})
+
+	viper.WatchConfig()
```
irfanurrehman, line 1056: I believe this is expected to block. Is it possible this can return unexpectedly? if its possible we need a for loop here to restart the watch.

irfanurrehman (commented):
Thanks for doing this @kevinwangcn

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: @irfanurrehman  Thanks for reviewing my code.  I believe you have seen the loop I use to check if the config file is ready for watching or not.  Just as you expect, I keep checking the config file until it's pared well and then add a watch on it.  If the file doesn't exist or is not parsed well, the code will be waiting there.

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: > 

This is the implementation of the `watchConfig` function.  It doesn't block, it starts another Go routine and keeps watching the file event.  It doesn't return an error,  all of the errors are handled internally.  
```
func (v *Viper) WatchConfig() {
	initWG := sync.WaitGroup{}
	initWG.Add(1)
	go func() {
		watcher, err := newWatcher()
		if err != nil {
			log.Fatal(err)
		}
		defer watcher.Close()
		// we have to watch the entire directory to pick up renames/atomic saves in a cross-platform way
		filename, err := v.getConfigFile()
		if err != nil {
			log.Printf("error: %v\n", err)
			initWG.Done()
			return
		}

		configFile := filepath.Clean(filename)
		configDir, _ := filepath.Split(configFile)
		realConfigFile, _ := filepath.EvalSymlinks(filename)

		eventsWG := sync.WaitGroup{}
		eventsWG.Add(1)
		go func() {
			for {
				select {
				case event, ok := <-watcher.Events:
					if !ok { // 'Events' channel is closed
						eventsWG.Done()
						return
					}
					currentConfigFile, _ := filepath.EvalSymlinks(filename)
					// we only care about the config file with the following cases:
					// 1 - if the config file was modified or created
					// 2 - if the real path to the config file changed (eg: k8s ConfigMap replacement)
					const writeOrCreateMask = fsnotify.Write | fsnotify.Create
					if (filepath.Clean(event.Name) == configFile &&
						event.Op&writeOrCreateMask != 0) ||
						(currentConfigFile != "" && currentConfigFile != realConfigFile) {
						realConfigFile = currentConfigFile
						err := v.ReadInConfig()
						if err != nil {
							log.Printf("error reading config file: %v\n", err)
						}
						if v.onConfigChange != nil {
							v.onConfigChange(event)
						}
					} else if filepath.Clean(event.Name) == configFile &&
						event.Op&fsnotify.Remove&fsnotify.Remove != 0 {
						eventsWG.Done()
						return
					}

				case err, ok := <-watcher.Errors:
					if ok { // 'Errors' channel is not closed
						log.Printf("watcher error: %v\n", err)
					}
					eventsWG.Done()
					return
				}
			}
		}()
		watcher.Add(configDir)
		initWG.Done()   // done initializing the watch in this go routine, so the parent routine can move on...
		eventsWG.Wait() // now, wait for event loop to end in this go-routine...
	}()
	initWG.Wait() // make sure that the go routine above fully ended before returning
}
```

deploy/kubeturbo/values.yaml
kevinwangcn: done

cmd/kubeturbo/app/kubeturbo_builder.go
kevinwangcn: @ading1977  I change the check interval to `30s`, does that make sense to you?

